PR #1, opened 2017-07-26
Update time usage

collijk:


vivarium/framework/state_machine.py
```diff
@@ -5,16 +5,18 @@
 import numpy as np
 
 
-def _next_state(index, transition_set, population_view):
+def _next_state(index, time, transition_set, population_view):
```
alec-deason: Elsewhere you're calling this event_time. Should that be consistent here?

collijk: Yes.

PR #3, opened 2017-09-13
Feature/mock dataset objects

alec-deason:
This PR has the new component management system that will be used for the data artifact changes that are coming soon. These changes are self contained and backward compatible.

New features:
- Components are read from config and initialized by a new ComponentManager class which can be replaced with a specialized subclass by specifying a different type in the config.
- The ComponentManager class tracks which components are being loaded which affords systems which need to be able to query the component list early in the initialization process (for example data artifact builders or risk factor implementations that are aware at setup time of what causes are being modeled (which would be future work))
- Adds the concept of DatasetManager, which is a dummy object in this version but is used by the data artifact implementation.
-  The ComponentManager can execute custom code in response to directives in component descriptions or placeholders in component object attributes. This is used by the data artifact implementation to determine which data to load out of the artifact during simulation setup and what data to include when creating the artifact.

### DIFF
--- a/tests/framework/test_components.py
+++ b/tests/framework/test_components.py
@@ -74,7 +74,7 @@ def test_import_function_by_path():
     assert func is _import_by_path
 
 def test_bad_import_by_path():
-    with pytest.raises(ModuleNotFoundError):
+    with pytest.raises(ImportError):
         cls = _import_by_path('junk.garbage.SillyClass')
     with pytest.raises(AttributeError):
         cls = _import_by_path('vivarium.framework.components.SillyClass')
--- a/tests/framework/test_components.py
+++ b/tests/framework/test_components.py
@@ -1,10 +1,15 @@
 import pytest
+import sys
+import os
 import yaml
 import ast
 from unittest.mock import patch, mock_open
 
 from vivarium.framework.components import _import_by_path, load_component_manager, ComponentManager, DummyDatasetManager, ComponentConfigError, _extract_component_list, _component_ast_to_path, _parse_component, ParsingError, _prep_components
 from vivarium import config
+
+# Fiddle the path so we can import from this module
+sys.path.append(os.path.dirname(__file__))
 
 TEST_COMPONENTS = """
 components:
@@ -54,6 +59,10 @@ class MockComponentB(MockComponentA):
 
 def mock_component_c():
     pass
+
+# This very strange import makes it so the classes in the current scope have the same
+# absolute paths as the ones my tests will cause the tools to import so I can compare them.
+from test_components import MockComponentA, MockComponentB, mock_component_c, MockDatasetManager, MockComponentManager
 
 def test_import_class_by_path():
     cls = _import_by_path('collections.abc.Set')
--- a/tests/framework/test_components.py
+++ b/tests/framework/test_components.py
@@ -1,12 +1,15 @@
-import pytest
 import sys
 import os
-import yaml
 import ast
 from unittest.mock import patch, mock_open
 
-from vivarium.framework.components import _import_by_path, load_component_manager, ComponentManager, DummyDatasetManager, ComponentConfigError, _extract_component_list, _component_ast_to_path, _parse_component, ParsingError, _prep_components
-from vivarium import config
+import pytest
+import yaml
+
+from vivarium.framework.components import (_import_by_path, load_component_manager, ComponentManager,
+                                           DummyDatasetManager, ComponentConfigError, _extract_component_list,
+                                           _component_ast_to_path, _parse_component, ParsingError, _prep_components)
+
 
 # Fiddle the path so we can import from this module
 sys.path.append(os.path.dirname(__file__))
@@ -33,19 +36,23 @@ TEST_CONFIG_CUSTOM_DATASET_MANAGER = """
         dataset_manager: test_components.MockDatasetManager
 """
 
+
 class MockComponentManager:
     def __init__(self, components, dataset_manager):
         self.components = components
         self.dataset_manager = dataset_manager
 
+
 class MockDatasetManager:
     def __init__(self):
         self.constructors = {}
+
 
 class MockComponentA:
     def __init__(self, *args):
         self.args = args
         self.builder_used_for_setup = None
+
 
 class MockComponentB(MockComponentA):
     def setup(self, builder):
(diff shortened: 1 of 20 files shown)


vivarium/framework/state_machine.py
```diff
@@ -339,11 +336,27 @@ def _normalize_probabilities(self, outputs, probabilities):
...
+
+    def __iter__(self):
+        return iter(self.transitions)
+
+    def __len__(self):
+        return len(self.transitions)
+
     def __str__(self):
-        return repr(self)
+        return repr(self.transitions)
```
collijk: Think you still want repr(self) here so that the transitions get passed through str.

alec-deason: I've changed this from being a list subclass to storing the transitions on the attribute self.transitions. So I think this does what you think it should, unless you're suggesting that there is something in the TransitionSet itself that should be captured in the string.

collijk: Oh, no, it's just that it feels inconsistent that `str(transition_set)` essentially returns `[repr(x) for x in transition_set]` while `repr(transition_set)` returns `[str(x) for x in transition_set]`.

vivarium/framework/components.py
```diff
@@ -1,63 +1,296 @@
...
+
+        Parameters
+        ----------
+        components:
+          Components to register
+        """
+
+        self.components = components + self.components
+
+    def setup_components(self, builder):
```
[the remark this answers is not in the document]
collijk: **For Discussion**
Also, we really should think about how to make the component setup more reasonable.  It should happen in three distinct phases: Gather the roots of component trees, flatten the trees, perform the setup.  I know that's not the goal of this PR, but we should do it soon.

vivarium/framework/components.py
```diff
@@ -1,63 +1,296 @@
...
+       The configuration to read from
+    """
+
+    def _process_level(level, prefix):
         component_list = []
-        for c in level:
-            if isinstance(c, dict):
-                for k, v in c.items():
-                    component_list.extend(process_level(v, prefix + [k]))
+        for child in level:
```
collijk: Thanks for words!

vivarium/framework/components.py
```diff
@@ -1,63 +1,296 @@
...
+
+    Parameters
+    ----------
+    desc
+        The component definition
+    constructors
+        Dictionary of callables for creating argument objects
+    """
+
+    component, *args = ast.iter_child_nodes(list(ast.iter_child_nodes(
```
collijk: This is maybe useful to have as a standalone function.  It needs an intelligible name and you also use it a few times in the tests.

vivarium/framework/components.py
```diff
@@ -1,63 +1,296 @@
...
+        list(ast.iter_child_nodes(ast.parse(desc)))[0]))[0])
+    component = _component_ast_to_path(component)
+    new_args = []
+    for arg in args:
+        parsed = False
```
collijk: **For discussion**
`continue`

vivarium/framework/components.py
```diff
@@ -1,63 +1,296 @@
...
+                constructor = constructors.get(val.__class__)
+                if constructor:
+                    setattr(component, attr, constructor(val))
 
             # Establish the initial configuration
             if hasattr(component, 'configuration_defaults'):
-                config.read_dict(component.configuration_defaults, layer='component_configs', source=module_path)
+                config.read_dict(component.configuration_defaults, layer='component_configs', source=component)
 
             if call:
```
collijk: I also think it's worth revisiting the all components should be classes discussion.  I think that makes a lot of the code in this module much simpler.

vivarium/framework/engine.py
```diff
@@ -1,12 +1,13 @@
 """The engine."""
+import argparse
```
collijk: **For Discussion**
Make engine a class.

tests/framework/test_components.py
```diff
@@ -0,0 +1,238 @@
...
+
+import pytest
+import yaml
+
+from vivarium.framework.components import (_import_by_path, load_component_manager, ComponentManager,
+                                           DummyDatasetManager, ComponentConfigError, _extract_component_list,
+                                           _component_ast_to_path, _parse_component, ParsingError, _prep_components)
+
+
+# Fiddle the path so we can import from this module
```
collijk: **For Discussion**
This weirdness.

tests/framework/test_components.py
```diff
@@ -0,0 +1,238 @@
+import sys
+import os
+import ast
+from unittest.mock import patch, mock_open
```
collijk: **For Discussion**
Moving to pytest-mock

vivarium/framework/components.py
```diff
@@ -1,63 +1,296 @@
...
+    pass
+
+
+class DummyDatasetManager:
+    """Placeholder implementation of the DatasetManager"""
+    def __init__(self):
+        self.constructors = {}
+
+
+def _import_by_path(path: str) -> Union[type, Callable]:
```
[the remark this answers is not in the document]
alec-deason: Hmm. You may be right that `Union[type, Callable]` is redundant. I gave up trying to use mypy to validate these declarations. In my head I'm going 'it's a class or a function' but you're right that a class is also a Callable.

vivarium/framework/components.py
```diff
@@ -1,63 +1,296 @@
...
+    """Import a class or function given it's absolute path.
+
+    Parameters
+    ----------
+    path:
+      Absolute class to object to import
```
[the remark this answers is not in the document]
alec-deason: Well, that typo makes my docstring almost unreadable but I think in our case the paths will always be absolute though that being said the function will also work on relative paths so it should specify in the docstring.

vivarium/framework/components.py
```diff
@@ -1,63 +1,296 @@
...
 
-def read_component_configuration(path):
-    if path.endswith('.yaml'):
-        with open(path) as f:
-            component_config = yaml.load(f)
-            return prepare_component_configuration(component_config, path=path)
+from vivarium import config
+
+
+class ComponentConfigError(Exception):
```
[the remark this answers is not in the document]
alec-deason: Yep

vivarium/framework/components.py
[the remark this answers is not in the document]
alec-deason: mypy agrees that type is a Callable so I'll use that.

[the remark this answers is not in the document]
collijk: Excellent.  I wasn't actually certain.

alec-deason:
Yeah, I need to actually read the peps to understand how this stuff is
meant to be interpreted. But I haven't done it yet.

PR #7, opened 2017-10-26
Feature/dataset objects

alec-deason:


vivarium/framework/components.py
```diff
@@ -50,34 +50,35 @@ def load_component_manager(config: ConfigTree):
...
+        component_config = config.components
+        del config.components
+    else:
+        component_config = {}
+
+    if 'configuration' in config:
+        model_config = config.configuration
+        source = model_config.source if 'source' in model_config else None
+        del model_config.source
+        del config.configuration
```
c-werner: What are you trying to accomplish with these `del`'s?
`model_config` should hold a reference to `config.configuration` and now you're deleting it, right before you pass it into `config.update()`

alec-deason: Embarrassingly, I don't remember. This is a stale PR anyway and the code has moved meaningfully since it was open. There'll be a new version of this up in the near future which it'd be awesome if you looked at. For now I'm going to close this one down.

PR #10, opened 2018-01-30
Feature/index map

collijk:
The hash algorithm needs some work in terms of optimization, but it's fast enough for all models we currently use and have used in the past..  While it fails the uniformity test, it does not do so badly based on visual inpsection of the hash distribution.


alec-deason:
This looks reasonable assuming the actual lookup code is coming in the next PR. My only concern is around the interperability of the integer hasher which I think is easily resolved with either a comment or explicit overflow handling, which I think I prefer.

PR #15, opened 2018-03-01
Feature/disconnect initialize simulants from event system

collijk:


vivarium/framework/population.py
```diff
@@ -178,21 +190,60 @@ def get_view(self, columns, query=None):
...
 
-    def _create_simulants(self, count, population_configuration=None):
+    def _order_initializers(self) -> None:
+        unordered_initializers = deque(self._population_initializers)
+        starting_length = -1
+        available_columns = []
+        self._population_initializers = []
+
+        # This is the brute force N^2 way because constructing a dependency graph is work
+        # and in practice this should run in about order N time due to the way dependencies are
```
alec-deason: Also, N is always meaningfully small

c-werner: Should it be an error for two initializers to create the same columns?

vivarium/framework/population.py
```diff
@@ -178,21 +190,60 @@ def get_view(self, columns, query=None):
         """
         return PopulationView(self, columns, query)
 
-    def get_simulant_creator(self):
+    def register_simulant_initializer(self, initializer: Callable,
+                                      creates_columns: Sequence[str]=(), requires_columns: Sequence[str]=()):
+        self._population_initializers.append((initializer, creates_columns, requires_columns))
```
c-werner: should this `raise` if `_initializers_ordered`?

collijk: There's an implicit contract that there won't be `_initializers_ordered` when this method gets called.  This method is only accessible from the `Builder` object that gets passed into each component's setup method.  `_initializers_ordered` only gets populated the first time we create simulants, an event guaranteed by the framework to occur after the setup phase.

vivarium/framework/population.py
collijk: Yes.  You can see below (Line 223) that makes this check.  We had to comment it out to deal with an issue in a downstream library we use internally.  It'll be updated soon.

PR #21, opened 2018-04-13
Feature/make component manager extensible

collijk:
Okay, so there a bunch of changes here.  

- Command Line Interface:
  I've swapped out the argparse implementation for one that uses `click`.  This is mostly a consistency thing.  `click` has some nice features and I've been using it to build a slightly nicer cli for the distributed runner.  For the user, it looks exactly the same.

- ConfigTree:
  I've made all methods that modify the `ConfigTree` except `update` private.  `update` handles files and dictionaries and other `ConfigTree` objects.  I think we should probably deprecate assignment via `config.a.b.c = value` as it is a source of many, many errors, but for now it still works.

- Configuration and component management:
   Several things have happened here.  
   1.  I've clarified the usage of `configuration` inside the framework.  The main `yaml` file and it's `ConfigTree` representation are now the `model_specification_file` and the `model_specification`, respectively.  This change reflects the fact that the main yaml file has both a `components` block with a list of all components to be included as well as a `configuration` block that dictates how those componenents and the simulation itself should be configured.  That `configuration` block in the `model_specification` is now what is canonically referred to as the `configuration`.  This is marginally important for users of the simulation and very important for the developers as it was a very frequent cause of regression issues.  
   2.  The transformation from `model_specification.components` to a group of completely setup components (initialized and with the `setup` method called) is now managed by two separate objects.  I've introduced a `ComponentConfigurationParser` whose only job is is to translate the `model_specification.components` configuration block into a group of initialized components.  It then passes these components to a `ComponentManager` which is responsible for setting them up.  This split is nice both because it separates two very different kinds of behavior and because it allows for users to provide custom parsing for the `components` section.  The default behavior is to treat that section like a bunch of python import paths.  This will always be the fallback behavior.  However, where we have generic components, we can provide custom parsing behavior.  E.g. instead of 
```
components: 
    ceam_public_health:
        risks:
            - CategoricalRiskFactor('riskA') 
            - CategoricalRiskFactor('riskB')
            - ContinuousRiskFactor('riskC')
            - ContinuousRiskFactor('riskD')
            - CategoricalRiskFactor('riskE')
            - CategoricalRiskFactor('riskF')
```
I can set up the parsing to just take in the list of risks and select the appropriate component:
```
components:
    ceam_public_health:
        risks: ['riskA', 'riskB', 'riskC', 'riskD', 'riskE', 'riskF']
```
Or to even dynamically select the appropriate risks based on the included diseases.

A user can implement behavior like this by inheriting from `vivarium.framework.components.parser.ComponentConfigurationParser` and overriding `parse_component_config`.  
    3. The `ConfigurationManager` is now available from the builder and has a public interface.  This interface allows the user to add components to the simulation at setup time (this replaces the behavior where components return their sub-components at setup time).  It also allows the user to add global components to the simulation.  The guarantee here is that global components will be initialized after all normal components.  Finally it provides an unimplemented method to query components in the system.  This behavior is meant to be implemented by a custom component manager class (e.g. a `PublicHealthComponentManager` that knows about `RiskFactor` and `DiseaseModel` components).  The intended behavior is to allow a component that needs or supplies more "global" information (such as a component to do risk factor correlations) to delay setup until we can guarantee that all necessary components (e.g. the risk factors) are instantiated and setup.  
    4. Additionally, I've introduced a `PluginManager`.  This was for two reasons.  The big important one was to systematize the way that the framework thinks about all non-component components.  The second is to allow downstream users to override core framework components (the simulation clock, the component manager, etc.) and to provide additional tools they want available from the builder.  Our immediate use case will be the data manager, which is completely unnecessary for our core simulation framework but essential for the public health work.  I'll detail this more later as I implement some of the use cases.  

- I've promoted the `Metrics` component from `ceam_public_health` into `vivarium`.  It was always used in the `engine` so now it's a guaranteed available pipeline in the system regardless of the simulation.

- I've removed a few of the results writer functions.  See the `ceam_experiments` PR.

- I've replaced the `setup_simulation` and `pump_simulation` tools in `vivarium.test_util`.  `setup_simulation` now only takes a list of components and a `ConfigTree` object as inputs.  Parameters must be set in the passed configuration object.  This cleans up some of the contracts about the simulation you should expect.  There is also a new `setup_simulation_from_model_specification` function that allows a user to build a simulation directly from a `yaml` file.  The simulation context object returned from these methods is a new `InteractiveContext` object.  This object has several convenience methods like `run_for(duration)`, `run_until(end_time)`, `step(step_size)`,  `take_steps(number_of_steps, step_size)`, `list_values()`, `reset()`, `replace_component(old_component, new_component)`, etc.  The ones around time are guaranteed to work.  The stranger ones about replacing components, interacting with the pipelines, resetting the simulation, etc., still need some user testing.  These tools are now available from `vivarium.interface.interactive` rather than `vivarium.test_util`.

- A ton of new tests.

This PR is ready for review.  I still have some documentation I want to write, but it works in local, cluster, and interactive environments so is essentially final.


vivarium/framework/components/parser.py
```diff
@@ -0,0 +1,138 @@
...
+    """Gets the component configuration parser"""
+    return import_by_path(simulation_config.vivarium.component_configuration_parser)(simulation_config)
+
+
+class ComponentConfigurationParser:
+    """Parses component configuration and initializes components.
+
+    To define your own set of parsing rules, you should write a parser class that
+    inherits from this class and overrides ``parse_component_config``.  You can then
+    can then define a set of parsing rules that turn component configuration information
```
alec-deason: double 'can then'

vivarium/framework/plugins.py
```diff
@@ -0,0 +1,78 @@
...
+        return self._plugins[name]['builder_interface']
+
+    def get_optional_controllers(self):
+        return {name: self.get_plugin(name) for name in self._plugin_configuration['optional'].keys()}
+
+    def get_optional_interfaces(self):
+        return {name: self.get_plugin_interface(name) for name in self._plugin_configuration['optional'].keys()}
+
+    def _get(self, name):
+        fixture = self._lookup(name)
```
alec-deason: fixture probably isn't the right name any more?

vivarium/framework/plugins.py
```diff
@@ -0,0 +1,78 @@
...
+        try:
+            controller = import_by_path(fixture['controller'])(self._configuration)
+        except ValueError:
+            raise PluginConfigurationError(f'Invalid plugin specification {fixture["controller"]}')
+        if fixture['builder_interface'] is not None:
+            try:
+                interface = import_by_path(fixture['builder_interface'])(controller)
```
alec-deason: Will the builder_interface ever change independently of the controller? If not, would it make sense to drop this from the configuration and give the controller an optional `builder_interface` method or attribute which returns an instantiated object if it's called?

vivarium/interface/interactive.py
```diff
@@ -0,0 +1,120 @@
...
+    def setup(self):
+        super().setup()
+        self._start_time = self.clock.time
+
+    def initialize_simulants(self):
+        super().initialize_simulants()
+        self._initial_population = self.population.population
+
+    def reset(self):
+        # This is super crude, but should work for a great deal of components.
```
alec-deason: Yeah, it's real hard to check for internal state in components. It could actually keep a record of the initial component configuration and recreate the component manager from scratch.

vivarium/interface/interactive.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def run_until(self, end_time, with_logging=True):
+        if not isinstance(end_time, type(self.clock.time)):
+            raise ValueError(f"Provided time must be an instance of {type(self.clock.time)}")
+        iterations = int(ceil((end_time - self.clock.time)/self.clock.step_size))
+        self.take_steps(number_of_steps=iterations, with_logging=with_logging)
+        assert self.clock.time - self.clock.step_size < end_time <= self.clock.time
+        return iterations
+
+    def step(self, step_size=None):
```
alec-deason: We don't have any guarantees that changing step size on an already configured simulation will work correctly. It probably will but it might be better not to offer the possibility in the first place.

vivarium/interface/interactive.py
```diff
@@ -0,0 +1,120 @@
...
+    config.update(input_config)
+
+    simulation = InteractiveContext(config, components)
+    simulation.setup()
+    simulation.initialize_simulants()
+
+    return simulation
+
+
+def setup_simulation_from_model_specification(model_specification_file):
```
alec-deason: It might be nice to allow this to take a yaml string as well as a path. I suspect that the yaml string is the more useful option, but maybe not.

vivarium/framework/configuration.py
```diff
@@ -0,0 +1,52 @@
...
+    return config
+
+
+def _validate_model_specification_file(model_specification_file: str) -> None:
+    if not os.path.isfile(model_specification_file):
+        raise ConfigurationError('If you provide a model specification file, it must be a file. '
+                                 f'You provided {model_specification_file}')
+
+    extension = model_specification_file.split('.')[-1]
+    if extension not in ['yaml', 'yml']:
```
alec-deason: There is a more complete validation in utilities.py. Might make sense to call that here.

setup.py
```diff
@@ -6,11 +6,22 @@
     version='0.5',
     packages=find_packages(),
     include_package_data=True,
-    scripts=['simulate'],
+    entry_points="""
```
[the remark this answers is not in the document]
collijk: I took this from the click documentation on how to properly setup an entry point.  I'm unsure of why it works this way too, but also not familiar enough with setuptools.

vivarium/framework/engine.py
```diff
@@ -1,59 +1,62 @@
...
+        self.clock = self.plugin_manager.get_plugin('clock')
+
         self.values = ValuesManager()
         self.events = EventManager()
         self.population = PopulationManager()
         self.tables = InterpolatedDataManager()
         self.randomness = RandomnessManager()
 
+        self.builder = Builder(self)
+        self.builder.components = self.plugin_manager.get_plugin_interface('component_manager')
```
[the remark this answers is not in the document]
collijk: Need to talk in person about this.

vivarium/framework/plugins.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class PluginConfigurationError(VivariumError):
+    """Error raised when plugin configuration is incorrectly specified."""
+    pass
+
+
+class PluginManager:
+
+    def __init__(self, simulation_configuration, plugin_configuration=DEFAULT_PLUGINS['plugins']):
```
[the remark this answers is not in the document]
collijk: This supports the interactive context where that information is never passed into the simulation constructor (it generates a default plugin manager on the fly).

vivarium/framework/plugins.py
collijk: The controller very well might change independently of the builder interface.  See `time.py` for an example.  I've been thinking of the interfaces as relatively static with the controller underneath changing.

vivarium/interface/interactive.py
collijk: I'm considering many of these features in essentially an alpha state of development.  This seems useful.  I don't want to jump into the more heavy weight solution of essentially rebuilding the simulation without some more thought.

I want some user testing to see how it breaks.  I think it's worth pushing against components maintaining state info outside the state table.  Can we categorize and generalize that behavior?

vivarium/interface/interactive.py
collijk: Do you mean correctly in terms of implementation or correctly in terms of the results being correct?

vivarium/interface/interactive.py
collijk: Do you have a specific use case in mind?  This is the feature as abie requested it.

PR #23, opened 2018-05-29
ENH: make build_simulation_configuration and setup_simulation available at top level of module

aflaxman:


collijk:
Being handled in https://github.com/ihmeuw/vivarium/pull/31

PR #29, opened 2018-06-14
ENH: add .run method to to interactive context

aflaxman:


vivarium/interface/interactive.py
```diff
@@ -26,6 +26,13 @@ def reset(self):
         self.population._population = self._initial_population
         self.clock._time = self._start_time
 
+    def run(self, with_logging=True):
```
collijk: just need 
```
def run(self, with_logging=True):
    return self.run_until(self.clock.stop_time, with_logging)
```

PR #30, opened 2018-06-14
 ENH: add load_model_specification method to vivarium.interface.interactive

aflaxman:


collijk:
Being handled in https://github.com/ihmeuw/vivarium/pull/31

PR #31, opened 2018-06-18
Made the simulation easily accessible before the setup lifecycle phase

collijk:


aflaxman:
I like this all.  Should we push ourselves to include more docs and tests before making small changes like this?  Or merge this now and come back for docs/tests?

collijk:
This is the first thing I plan on coming back for in the narrative documentation.  It's super important, but so are several other things right now.

aflaxman:
👍

vivarium/interface/interactive.py
```diff
@@ -61,52 +86,75 @@ def take_steps(self, number_of_steps=1, step_size=None, with_logging=True):
                 self.step(step_size)
 
     def list_values(self):
+        if not self._setup:
```
alec-deason: This is repeated enough times that it might be worth building a simple validation decorator. Something like:
```
def requires_setup(func):
  def inner(self, *args, **kwargs):
    if not self._setup:
      raise InteractiveError("Value pipeline configuration is not complete until the simulation is setup.")
￼    return func(self, *args, **kwargs)
  return inner
```

PR #32, opened 2018-06-19
ENH: expose names of model components if they are available

aflaxman:


collijk:
I want this feature too, but it's quite tricky to implement correctly.  It's one of the reasons I did so much work on the component manager.  You'll get something very similar along side the RF correlation work.

PR #36, opened 2018-07-22
Feature/simple disease example

collijk:
This is essentially a no data version of our disease model with a handful of modifications.  I've validated that the components work as expected.  I've also looked at counterfactual scenarios to make sure the trends seem right (this by including the population level deletion with csmrs or pafs but not the individual level addition with excess mortalities or rrs).  For the included configuration:

| Scenario                                                               | YLLs  | YLLs (counterfactual) |
|------------------------------------------------------------------------|-------|-----------------------|
| Only Population and Mortality                                          | 31865 | N/A                   |
| Population, Mortality, Disease                                         | 32785 | 28584                 |
| Population, Mortality, Disease,  Risk on Incidence                     | 32969 | 30499                 |
| Population, Mortality, Disease, Risk on Excess Mortality               | 32957 | 29948                 |
| Population, Mortality, Disease, Risk on Incidence and Excess Mortality | 34984 | 29133                 |
| Row 6 + Treatment                                                      | 31725 | N/A                   |

Using both risk effects seems to have some covariance effect (a think to keep in mind in actual model development), but the purpose of this demo is not necessarily getting the math methods correct, just the coding implementation.

This needs both docstrings and a narrative tutorial, but I want to get this in first as a installation verification example so I can get `pip` and `conda` installation working.

Also included, some additional logging information about pipelines that should make debugging naming issues significantly easier.


vivarium/examples/disease_model/population.py
```diff
@@ -0,0 +1,129 @@
...
+
+        self.population_view.update(population)
+
+    def age_simulants(self, event):
+        population = self.population_view.get(event.index, query="alive == 'alive'")
+        population['age'] += event.step_size / pd.Timedelta(days=365)
+        self.population_view.update(population)
+
+
+class BasePopulation(NonCRNBasePopulation):
```
aflaxman: Why have two classes and inheritance here?  Would one class be sufficient?

collijk: They're both here so I can write a demo about the effects of CRN on variance in the system.  Do you think this is the wrong place for that?

collijk: Actually, I have an idea for this.  Update forthcoming.

PR #42, opened 2018-09-04
Bugfix/population dependencies

halfhorst:
This addresses the ambiguity of the PopulationError thrown when the pop initializers can't be ordered. It now distinguishes between cyclic dependencies and a missing component. I added unit tests for these, as well as a test for the other error case, which is multiple components trying to generate the same column.

I chose to piggy-back on the loop already being run to generate a list of columns that would be created/required, which results in potentially extra extensions to the lists. Alternative would be to make one pass at the beginning to populate those lists, or in the exception branch. 

Regarding the comment above the loop: isn't the algorithm listed n!, because worst case is full iteration of the list of initializers that decreases by one after each iteration (reverse order)? Rather than iterating the full list and then failing next time (n ** 2)

### DIFF
--- a/src/vivarium/framework/population.py
+++ b/src/vivarium/framework/population.py
@@ -207,11 +207,27 @@ class PopulationManager:
         status = pd.Series(True, index=pop_data.index)
         self.get_view(['tracked']).update(status)
 
+
+    @staticmethod
+    def _validate_no_missing_initializers(initializers):
+        created_columns = []
+        required_columns = []
+        for _, created, required in initializers:
+            created_columns.extend(created)
+            required_columns.extend(required)
+
+        if not set(required_columns) <= set(created_columns):
+            raise PopulationError(f"The initializers {initializers} could not be added.  "
+                                  "Check for missing dependencies in your components.")
+
+
     def _order_initializers(self) -> None:
         unordered_initializers = deque(self._population_initializers)
         starting_length = -1
         available_columns = []
         self._population_initializers = []
+
+        self._validate_no_missing_initializers(unordered_initializers)
 
         # This is the brute force N^2 way because constructing a dependency graph is work
         # and in practice this should run in about order N time due to the way dependencies are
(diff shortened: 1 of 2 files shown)


src/vivarium/framework/population.py
```diff
@@ -220,15 +222,22 @@ def _order_initializers(self) -> None:
...
                 if set(columns_required) <= set(available_columns):
                     self._population_initializers.append((initializer, columns_created, columns_required))
                     available_columns.extend(columns_created)
                 else:
                     unordered_initializers.appendleft((initializer, columns_created, columns_required))
 
         if unordered_initializers:
-            raise PopulationError(f"The initializers {unordered_initializers} could not be added.  "
-                                  f"Check for cyclic dependencies in your components.")
+
```
collijk: You typically want to create variables as close as possible to where you're using them.  Try:

```
if unordered_initializers:
    _, initializer_required, initializer_created = zip(*unordered_initializers)
    if set(initializer_required)...
```

halfhorst: when comparing requires and creates, can I safely ignore tracked?

src/vivarium/framework/population.py
```diff
@@ -220,15 +222,22 @@ def _order_initializers(self) -> None:
             starting_length = len(unordered_initializers)
             for _ in range(len(unordered_initializers)):
                 initializer, columns_created, columns_required = unordered_initializers.pop()
+                initializer_created.extend(columns_created)
+                initializer_required.extend(columns_required)
                 if set(columns_required) <= set(available_columns):
```
halfhorst: @collijk  Actually, I don't see how this was ever true, because required contains 'tracked' and, to the best of my knowledge, creates never does? It must be that tracked gets created ?

halfhorst: NVM

src/vivarium/framework/population.py
```diff
@@ -219,16 +219,27 @@ def _order_initializers(self) -> None:
         while len(unordered_initializers) != starting_length:
             starting_length = len(unordered_initializers)
             for _ in range(len(unordered_initializers)):
-                initializer, columns_created, columns_required = unordered_initializers.pop()
+                initializer, columns_created, columns_required = unordered_initializers.popleft()
```
halfhorst: the first _pop_initializer must always be ordered first, because it makes tracked and requires nothing, and everything else requires tracked. So, I flipped how we iterate over the unordered_initializers so that the first element is actually grabbed first by default

collijk: That's also worth writing a test around.

PR #43, opened 2018-09-07
Switch metrics to long storage format to overcome column limitation

collijk:


halfhorst:
Out of curiosity, what's the limitation ?

collijk:
It has to do with the size in bytes of the columns, so there's no hard limit of the number of columns.  `tables` stores the column names as metadata.  It's just some number of kB (here: https://github.com/pandas-dev/pandas/issues/7653 says 64, but that seems low).

PR #45, opened 2018-09-21
Bugfix/population dependencies

halfhorst:
Made the initializer test graphs slightly more interesting and added a test for a graph containing both circular and missing dependencies. They all currently pass


halfhorst:
Going to edit this

PR #48, opened 2018-10-04
Refactor lookup

kfwilson:
Refactoring lookup. Changes as follows:

In lookup.py:
- address the ToDos from the design meeting, including removing the option for callable data and lazy generation of interpolation and verifying that return types are consistent
- refactor so that the manager always returns the same type (a LookupTable) rather than sometimes a ScalarView and sometimes an InterpolatedTableView
- allow for lists/tuples to be passed to build scalar tables
- naming changes for consistency & clarity
- more documentation

In test_lookup.py:
- updating tests that use the manager build_table function to include the added value_columns parameter
- adding tests for scalar table and additional tests for interpolated table

In plugins.py:
- changing controller name for lookup to reflect naming changes in lookup.py

In test_engine.py:
- updating tests that use the manager build_table function to include the added value_columns parameter

All smoke tests pass with these changes. All tests in vivarium_inputs and vivarium_public_health (once updated - see refactor/lookup pr in vivarium_public_health) also pass.

### DIFF
--- a/src/vivarium/framework/lookup.py
+++ b/src/vivarium/framework/lookup.py
@@ -21,12 +21,7 @@ class InterpolatedTableView:
     -----
     These cannot be created directly. Use the `lookup` method on the builder during setup.
     """
-    # FIXME: take out default for clock
-    def __init__(self, data, population_view, key_columns, parameter_columns, interpolation_order, clock=None):
-        # FIXME: validation should be in manager
-        if data is None or (isinstance(data, pd.DataFrame) and data.empty):
-            raise ValueError("Must supply some data")
-
+    def __init__(self, data, population_view, key_columns, parameter_columns, interpolation_order, clock):
         self._data = data
         # TODO: figure out if we need interpolation to be lazily generated
         self._interpolation = None
@@ -46,7 +41,7 @@ class InterpolatedTableView:
 
             if isinstance(data, Interpolation):
                 self._interpolation = data
-                warnings.warn("Creating lookup tables from pre-initialized Intrepolation objects is deprecated. "
+                warnings.warn("Creating lookup tables from pre-initialized Interpolation objects is deprecated. "
                               "Use key_columns and parameter_columns to control interpolation. If that isn't possible "
                               "then please raise an issue with your use case.", DeprecationWarning)
             else:
@@ -56,14 +51,13 @@ class InterpolatedTableView:
 
     def __call__(self, index):
         pop = self.population_view.get(index)
-        # FIXME: this check should be if year is in param_columns
-        if self.clock:
+        if 'year' in self._parameter_columns:
             current_time = self.clock()
             fractional_year = current_time.year
             fractional_year += current_time.timetuple().tm_yday / 365.25
             pop['year'] = fractional_year
 
-        return self.interpolation(pop)
+        return self.interpolation(pop)  # a series if only one column
 
     def __repr__(self):
         return "InterpolatedTableView()"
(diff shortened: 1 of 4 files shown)


src/vivarium/framework/lookup.py
```diff
@@ -8,92 +8,131 @@
...
+        if isinstance(data, (Number, datetime, timedelta, list, tuple)):
+            self.table = ScalarTable(data, value_columns)
+
+        elif isinstance(data, pd.DataFrame):
+            if set(key_columns).intersection(set(parameter_columns)):
+                raise ValueError(f'There should be no overlap between key columns: {key_columns} '
+                                 f'and parameter columns: {parameter_columns}.')
+
+            view_columns = sorted((set(key_columns) | set(parameter_columns)) - {'year'})
+            self.table = InterpolatedTable(data, population_view(view_columns), key_columns,
```
collijk: The table here, on the other hand, should be private.  The `LookupTable` instance is the user's interface into this system.  We don't want them digging around in the stuff underneath.

src/vivarium/framework/lookup.py
```diff
@@ -110,21 +149,22 @@ def setup(self, builder):
...
                              f'You specified {self._interpolation_order}')
 
-    def build_table(self, data, key_columns, parameter_columns):
-        """Construct a TableView from a ``pandas.DataFrame``. An interpolation
-        function of the specified order will be calculated for each permutation
-        of the set of key_columns. The columns in parameter_columns will be used
-        as parameters for the interpolation functions which will estimate all
-        remaining columns in the table.
+    def build_table(self, data, key_columns, parameter_columns, value_columns):
+        """Construct a LookupTable from input data.
```
collijk: This documentation should really be on the ``LookupTableInterface`` since that's the public interface.

src/vivarium/framework/lookup.py
```diff
@@ -1,68 +1,71 @@
...
+class InterpolatedTable:
     """A callable that returns the result of an interpolation function over input data.
 
-    Parameters
+    Attributes
     ----------
-    interpolation : callable
-    population_view : `vivarium.framework.population.PopulationView`
-    clock : callable
+    data : `pandas.DataFrame`
```
collijk: The typing in the signature is sufficient.  The autodoc tools (and more importantly, most editor static analysis tools) pick it up from there.

src/vivarium/framework/lookup.py
```diff
@@ -1,68 +1,71 @@
...
 from datetime import datetime, timedelta
-import warnings
+from typing import Union, List, Tuple, Callable, TypeVar
 
 import pandas as pd
 
 from vivarium.interpolation import Interpolation
+from vivarium.framework.population import PopulationView
 
+value_type = TypeVar('value_type', Number, timedelta, datetime)
```
collijk: I haven't seen TypeVar!  I've been using `Union` to construct types.  This seems better!  

I recommend giving it a CamelCase name like `ScalarValue`.  Lowercase stuff used in typing usually refer to python standard types or to numpy dtypes (meaning they likely are thin wrappers around C types), at least from what I've seen.  

`datetime` and `timedelta` being immediate exceptions, of course.  I think those just got named before conventions were standardized and are impossible to change because dealing with time is the worst.

src/vivarium/framework/lookup.py
```diff
@@ -71,29 +74,136 @@ def __call__(self, index):
...
+    def __init__(self, values: Union[List[value_type], Tuple[value_type]],
+                 value_columns: Union[List[str], Tuple[str]]):
 
-class ScalarView(TableView):
-    def __init__(self, value):
-        self.value = value
+        self.values = values
+        self.value_columns = value_columns
 
     def __call__(self, index):
```
collijk: Document return types like 
```
def __call__(self, index) -> pd.DataFrame:
```

PR #49, opened 2018-10-06
raise an error if multiple components try to set the default config

michaella21:
This PR adds a function to raise an error when there are multiple components have different default configurations for the same key. Our default configurations can be set at the 4 different layers(base, component_configs, model_override (from yaml), override (usually for tests)) and higher layer can override the lower layer configurations. Since we have many different components and we cannot guarantee which component is set up first, we should not allow the multiple components set the default value for the same configuration key. 

The following `vivarium_public_health` PR is just to adjust the some default configuration values which do not follow this rule.

### DIFF
--- a/src/vivarium/framework/components/manager.py
+++ b/src/vivarium/framework/components/manager.py
@@ -123,11 +123,11 @@ def _check_duplicated_default_configuration(component, config):
         key = overlapped.pop()
         if not isinstance(component[key], dict) and not isinstance(config[key], ConfigTree):
             try:
-                if component[key] != config.get_from_layer(key, layer='component_configs'):
+                if config.get_from_layer(key, layer='component_configs'):
                     raise ComponentConfigError(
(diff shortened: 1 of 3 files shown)


src/vivarium/framework/components/manager.py
```diff
@@ -102,11 +104,39 @@ def _setup_components(builder, component_list, configuration):
...
+    while overlapped:
+        key = overlapped.pop()
+
+        try:
+            sub_config = config.get_from_layer(key, layer='component_configs')
+            sub_component = component[key]
+
+            if isinstance(sub_component, dict) and isinstance(sub_config, ConfigTree):
+                _check_duplicated_default_configuration(sub_component, sub_config, source)
+            elif isinstance(sub_component, dict) or isinstance(sub_config, ConfigTree):
```
halfhorst: I don't quite understand this elif block. Why does this imply they have different structures?

PR #52, opened 2018-10-15
Feature/order 0 interp

kfwilson:
Probably easier if you review the refactor PR first, which cleans up the code a bunch. This PR is just the added order 0/bin edges functionality. 

Adds order 0 interpolation with bin edges for any number of parameters. Requires bin edges for all continuous parameters. Uses start of bin as inclusive and end as exclusive (which is how gbd age bins are constructed). 

For now only allows order 0 interpolation (order 1 hasn't been updated to use bin edges).

Adds an 'extrapolate' option to the interpolation config that allows constant (for order 0) extrapolation beyond bin edges.

All smoke tests run with interpolation order set to 0. See attached files for how changing order to 0 changes incidence rate pipeline in sim. 
incidence_rate_pipeline_order0
incidence_rate_pipeline_order1

### DIFF
--- a/src/vivarium/interpolation.py
+++ b/src/vivarium/interpolation.py
@@ -139,7 +139,7 @@ def validate_parameters(data, categorical_parameters, continuous_parameters, ord
 
     out = []
     for p in continuous_parameters:
-        if not isinstance(p, (List, Tuple)) and len(p) != 3:
+        if not isinstance(p, (List, Tuple)) or len(p) != 3:
             raise ValueError(f'Interpolation is only supported for binned data. You must specify a list or tuple '
                              f'containing, in order, the column name used when interpolation is called, '
                              f'the column name for the left edge (inclusive), and the column name for '
--- a/src/vivarium/interpolation.py
+++ b/src/vivarium/interpolation.py
@@ -2,40 +2,36 @@ import warnings
 
 import pandas as pd
 from scipy import interpolate
-
+from typing import Sequence
 
 class Interpolation:
-    def __init__(self, data, categorical_parameters, continuous_parameters, order, func=None):
-        data = data
+    def __init__(self, data: pd.DataFrame, categorical_parameters: Sequence[str],
+                 continuous_parameters: Sequence[str], order: int):
+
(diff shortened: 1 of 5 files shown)


halfhorst:
I think you've mentioned this before but I want to make sure: You've tested this on bins of differing sizes ? I don't see a unit test for it

src/vivarium/interpolation.py
```diff
@@ -1,120 +1,324 @@
...
+            Order of interpolation.
+        """
 
-        if len(self.parameter_columns) not in [1, 2]:
-            raise ValueError("Only interpolation over 1 or 2 variables is supported")
+    def __init__(self, data: pd.DataFrame, categorical_parameters: Union[List[str], Tuple[str]],
+                 continuous_parameters: ParameterType, order: int, extrapolate: bool):
+        # TODO: allow for order 1 interpolation with binned edges
+        if order != 0:
+            raise NotImplementedError(f'Interpolation is only supported for order 0. You specified order {order}')
```
michaella21: I'm a little bit confused with this line. Kate explained why she kept it here. I'm just leaving a comment in case that it's not just me who get confused by this check.

kfwilson:
> I think you've mentioned this before but I want to make sure: You've tested this on bins of differing sizes ? I don't see a unit test for it. Otherwise it looks good. I'm relying a bit on your tests and figures to eval, this is a big PR!

Very good point. I did test that by playing around with it but I didn't add an explicit unit test for different bin sizes. Adding that now.

src/vivarium/interpolation.py
kfwilson: Yeah this is probably sufficient without the additional check in validate_params for order in [0, 1]. Basically, what I was thinking was that while this is currently redundant, at some point, we'll want to remove this restriction that order has to be 0, at which point, we'll want the later check (which is what was originally in the interpolation).

kfwilson: I got rid of the later redundant check + all the other dead order 1 code so it should be cleaner and less confusing now

PR #54, opened 2018-10-24
Feature/extend interactive context

kfwilson:
Adding an add components method to InteractiveContext that Abie had patched in in ceam experiments

Adding type hints because I got a little confused trying to extend these in public health when I thought plugin config could be more types than it is

### DIFF
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -6,6 +6,7 @@ from time import time
 
 import pandas as pd
 
+from vivarium import VivariumError
 from vivarium.framework.configuration import build_model_specification
 from .components import ComponentInterface
 from .event import Event, EventInterface
@@ -26,6 +27,7 @@ class SimulationContext:
     def __init__(self, configuration, components, plugin_manager=None):
         plugin_manager = plugin_manager if plugin_manager else PluginManager()
 
+        self._setup = False
         self.configuration = configuration
         self.builder = Builder(configuration, plugin_manager)
 
@@ -57,6 +59,7 @@ class SimulationContext:
         self.time_step_emitters = {k: self.builder.event.get_emitter(k) for k in self.time_step_events}
         self.end_emitter = self.builder.event.get_emitter('simulation_end')
         self.builder.event.get_emitter('post_setup')(None)
+        self._setup = True
 
     def step(self):
         _log.debug(self.clock.time)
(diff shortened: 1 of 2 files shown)


src/vivarium/framework/engine.py
```diff
@@ -57,6 +59,7 @@ def setup(self):
         self.time_step_emitters = {k: self.builder.event.get_emitter(k) for k in self.time_step_events}
         self.end_emitter = self.builder.event.get_emitter('simulation_end')
         self.builder.event.get_emitter('post_setup')(None)
+        self._setup = True
```
collijk: Move this to before you emit post setup so components don't try to do anything weird during that lifecycle phase.

src/vivarium/framework/engine.py
```diff
@@ -79,6 +82,11 @@ def finalize(self):
     def report(self):
         return self.values.get_value('metrics')(self.population.population.index)
 
+    def add_components(self, component_list):
```
collijk: Replace the call on line 51 with a call to this method.

src/vivarium/interface/interactive.py
```diff
@@ -116,22 +117,24 @@ def replace_component(self, old_component, new_component):
...
+                          plugin_config: Mapping=None) -> InteractiveContext:
     config = build_simulation_configuration()
     config.update(input_config)
     plugin_manager = PluginManager(plugin_config)
 
     return InteractiveContext(config, components, plugin_manager)
 
 
-def setup_simulation(components, input_config=None, plugin_config=None):
+def setup_simulation(components: List, input_config: Union[Mapping, ConfigTree]=None,
```
collijk: The ConfigTree is a Mapping

src/vivarium/interface/interactive.py
```diff
@@ -116,22 +117,24 @@ def replace_component(self, old_component, new_component):
...
+                     plugin_config: Mapping=None) -> InteractiveContext:
     simulation = initialize_simulation(components, input_config, plugin_config)
     simulation.setup()
 
     return simulation
 
 
-def initialize_simulation_from_model_specification(model_specification_file):
+def initialize_simulation_from_model_specification(model_specification_file: str) -> InteractiveContext:
```
collijk: We should probably only have one place that returns an actual Simulation context.  Drop line 144 and instead return `initialize_simulation(components, simulation_config, plugin_config)`

kfwilson: I need the plugin manager to get the parser for components though.

collijk: Gross.

PR #59, opened 2018-12-12
Feature/separate component config setup

halfhorst:
This separates configuration of top-level components from all calls to setup and any calls to configuration of lower-level components. It's the first step towards a dependency manager.

Passes all the smoke tests (minus Sanofi) and all unit tests.

### DIFF
--- a/src/vivarium/framework/components/manager.py
+++ b/src/vivarium/framework/components/manager.py
@@ -85,23 +85,28 @@ def _setup_components(builder, component_list: Sequence, configuration: ConfigTr
 
     configured = []
     for c in component_list:  # apply top-level configurations first
-        if hasattr(c, "configuration_defaults"):
+        if hasattr(c, "configuration_defaults") and not c in configured:
             _apply_component_default_configuration(configuration, c)
-        configured.append(c)
+            configured.append(c)
 
     setup = []
     while component_list:  # mutated at runtime by calls to setup
         c = component_list.pop(0)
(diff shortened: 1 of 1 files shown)


src/vivarium/framework/components/manager.py
```diff
@@ -81,26 +81,34 @@ def get_components(self, component_type: str):
...
+def _setup_components(builder, component_list: Sequence, configuration: ConfigTree) -> Sequence:
+
+    configured = []
+    for c in component_list:  # apply top-level configurations first
+        if hasattr(c, "configuration_defaults") and not c in configured:
+            _apply_component_default_configuration(configuration, c)
+            configured.append(c)
+
+    setup = []
+    while component_list:  # mutated at runtime by calls to setup
```
kfwilson: thanks for adding this comment! was confused why you had to keep track of what had already been setup

src/vivarium/framework/components/manager.py
```diff
@@ -81,26 +81,34 @@ def get_components(self, component_type: str):
...
-    return done
+
+        if hasattr(c, "configuration_defaults") and c not in configured:
+            _apply_component_default_configuration(configuration, c)
+            configured.append(c)
+
+        if c not in setup:
+            if hasattr(c, "setup"):
+                result = c.setup(builder)
+                if result is not None:
```
kfwilson: Do we still need this based on the TODO in the old version?

PR #61, opened 2018-12-17
Add profile command

collijk:


src/vivarium/interface/cli.py
```diff
@@ -64,3 +65,17 @@ def test():
         raise
     finally:
         shutil.rmtree(results_directory, ignore_errors=True)
+
+
+@simulate.command()
+@click.argument('model_specification', callback=verify_yaml,
+                type=click.Path(exists=True, dir_okay=False, resolve_path=True))
+@click.option('--results_directory', '-o', type=click.Path(resolve_path=True),
```
kfwilson: Is there a reason why we don't have the verbose or log to file options here? I always like to be able to check the progress on something but would those options mess with the profiling?

collijk: Mostly because I wanted to run a profile but didn't want to take the time to implement all the bells and whistles.  PR additional features you find useful.  But do it when you run a profile and want to see more stuff.

PR #62, opened 2018-12-19
Bugfix/pipeline not filter untracked simulants

michaella21:
Only two main changes (and each of them is 1-2 lines). 
1. In `lookup.py`, we do not filter down to the tracked people only any more. Always return all the simulants.
2. We do not have `simulation.population.population` any more. If a user wants to call all untracked and tracked simulants, should use `simulation.get_population(untracked=True)`. (otherwise simulation.get_population() only returns the tracked simulants) 

The other changes are  all according to the change in 2.

### DIFF
--- a/src/vivarium/framework/population.py
+++ b/src/vivarium/framework/population.py
@@ -42,7 +42,7 @@ class PopulationView:
     @property
     def columns(self) -> List[str]:
         if not self._columns:
-            return list(self.manager.population.columns)
+            return list(self.manager._population.columns)
         return list(self._columns)
 
     def subview(self, columns: Sequence[str]) -> 'PopulationView':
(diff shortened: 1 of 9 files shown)


src/vivarium/framework/population.py
```diff
@@ -279,6 +279,12 @@ def metrics(self, index, metrics):
         metrics['total_population'] = len(untracked)+len(tracked)
         return metrics
 
+    def get_population(self, untracked) -> pd.DataFrame:
```
kfwilson: do we want a deafult on this so it filters out untracked unless you specifically request otherwise?

collijk: Not in the population manager.  Should be explicit here and have the sensible default at the interactive interface level

PR #63, opened 2019-01-03
Make error message clearer

collijk:
The previous error message dumped a bunch of info about initializers but didn't state what was missing clearly.

### DIFF
--- a/src/vivarium/framework/population.py
+++ b/src/vivarium/framework/population.py
@@ -220,8 +220,9 @@ class PopulationManager:
             required_columns.extend(required)
 
         if not set(required_columns) <= set(created_columns):
-            raise PopulationError(f"The initializers {initializers} could not be added.  "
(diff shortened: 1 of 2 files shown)


src/vivarium/framework/population.py
```diff
@@ -220,8 +220,9 @@ def _validate_no_missing_initializers(initializers: Sequence[Tuple]) -> None:
             required_columns.extend(required)
 
         if not set(required_columns) <= set(created_columns):
-            raise PopulationError(f"The initializers {initializers} could not be added.  "
-                                  "Check for missing dependencies in your components.")
+            missing_columns = set(required_columns).difference(set(created_columns))
```
kfwilson: maybe make this missing_columns set and then use it in the check (e.g., if missing_columns) so you don't have to do the set comparison twice?

PR #64, opened 2019-01-03
Feature/ci updates

collijk:
Making sure doctests run successfully, ditch requirements.txt


collijk:
@halfhorst Yeah, the doctests were broken.  The ci runs them now so it shouldn't happen again.

PR #65, opened 2019-02-13
switch decision points to a dict

halfhorst:
Here we alter the randomness manager to maintain a dict of streams rather than a set. We can use that dict to return an existing stream by key if explicitly told not to throw an error

### DIFF
--- a/src/vivarium/framework/randomness.py
+++ b/src/vivarium/framework/randomness.py
@@ -627,8 +627,7 @@ class RandomnessManager:
         pop_size = builder.configuration.population.population_size
         self._key_mapping.map_size = max(map_size, 10*pop_size)
 
-    def get_randomness_stream(self, decision_point: str, for_initialization: bool=False,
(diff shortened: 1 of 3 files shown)


src/vivarium/framework/randomness.py
```diff
@@ -649,11 +650,16 @@ def get_randomness_stream(self, decision_point: str, for_initialization: bool=Fa
             with the same identifier.
         """
         if decision_point in self._decision_points:
-            raise RandomnessError(f"Two separate places are attempting to create "
-                                  f"the same randomness stream for {decision_point}")
-        self._decision_points.add(decision_point)
-        return RandomnessStream(key=decision_point, clock=self._clock, seed=self._seed,
-                                index_map=self._key_mapping, manager=self, for_initialization=for_initialization)
+            if throw_error:
```
collijk: Drop the throw error (revert to the old code) with just the change to a dict instead of a set.

PR #71, opened 2019-04-23
Feature/interactive docstrings

halfhorst:
Docstrings for all the relevant things in the interactive module

### DIFF
--- a/src/vivarium/interface/interactive.py
+++ b/src/vivarium/interface/interactive.py
@@ -6,7 +6,7 @@ import pandas as pd
 
 from vivarium.framework.configuration import build_simulation_configuration, build_model_specification
 from vivarium.framework.plugins import PluginManager
-from vivarium.framework.time import _Timedelta, _Time
+from vivarium.framework.time import Timedelta, Time
 from vivarium.framework.values import Pipeline
 from vivarium.framework.engine import SimulationContext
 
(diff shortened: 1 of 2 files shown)


src/vivarium/interface/interactive.py
```diff
@@ -1,45 +1,99 @@
...
+        warnings.warn("This reset method is very crude.  It should work for "
+                      "many simple simulations, but we make no guarantees. In "
+                      "particular, if you have components that manage their "
+                      "own state in any way, this might not work.")
         self.population._population = self._initial_population
         self.clock._time = self._start_time
 
     @raise_if_not_setup(system_type='run')
-    def run(self, with_logging=True):
+    def run(self, with_logging: bool=True) -> int:
```
kfwilson: should we also type hint None return types or do we not care as much about that?

src/vivarium/interface/interactive.py
```diff
@@ -1,45 +1,99 @@
...
         super().setup()
         self._start_time = self.clock.time
         self.initialize_simulants()
 
     def initialize_simulants(self):
+        """Initialize this simulation's population. Should not be called directly."""
         super().initialize_simulants()
         self._initial_population = self.population.get_population(True)
 
     def reset(self):
```
kfwilson: Not technically related to doc strings but noticed while I was looking over things so mentioning so I don't forget: shouldn't we also raise if not setup on the reset method since we won't have self._start_time until after setup has been called?

src/vivarium/interface/interactive.py
halfhorst: I vote for no

src/vivarium/interface/interactive.py
halfhorst: yes I think that is right. I will open a 1 pt ticket for it since it is out of scope here

PR #73, opened 2019-04-25
Docs/simulate run update

collijk:


docs/source/concepts/lifecycle.rst
```diff
@@ -1,22 +1,148 @@
...
+--------------
+
+The initialization phase of ``vivarium`` lasts from when a user first interacts
+with a simulation :ref:`entry point <entry_points_concept>` to when the
+``__init__`` method of the :class:`vivarium.framework.engine.SimulationContext`
+completes.  In it, the following things happen (roughly in this order):
+
+1. Simulation outputs are configured. This is usually just setting up some
+   directories for the final results and the simulation logs.
+2. :term:`Model specification <Model Specification>` defaults are collected and
```
kfwilson: Is it worth clarifying here that only the configuration defaults + configuration block of the model spec file are compiled into the ConfigTree not the whole model spec?

docs/source/concepts/lifecycle.rst
```diff
@@ -1,22 +1,148 @@
...
+   compiled along with the model specification file (if provided) into a
+   :class:`vivarium.config_tree.ConfigTree` object.
+3. A :class:`vivarium.framework.plugins.PluginManager` is generated around the
+   :term:`plugins <Plugin>` section of of the model specification.  The plugin
+   manager is responsible for parsing the plugin section and instantiating
+   plugin controllers and interfaces for the framework.
+4. The :class:`vivarium.framework.components.ComponentConfigurationParser` is
+   created to parse and instantiate all standard simulation
+   :term:`components <Component>`, if not provided directly by the user in
```
kfwilson: this is a little confusing - is this saying if the components aren't provided directly by the user or if the ComponentConfigurationParser is not provided by the user?

docs/source/glossary.rst
```diff
@@ -6,15 +6,52 @@ Glossary
...
+        position.
+
+    Component
+        Any self-contained piece of code that can be plugged into the
+        simulation to add some functionality. In ``vivarium`` we typically
+        think of components as encapsulating and managing some behavior or
+        attributes of the :term:`simulants <Simulant>`.
+
+    Configuration
+        The configuration is a a set of parameters a user can modify to
```
kfwilson: Hmm is this definition accurate? Because configuration parameters affect more than just components (specifically, the time block affects the framework itself, not any component). Maybe that's more detail than we care about here.

docs/source/tutorials/running_a_simulation/cli.rst
```diff
@@ -0,0 +1,120 @@
...
+If you navigate to that directory, you should see a subdirectory with the name
+of your model specification.  Inside the model specification directory, there
+will be another subdirectory named for the start time of the run. In here, you
+should see two hdf files: ``final_state.hdf``, which is the population
+:term:`state table <State Table>` at the end of the simulation, and
+``output.hdf``, which is the results of the :term:`metrics <Metrics>` generated
+by the simulation.
+
+For example, say we've run a simulation for a model specification called
+``potatoes.yaml`` (maybe we're really into gardening).  Our directory tree
```
kfwilson: lol love the potatoes example

src/vivarium/config_tree.py
```diff
@@ -391,7 +391,7 @@ def _loads(self, data_string, layer=None, source=None):
         source : str
             Source to attribute the values to
         """
-        data_dict = yaml.load(data_string)
+        data_dict = yaml.full_load(data_string)
```
kfwilson: :raised_hands:

docs/source/tutorials/running_a_simulation/cli.rst
collijk: Have to keep @alec-deason 's legacy alive.

docs/source/concepts/lifecycle.rst
kfwilson: I like the clarification that component defaults aren't compiled here. But I'm still a little confused - maybe I'm misreading this but if the user doesn't supply a model specification file, where are the 'Model specification defaults' collected from?

collijk: these come from the user's ``vivarium.yaml`` file if present.  I'll write more about it in the ``Configuration`` concept section and link that here when done.

docs/source/glossary.rst
collijk: More detail than we care about.

docs/source/concepts/lifecycle.rst
kfwilson: oh I forgot about `vivarium.yaml`

docs/source/concepts/lifecycle.rst
```diff
@@ -1,22 +1,152 @@
...
+At this point, all input arguments have been parsed and all top-level
+components have been instantiated.  This is a useful phase in the simulation
+lifecycle because you can typically modify what components are in the system
+or how they are configured without any consequences.
+
+Setup
+-----
+
+In this stage, the framework moves to setting up the
+:term:`components <Component>`. For each top-level component, the framework
```
halfhorst: I think top-level is not clear here to the uninitiated

collijk: Will address in a different section and provide a reference here.

PR #77, opened 2019-05-01
component config parser docs

kfwilson:


src/vivarium/framework/components/parser.py
```diff
@@ -132,5 +217,21 @@ def _clean_args(args, path):
...
+        A list of tuples representing components, where the first element of
+        each tuple is a string with the full import path to the component and
+        the component name and the second element is a tuple of the arguments to
+        the component.
+
+    Returns
+    -------
+        A list of instantiated component objects.
+
+    """
```
collijk: Can you use smarter tuple unpacking in this method.  e.g.

```
return [import_by_path(component)(*args) for component, args in component_list]
```

PR #78, opened 2019-05-02
Docs/component manager

halfhorst:
This one is just pure doc/module strings to guide a developer.


src/vivarium/framework/components/manager.py
```diff
@@ -16,47 +19,63 @@ class ComponentConfigError(VivariumError):
...
         Parameters
         ----------
         components:
-          Components to register
+          Instantiated components to register
         """
         self._add_components(self._components, components)
 
-    def _add_components(self, component_list, components):
+    def _add_components(self, component_list: Sequence, components: Sequence):
```
kfwilson: I don't think you want sequence here - that includes strings

src/vivarium/framework/components/manager.py
```diff
@@ -70,19 +89,54 @@ def setup_components(self, builder, configuration):
...
         self._component_manager = component_manager
 
     def add_components(self, components: Sequence):
+        """Register new components with the system. Components are setup after
+        managers.
+
+        Parameters
+        ----------
+        components:
+          Instantiated components to register
```
kfwilson: Ok I'm just going to be nitpicky here: do we want to end these parameter descriptions with a period or not? I vote period because that's what the numpydoc guide says but I don't feel super strongly as long as we're consistent

src/vivarium/framework/components/manager.py
```diff
@@ -1,8 +1,11 @@
-"""The ``vivarium`` component management system.
-
-This module contains the interface and the default implementation of the ``vivarium`` component manager.
-The manager is responsible for tracking all components in the system and for initiating the ``setup``
-life-cycle stage of each component.
+"""This module contains the interface and the default implementation of the
+``vivarium`` component manager system.
+
+The component manager is a plugin that manages all of the components in a
```
kfwilson: I don't know how I feel about describing the component manager as something that 'manages all of the components' - that feels like a bit of a circular definition. Maybe elaborate on what managing components means? You highlight specifically the ``setup`` phase - if managing just encompasses that, only say that. Otherwise, maybe give an example of what else managing entails?

src/vivarium/framework/components/manager.py
```diff
@@ -1,8 +1,11 @@
-"""The ``vivarium`` component management system.
-
-This module contains the interface and the default implementation of the ``vivarium`` component manager.
-The manager is responsible for tracking all components in the system and for initiating the ``setup``
-life-cycle stage of each component.
+"""This module contains the interface and the default implementation of the
```
collijk: We want to frame the two distinct behaviors of the component manager clearly:

1.  It maintains a registry of all components in the system.  In this role it allows other components to register new sub-components or to query for particular component types in the system.  
2. It manages the setup lifecycle phase for all components.

collijk: You can use full rst syntax here, And I think that's the way to go.  Start the docstring with an rst heading:
```
=============================
The Component Management System
=============================

Summary of what it does

contents tree

subsection on component registry behavior

Maybe sub-subsection on the distinction between normal components and managers from the perspective of the component manager.  To include: who can register components, who can register managers

subsection on the setup lifecycle
talk about application of config defaults
talk about the main setup loop

etc.

File: src/vivarium/framework/components/manager.py [hunk 5]
```diff
@@ -1,8 +1,11 @@
...
+``vivarium`` component manager system.
+
+The component manager is a plugin that manages all of the components in a
+simulation and is responsible for the ``setup`` phase of their lifecycle.  It
+maintains a distinction between the top-level (or "manager") components that are
```
collijk: We've been using "top-level" components to refer to those specified directly in the model specification.  We should be consistent.

src/vivarium/framework/components/manager.py
```diff
@@ -1,8 +1,11 @@
...
+core to the framework's operation and lower-level components that are
+simulation-specific. This is the only dependency management currently provided.
```
collijk: Don't mention dependency management here.  it's not applicable.

src/vivarium/framework/components/manager.py
collijk: Include links to relevant concept docs/glossary defs

src/vivarium/framework/components/manager.py
```diff
@@ -16,47 +19,63 @@ class ComponentConfigError(VivariumError):
 
 
 class ComponentManager:
-    """Handles the setup and access patterns for components in the system."""
+    """Handles tracking and access patterns for all components in a Vivarium
```
collijk: We have a mix of Vivarium and ``vivarium`` when refering to the framework.  I'm leaning towards the latter, and have been writing it as such in all my recent docs.  We should be consistent about it.

Also, try to give docstrings of the format:

```
"""Single line description.

More detailed description across
multiple lines.

Parameters
---------------
...
etc.
            <- Final newline before end of comment.
"""
```

src/vivarium/framework/components/manager.py
collijk: yep, that's my holdout from before I knew typing very well.  ``Union[List, Tuple]`` is what we want.

src/vivarium/framework/components/manager.py
```diff
@@ -70,19 +89,54 @@ def setup_components(self, builder, configuration):
 
 
 class ComponentInterface:
+    """The component manager interface made available from the builder."""
```
collijk: We want significantly more detail on the interfaces.  We'll collect the docstrings into the builder concept note to detail what services the builder provides.  The target of the docstrings for the interfaces is users and component developers, so it's not nitty gritty, but should provide at least a heuristic overview of what's happening. and when.

src/vivarium/framework/components/manager.py
collijk: Use periods!

src/vivarium/framework/components/manager.py
```diff
@@ -70,19 +89,54 @@ def setup_components(self, builder, configuration):
...
+        managers.
+
+        Parameters
+        ----------
+        components:
+          Instantiated components to register
+        """
         self._component_manager.add_components(components)
 
     def get_components(self, component_type: str):
```
collijk: Don't forget this one

src/vivarium/framework/components/manager.py
```diff
@@ -70,19 +89,54 @@ def setup_components(self, builder, configuration):
...
         return self._component_manager.get_components(component_type)
 
 
 def _setup_components(builder, component_list: Sequence, configuration: ConfigTree) -> Sequence:
-
+    """Helper method for configuring and setting up a list of components.
+
+    This function first loops over `component_list` and applies configuration
+    defaults if present.  The application entails updating the `configuration`
```
collijk: Double backtics on things you want to render as fixed-width

src/vivarium/framework/components/manager.py
```diff
@@ -70,19 +89,54 @@ def setup_components(self, builder, configuration):
...
+        """
         self._component_manager.add_components(components)
 
     def get_components(self, component_type: str):
         return self._component_manager.get_components(component_type)
 
 
 def _setup_components(builder, component_list: Sequence, configuration: ConfigTree) -> Sequence:
-
+    """Helper method for configuring and setting up a list of components.
```
collijk: Let's make all these methods public.  Otherwise they won't render in the api docs

src/vivarium/framework/components/manager.py
```diff
@@ -70,19 +89,54 @@ def setup_components(self, builder, configuration):
...
+
+    This function first loops over `component_list` and applies configuration
+    defaults if present.  The application entails updating the `configuration`
+    object. Then, the list is looped over until empty, with components being
+    popped and setup. Because of the side effect below, new components can be
+    added. This necessitates the while loop and also another check of
+    configuration.
+
+    Importantly, this function is called by the ComponentManager and it is
```
collijk: use rst references where possible (e.g. :class:`ComponentManager`).  Anything defined in the module only needs a short name for the link to work properly.

src/vivarium/framework/components/manager.py
```diff
@@ -108,7 +162,19 @@ def _setup_components(builder, component_list: Sequence, configuration: ConfigTr
     return setup
 
 
-def _apply_component_default_configuration(configuration, component):
+def _apply_component_default_configuration(configuration: ConfigTree, component: Any):
+    """Checks if a default configuration attached to a component is duplicated in
+    the current config tree, which will raise. If it is not, the configuration
```
collijk: simulation configuration, not config tree

src/vivarium/framework/components/manager.py
halfhorst: This isn't quite right, because we are ok with a singular component, but we can't really type check that :-/

src/vivarium/framework/components/manager.py
halfhorst: I nabbed everything in this module, but I am hesitant to go further. Maybe in the module level docstring. But references on classes to, say, lifecycle stages -- if we make it an rst reference it will render nice in sphinx but on the command line you'd see the whole rst syntax, right ?

PR #81, opened 2019-05-14
add component names

halfhorst:
This is mostly one line names and editing / adding reprs and strs. I wanted everything to have a repr so at minimum prints look nice. I added str where the repr may be verbose, and I left out dataframes and big things.

Names are protected behind property attribute functions. I removed the key kwarg functionality. A state id is expected to be unique, and that is used to identify the transition set that it holds.

When making these changes I altered the component manager to print all components and look for duplicates. I currently don't get any.  Attached is some printed output of a bunch of component names from bmi and bfp_eggs

below is an example of what the names look like. They are decently readable
[bmi_names.txt](https://github.com/ihmeuw/vivarium/files/3204056/bmi_names.txt)


src/vivarium/framework/state_machine.py
```diff
@@ -354,10 +356,10 @@ def __len__(self):
         return len(self.transitions)
 
     def __str__(self):
-        return str([str(x) for x in self.transitions])
+        return "TransitionSet()"
```
kfwilson: do we at least want to include the name in str representations where we may have multiple of the objects?

src/vivarium/framework/components/manager.py
```diff
@@ -106,6 +107,11 @@ def setup_components(self, builder, configuration: ConfigTree):
         self._managers = setup_components(builder, self._managers, configuration)
         self._components = setup_components(builder, self._components, configuration)
 
+    def __str__(self):
+        return "ComponentManager()"
+
+    def __repr__(self):
+        return f"ComponentManager(_components={self._components}, _managers={self._managers})"
```
collijk: Don't need to include the components/managers in the repr.

src/vivarium/framework/plugins.py
```diff
@@ -125,3 +125,9 @@ def _lookup(self, name):
             return _MANAGERS[name]
         else:
             raise PluginConfigurationError(f'Plugin {name} not found.')
+
+    def __str__(self):
+        return "PluginManager()"
+
+    def __repr__(self):
+        return f"PluginManager(_plugins = {self._plugins})"
```
collijk: Don't include the plugins here

src/vivarium/framework/randomness.py
```diff
@@ -684,6 +685,9 @@ def register_simulants(self, simulants: pd.DataFrame):
             raise RandomnessError("The simulants dataframe does not have all specified key_columns.")
         self._key_mapping.update(simulants.set_index(self._key_columns).index)
 
+    def __str__(self):
+        return "RandomnessManager()"
+
```
collijk: Drop the decision_points from the randomness manager

src/vivarium/framework/state_machine.py
halfhorst: I think we are leaning to little info. The name should be unique and informational. In Jame's words if you are looking at reprs you're holding the object so you can inspect the transitions

I mostly just want to be able to print and not get garbage

collijk: So the way I think about this is that if you have a typical object that has some methods and attributes that themselves may be complex objects, your str/repr should be informative and succinct (and in the case of the repr help you uniquely identify the object).

If you're hanging on to an object that is some kind of data structure or container (as the transition set is) you want to detail the container type as succinctly as possible but also give an indication of what it contains.

Something like f"TransitionSet(self._state_name, [t.output_state.name for t in self.transitions])"
or similar seems like a good str.

src/vivarium/framework/state_machine.py
```diff
@@ -445,5 +452,8 @@ def _get_state_pops(self, index):
         population = self.population_view.get(index)
         return [[state, population[population[self.state_column] == state.state_id]] for state in self.states]
 
+    def __str__(self):
+        return "Machine()"
+
     def __repr__(self):
-        return "Machine(states= {}, state_column= {})".format(self.states, self.state_column)
+        return f"Machine(states= {self.states}, state_column= {self.state_column})"
```
collijk: Here though, a Machine is more a complex object than a container for states.

src/vivarium/framework/state_machine.py
```diff
@@ -353,11 +356,8 @@ def __iter__(self):
     def __len__(self):
         return len(self.transitions)
 
-    def __str__(self):
-        return str([str(x) for x in self.transitions])
-
     def __repr__(self):
-        return repr([repr(x) for x in self.transitions])
+        return f"TransitionSet(transitions={[x for x in self.transitions]})"
```
kfwilson: I think I'm still a little confused about why sometimes we have just repr and sometimes both repr and str. I haven't been able to figure out the pattern - e.g., here we only have a repr with information about the specific transition set but machine below has both a repr with info about the specific machine and str with just Machine?

halfhorst: I'll remove the str from Machine

halfhorst: that was an oversight

PR #83, opened 2019-05-17
track names and enforce uniqueness

halfhorst:
Adds a component container class that handles vivarium components and behaves like an ordered set, checking the conditions we care about:
1. has a name
2. name not already present

Several dunders are necessary for the way this is used in setting up components. I think this is a satisfactory minimal set we could build from.

I had to refactor the tests just a bit because we were testing behavior of duplicates. I added tests for the component list class, and smudged together exercising the dunder capabilities rather than making lots of little methods. Lemme know if you'd rather have separate funcs

### DIFF
--- a/src/vivarium/framework/components/manager.py
+++ b/src/vivarium/framework/components/manager.py
@@ -26,6 +26,39 @@ class ComponentConfigError(VivariumError):
     pass
 
 
+class ComponentList(list):
+
+    def _check_conditions(self, component):
+        if not hasattr(component, "name"):
+            raise ComponentConfigError(f"Attempting to add a component {component} with no name. Vivarium "
+                                       "requires each component in a simulation have a name attribute with"
+                                       "a unique name.")
+        if component in self:
+            raise ComponentConfigError(f"Attempting to append component with duplicate name: {component}")
+
+    def append(self, component) -> None:
+        self._check_conditions(component)
+        super().append(component)
+
+    def extend(self, components):
+        for c in components:
+            self.append(c)
+
+    def insert(self, i, component):
+        self._check_conditions(component)
+        super().insert(i, component)
+
+    def __setitem__(self, key, value):
+        self._check_conditions(value)
+        super().__setitem__(key, value)
+
+    def __contains__(self, component):
+        for c in self:
+            if component.name == c.name:
+                return True
+        return False
+
+
 class ComponentManager:
     """Maintains references to all components and managers in a ``vivarium``
     simulation, applies their default configuration and initiates their
@@ -41,15 +74,14 @@ class ComponentManager:
     """
 
     def __init__(self):
-        self._managers = []
-        self._components = []
-        self._names = []
+        self._managers = ComponentList()
+        self._components = ComponentList()
 
     @property
     def name(self):
         return "component_manager"
 
-    def add_managers(self, managers: Union[List, Tuple, Any]):
+    def add_managers(self, managers: Union[List, Tuple]):
         """Registers new managers with the component manager. Managers are
         configured and setup before components.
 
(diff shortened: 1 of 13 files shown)


src/vivarium/framework/components/manager.py
```diff
@@ -26,6 +26,40 @@ class ComponentConfigError(VivariumError):
...
+
+    def extend(self, components):
+        for c in components:
+            self.append(c)
+
+    def insert(self, i, component):
+        self._check_conditions(component)
+        super().insert(i, component)
+
+    def __setitem__(self, key, component):
```
kfwilson: I'm not sure about using the same check_conditions method for setitem. Won't this throw the 'attempting to append' error if I try to replace a component with the same name? e.g., if index 1 is a component with name 'NAME' and I try to set index 1 to a new component also with name 'NAME', I'll get an error. Or is that not a use case we think will come up?

src/vivarium/framework/components/manager.py
```diff
@@ -26,6 +26,60 @@ class ComponentConfigError(VivariumError):
...
+        return iter(self.components)
+
+    def __len__(self) -> int:
+        return len(self.components)
+
+    def __bool__(self) -> bool:
+        return bool(self.components)
+
+    def __eq__(self, other) -> bool:
+        try:
```
collijk: I suggest doing:

```
try:
    return type(self) is type(other) and self.names == other.names
except TypeError:
    return False

Comment (kfwilson) [reply src/vivarium/framework/components/manager.py]: but comparing names won't check order and that seems like an important condition for equality here?

File: src/vivarium/framework/components/manager.py [hunk 3]
```diff
@@ -26,6 +26,60 @@ class ComponentConfigError(VivariumError):
...
+            if len(self) != len(other):
+                return False
+            for a, b in zip(self, other):
+                if a.name != b.name:
+                    return False
+            return True
+        except TypeError:
+            return False
+
+    def pop(self) -> Any:
```
kfwilson: I'm wondering if it might make more sense to use a queue than a list for the components since you're always appending to the end and removing from the beginning. Maybe we don't expect to have enough components for the additional time required to remove from the beginning of a list to matter but it seems to fit the use case a little better?

src/vivarium/framework/components/manager.py
halfhorst: Since ordering matters I'm going to convert names to a list and then do this

src/vivarium/framework/components/manager.py
halfhorst: I don't think performance will be much of an issue because these lists are gonna stay relatively small orders of magnitude wise. It seems common to use lists as stacks and queues, they cover the same operations. I think the alternative would be a dequedeque but I'm ambivalent towards that, except that it's an extra import

src/vivarium/framework/components/manager.py
kfwilson: just want to make sure I'm understanding things here: I understood why we had names as a set to make the lookups fast but is the reason we still keep the names separate as a list (even though they're also contained in the components list) to make it so we don't have to do explicit list comprehensions for contains/equality and can just pass it on to the list methods?

src/vivarium/framework/components/manager.py
kfwilson: I think lists are commonly used as stacks but less so for queues - the list documentation says they're inefficient for that. but yeah probably not a huge performance issue so your call.

src/vivarium/framework/components/manager.py
halfhorst: it's just kruft now

PR #84, opened 2019-05-31
Feature/get component by name

halfhorst:
Alters the functions that allow access to the underlying components in the Component Manager and makes those functions available on the interface and from the interactive simulation context.

Also unpins tables

### DIFF
--- a/src/vivarium/framework/components/manager.py
+++ b/src/vivarium/framework/components/manager.py
@@ -15,7 +15,7 @@ setup everything it holds when the context itself is setup.
 
 """
 import inspect
-from typing import Union, List, Tuple, Iterator, Dict, Any
+from typing import Union, List, Tuple, Iterator, Dict, Any, Type
 
 from vivarium.config_tree import ConfigTree
 from vivarium.exceptions import VivariumError
@@ -123,7 +123,7 @@ class ComponentManager:
             else:
                 component_list.add(component)
 
-    def get_components_by_type(self, component_type: Any) -> List[Any]:
+    def get_components_by_type(self, component_type: Type) -> List[Any]:
         """Get all components currently held by the component manager that are an
         instance of ``component_type``.
 
@@ -152,6 +152,7 @@ class ComponentManager:
         for c in self._components:
             if c.name == name:
                 return c
+        raise ValueError(f"No component found with name {name}")
 
     def list_components(self) -> Dict[str, Any]:
         """Get a mapping of component names to components held by the manager.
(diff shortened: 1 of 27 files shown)


src/vivarium/framework/components/manager.py
```diff
@@ -116,27 +116,52 @@ def add_components(self, components: Union[List, Tuple]):
...
         for component in components:
             if isinstance(component, list) or isinstance(component, tuple):
                 self._add_components(component_list, component)
             else:
                 component_list.add(component)
 
-    def get_components(self, component_type: Any) -> List:
-        """Gets a list of components currently held by the component manager
-        that are instances of a specific type. Does not include managers.
+    def get_components_by_type(self, component_type: Any) -> List[Any]:
```
collijk: component type can actually be `type` here

src/vivarium/framework/components/manager.py
```diff
@@ -116,27 +116,52 @@ def add_components(self, components: Union[List, Tuple]):
...
+
+        Parameters
+        ----------
+        name
+            A component name.
+        Returns
+        -------
+            A component that has name ``name`` else None.
+        """
+        for c in self._components:
```
collijk: We don't want to return ``None`` here. we want to error.

src/vivarium/framework/components/manager.py
```diff
@@ -183,21 +208,46 @@ def add_components(self, components: Union[List, Tuple, Any]):
...
         Parameters
         ----------
         component_type
-            A component type to get, compared against internal components
+            A component type to retrieve, compared against internal components
             using isinstance().
         Returns
         -------
-            A list of components.
+            A list of components of type ``component_type``.
```
collijk: Your return type and return docstring are incorrect.

src/vivarium/interface/interactive.py
```diff
@@ -234,9 +234,46 @@ def get_emitter(self, event_type: str) -> Callable:
...
+            A component name.
+        Returns
+        -------
+            A component that has the name ``name`` else None.
+
+        """
+        return self.component_manager.get_component(name)
+
+    @raise_if_not_setup(system_type='component')
+    def get_components_by_type(self, component_type: Any) -> List[Any]:
```
collijk: Don't need this on the interface, I think.

src/vivarium/framework/components/manager.py
```diff
@@ -183,21 +208,46 @@ def add_components(self, components: Union[List, Tuple, Any]):
...
+        name
+            A component name.
+        Returns
+        -------
+            A component that has name ``name`` else None.
+
+        """
+        return self._component_manager.get_component(name)
+
+    def get_components_by_type(self, component_type: Any) -> List[Any]:
```
kfwilson: Should this be ``Union[type, Tuple[type]]`` instead of ``Any``?

halfhorst: yup just saw that in VPH

src/vivarium/framework/components/manager.py
halfhorst: this whole time I didn't realize you meant lowercase type :man_facepalming:

PR #85, opened 2019-06-04
Feature/dep manager

halfhorst:
This is a stepping-stone that introduces a dependency manager and gives it its first task -- managing population initializers. It extracts pop manager tests that were testing dependencies and names them as dependency manager tests as well.

Until we meet to discuss designing some sort of Resource and fleshing out the manager a bit more, it remains pop-manager specific.

This also pins sphinx < 2.1, which introduced some oddness with automodule


src/vivarium/framework/dependency.py
```diff
@@ -0,0 +1,124 @@
+"""
+=================================
+Dependency Management in Vivarium
+=================================
+
+This module provides a tool to manage dependencies on resources within a
+``Vivarium`` simulation. These resources take the form of things that can be
+created and utilized by components, for example columns in the state table or
```
halfhorst: I think the best thing to link here would be the top-level API reference (as opposed to a term in the glossary) but I'm not actually sure how yet

src/vivarium/framework/dependency.py
```diff
@@ -0,0 +1,124 @@
...
+
+class DependencyManager:
+
+    def __init__(self):
+        self.population_initializers = []
+        self.population_initializers_ordered = False
+
+    @property
+    def name(self):
+        return "DependencyManager"
```
kfwilson: ok being picky here - do we want to use camel case or snake case by convention for names? I think most of the names that we have so far are snake case (e.g., 'population_manager')?

src/vivarium/framework/dependency.py
```diff
@@ -0,0 +1,124 @@
...
+        self.population_initializers.append(initializer)
+        self.population_initializers_ordered = False
+
+
+class DependencyInterface:
+
+    def __init__(self, dependency_manager: DependencyManager):
+        self._dependency_manager = dependency_manager
+
+    def register_population_initializer(self, initializer: Tuple[Callable, Sequence[str], Sequence[str]]):
```
kfwilson: can the created/required columns be single strings? If not, I don't think you want to use Sequence

src/vivarium/framework/dependency.py
halfhorst: oh yes it should be snake case

PR #86, opened 2019-06-04
Feature/events docs

kfwilson:


docs/source/concepts/event.rst
```diff
@@ -9,8 +9,133 @@ The Event System
...
+
+
+Registering Listeners
++++++++++++++++++++++
+In order to register a listener to an event to respond when that event is
+emitted, we can use the :func:`register_listener <EventInterface.register_listener()>`.
+The listener itself should be a callable function that takes as its only argument
+the :class:`Event <vivarium.framework.event.Event>` that is emitted.
+
+Suppose we wish to track how many simulants are affected each time step. We
```
kfwilson: this part felt a bit more like a tutorial but it felt odd to talk about registering listeners and not explain how to do it

docs/source/concepts/event.rst
```diff
@@ -9,8 +9,133 @@ The Event System
...
+off the :ref:`Builder <builder_concept>` and provides two options for interacting
+with the event system:
+
+1. :func:`register_listener <EventInterface.register_listener()>` to add a
+listener to a given event to be called on emission
+
+2. :func:`get_emitter <vivarium.framework.event.EventInterface.get_emitter()>` to
+retrieve a callable emitter for a given event
+
+Although methods for both getting emitters and registering listeners are provided,
```
halfhorst: This is  a good candidate for a .. note::  directive to make it stick out more

kfwilson: I put a longer .. caution:: directive below in the section about emitting events explaining that you shouldn't actually do it and why not but if you think it would be helpful to have it highlighted here too, I can definitely add that

halfhorst: The only consideration I have is that people will stop reading when they get the info they want, so attaching the warning close to the information is helpful

PR #93, opened 2019-07-30
Mic951 immutable events

KjellSwedin:
Hi all - two approaches to both implementation and testing. Feedback welcome.


KjellSwedin:
Mic951 immutable events incorporates mic1019. The only change with this is to coerce user_data to {} if it is None.

PR #94, opened 2019-08-01
Mic1019 emitters no events

KjellSwedin:
Event emitters no longer take an Event as an argument:
 - correct arguments in engine.py
 - modify testing code in test_event.py 
Event becomes a namedtuple
 - change type information in Event.py
 - modify signatures in Event.py
 - update documentation in Events.py

### DIFF
--- a/src/vivarium/framework/event.py
+++ b/src/vivarium/framework/event.py
@@ -77,6 +77,12 @@ class Event(NamedTuple):
         """
         return Event(new_index, self.user_data, self.time, self.step_size)
 
+    def __repr__(self):
+        return f"Event(user_data={self.user_data}, time={self.time}, step_size={self.step_size})"
+
+    def __eq__(self, other):
+        return self.__dict__ == other.__dict__
+
 
 class _EventChannel:
     """A named subscription channel that passes events to event listeners."""
--- a/src/vivarium/framework/event.py
+++ b/src/vivarium/framework/event.py
@@ -41,15 +41,15 @@ class Event(NamedTuple):
 
     Attributes
     ----------
-    index :
+    index 
         An index into the population table containing all simulants
         affected by this event.
-    user_data :
+    user_data 
         Any additional data provided by the user about the event.
-    time  :
+    time  
         The simulation time at which this event will resolve. The current
         simulation size plus the current time step size.
-    step_size :
+    step_size 
         The current step size at the time of the event.
     """
     index: pd.Index
@@ -99,10 +99,10 @@ class _EventChannel:
 
         Parameters
         ----------
-        index :
+        index 
             An index into the population table containing all simulants
             affected by this event.
-        user_data :
+        user_data 
             Any additional data provided by the user about the event.
 
         """
--- a/src/vivarium/framework/event.py
+++ b/src/vivarium/framework/event.py
@@ -90,7 +90,7 @@ class _EventChannel:
         self.manager = manager
         self.listeners = [[] for _ in range(10)]
 
-    def emit(self, index: pd.Index, user_data: Dict = None) -> Event:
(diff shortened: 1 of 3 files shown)


src/vivarium/framework/event.py
```diff
@@ -41,15 +41,15 @@ class Event(NamedTuple):
 
     Attributes
     ----------
-    index :
+    index
```
kfwilson: just fyi if you're using type annotations and so not including the types in the docstring, you don't put the colons
see e.g., https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

src/vivarium/framework/event.py
```diff
@@ -111,7 +90,7 @@ def __init__(self, manager):
         self.manager = manager
         self.listeners = [[] for _ in range(10)]
 
-    def emit(self, event: Event) -> Event:
+    def emit(self, index: pd.Index, user_data: Dict = None) -> Event:
```
kfwilson: I think defaulting user_data to None and never setting it to an empty dict if it is None in the code itself (as was previously done in line 58 in the old code) may break things that e.g., check ``if 'sim_state' in event.user_data``

src/vivarium/framework/event.py
```diff
@@ -169,7 +150,7 @@ def setup(self, builder):
         self.clock = builder.time.clock()
         self.step_size = builder.time.step_size()
 
-    def get_emitter(self, name: str) -> Callable:
+    def get_emitter(self, name: str) -> Callable[[pd.Index, Optional[Dict]], Event]:
```
kfwilson: why include Event in the type annotation for the callable arguments to emit? Doesn't emit only take an index and user_data now?

tests/framework/test_event.py
```diff
@@ -4,72 +4,54 @@
...
+        },
+        'new_val': {
+            'index': pd.Index(range(3)),
+            'user_data': {'new_key': 'new_value'},
+            'time': pd.Timestamp.now(),
+            'step_size': 30,
         }
-    )
+    }
 def test_proper_access2(event_init):
```
kfwilson: probably just rename this to ``test_proper_access`` now that you've removed the other test

src/vivarium/framework/event.py
kfwilson: actually I'm not positive if it will break things. It may be that anything that does a check like that will check first if user_data is None - I'm not sure. but since previously we had the user_data always being an empty dict, it might be best just to keep that convention

halfhorst:
using NamedTuple makes this a lot cleaner, it's nice

src/vivarium/framework/event.py
collijk: Definitely go ahead and coerce Nones to empty dicts

src/vivarium/framework/event.py
collijk: @kfwilson The event is the return type of the callable

halfhorst: i think it is return type of the callable

src/vivarium/framework/event.py
halfhorst: having it non-mutable is better practice, though. Maybe we should work out the kinks

src/vivarium/framework/event.py
```diff
@@ -111,7 +90,7 @@ def __init__(self, manager):
         self.manager = manager
         self.listeners = [[] for _ in range(10)]
 
-    def emit(self, event: Event) -> Event:
+    def emit(self, index: pd.Index, user_data: Dict = {}) -> Event:
```
kfwilson: I think you still want this default to be None because you don't want something mutable here but then in the function itself, if user_data comes in as None, reassign it to an empty dict

PR #95, opened 2019-08-03
Feature/randomness docs

kfwilson:
Here's my attempt at conceptualizing randomness in a high-level way. I put in a lot of the pictures that @collijk drew when he was explaining this because I found them very helpful and I tried to write enough that they would be helpful for someone reading this too. 

I think it might also be helpful to have a tutorial explaining everything you can do with RandomnessStreams but that's a separate effort.


docs/source/concepts/crn.rst
```diff
@@ -9,8 +9,212 @@ Common Random Numbers
...
+Randomness within a single simulation
+-------------------------------------
+
+We can think of the randomness system in a single ``vivarium`` simulation as
+represented by a cube. Along the x-axis, we have **time** progressing through
+the simulation. Along the y-axis, we have the **randomness index**, which tracks
+simulants. Along the z-axis, we have **decision points**: single decisions
+within a simulation
+
+.. image:: ../images/crn_cube.jpg
```
halfhorst: I really want to make these in illustrator ^_^

PR #98, opened 2019-08-06
Mic992 pull pdb behavior to vivarium

KjellSwedin:
Copy handle_exceptions() into vivarium utilities:
 - add loguru to requirements
 - necessary imports to utilities.py
 - basic test to test_utilities.py

Implementations still exist in:
 - vivarium_cluster_tools/psimulate/utilities.py
 - vivarium_development/src/vadmin/utilities.py

When we have a new vivarium release we need to:
 - reference the new version
 - remove versions from existing locations
 - use the implementation in the new vivarium release

### DIFF
--- a/src/vivarium/framework/utilities.py
+++ b/src/vivarium/framework/utilities.py
@@ -8,8 +8,11 @@ Collection of utility functions shared by the ``vivarium`` framework.
 """
 from importlib import import_module
 from typing import Callable
+from bdb import BdbQuit
+from loguru import logger
 
 import numpy as np
+import functools
 
 
 def from_yearly(value, time_step):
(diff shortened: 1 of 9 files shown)


src/vivarium/framework/utilities.py
```diff
@@ -8,8 +8,11 @@
 """
 from importlib import import_module
 from typing import Callable
+from bdb import BdbQuit
+from loguru import logger
```
kfwilson: I think we may need to think about logging choices here. Since we're using different logging libraries in different places, using loguru here I think means that the logging message for the uncaught exception won't show up for any of the libraries that configure standard logging (e.g., vivarium). I might be wrong on that but we should definitely test it. If we switch to the standard logging library, I don't think vivarium cluster tools will pick it up unless we explicitly forward the logs (I only forwarded for the worker logs)

KjellSwedin:
I'm going to close this PR and do a new one off the develop branch.
To address the logger issue handle_exceptions will take a logger argument as a stop gap measure

PR #100, opened 2019-08-11
Bugfix/state machine graph

kfwilson:
little bugs I found in trying to use the state machine graphing code for my tx profile state machine

### DIFF
--- a/src/vivarium/framework/state_machine.py
+++ b/src/vivarium/framework/state_machine.py
@@ -442,7 +442,7 @@ class Machine:
             else:
                 dot.node(state.state_id)
             for transition in state.transition_set:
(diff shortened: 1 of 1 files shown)


src/vivarium/framework/state_machine.py
```diff
@@ -442,7 +442,7 @@ def to_dot(self):
             else:
                 dot.node(state.state_id)
             for transition in state.transition_set:
-                dot.edge(state.state_id, transition.output.state_id, transition.label())
+                dot.edge(state.state_id, transition.output_state.state_id, transition.name)
```
kfwilson: @collijk what do we actually want to label these transitions as? Using transition.name seems silly because it's just output state, input state but I wasn't sure what else we'd want to use - maybe the probability function?

PR #102, opened 2019-08-21
correct ConfigTree __str__ behavior. Unrelated - add test for freeze()

KjellSwedin:
First change addresses mic1026

Second change - I saw an infinite recursion error invoking freeze that I was not able to reproduce. It wasn't tested so I added one.

### DIFF
--- a/src/vivarium/config_tree.py
+++ b/src/vivarium/config_tree.py
@@ -192,7 +192,7 @@ class ConfigNode:
                                    for layer, value in self._values.items()]))
 
     def __str__(self):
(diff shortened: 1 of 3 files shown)


src/vivarium/config_tree.py
```diff
@@ -192,7 +192,9 @@ def __repr__(self):
                                    for layer, value in self._values.items()]))
```
kfwilson: the __str__ method looks good! I think we have the same ordering problem in the __repr__ though right? Because we're using the order in the values dict instead of the layers list

PR #103, opened 2019-08-21
Feature/integrate artifact into core

collijk:


docs/source/tutorials/artifact.rst
```diff
@@ -0,0 +1,258 @@
...
+Artifact
+========
+
+A data artifact is a bundle of input data associated with a particular
+model. It is typically stored as an ``hdf`` file on disk with very particular
+formatting. This file is then used by the :mod:`vivarium` simulations to fill
+in all the relevant parameter data.
+
+It is frequently useful to be able to view or modify this data outside the
+simulation.  The :class:`vivarium.Artifact` provides a high level interface to
```
kfwilson: elsewhere you seem to use :class:`~vivarium.Artifact` but here you're not using the tilda - does that matter?

docs/source/tutorials/artifact.rst
```diff
@@ -0,0 +1,258 @@
...
+with the data stored in the hdf file with which we created it.
+
+
+Filter Terms
++++++++++++++
+
+The data stored in artifacts may be large, potentially on the order of millions
+of rows for a single dataset, and loading a full dataset requires time and
+memory, both of which may be limited. If you are only interested in certain
+subsets of the data, either rows or columns, you may want to read only the
```
kfwilson: I don't think you can filter to select columns besides draw columns via the artifact - you can via the raw hdf but the Artifact only exposes a single set of filter terms, splits off the term relevant to draws and then passes that and the remaining terms onto the underlying hdf separately. I don't know if it's worth making that distinction here/if it really matters, but the way this sentence is written, it does make it sound like you could select any set of columns

docs/source/tutorials/artifact.rst
```diff
@@ -0,0 +1,258 @@
...
+
+.. code-block:: python
+
+    art.load('a.fake.key')
+
+::
+
+    Traceback (most recent call last):
+      File "<stdin>", line 1, in <module>
+      File "/home/kate/code/vivarium/vivarium_public_health/src/vivarium_public_health/dataset_manager/artifact.py", line 75, in load
```
kfwilson: This exception example still references the path in vph so we should probably update

docs/source/tutorials/artifact.rst
```diff
@@ -0,0 +1,258 @@
...
+
+.. code-block:: python
+
+    art.write('locations.names', ['New York', 'Florida'])
+
+::
+
+    Traceback (most recent call last):
+      File "<stdin>", line 1, in <module>
+      File "/home/kate/code/vivarium/vivarium_public_health/src/vivarium_public_health/dataset_manager/artifact.py", line 105, in write
```
kfwilson: same not as above re exception path is in vph

src/vivarium/framework/artifact/artifact.py
```diff
@@ -0,0 +1,281 @@
...
+
+        Raises
+        ------
+        ArtifactException :
+            If the provided key already exists in the artifact.
+
+        """
+        if entity_key in self.keys:
+            raise ArtifactException(f'{entity_key} already in artifact.')
+        elif data is None:
```
kfwilson: remind me again why this is here? what use case do we have for calling write with None?

docs/source/tutorials/artifact.rst
collijk: The tilda removes leading import paths.  I left the full path in since it's the first mention.

PR #107, opened 2019-09-02
Feature/cleanup entry points

collijk:
The main point of this PR is to cleanup the entry points of the simulation to make the interfaces cleaner.  This is done by changing the `__init__` method of the simulation context to 
```
def __init__(model_specification, components, configuration, plugin_configuration):
```
And then handling all use cases internally.  This gets rid of the many functions used to create simulations in the interactive interface.  

Also, I axed the ResultsWriter.  We're overhauling this in the next major feature release and we already removed all usages in cluster tools in a previous release. Most of the code was dead code at this point.

I still need to put in a bunch of docs and tests, but I'm going to get to those in the last feature prs once all the code is stable.

More detailed comments in the code.


src/vivarium/__init__.py
```diff
@@ -2,9 +2,8 @@
 numpy.seterr(all='raise')
 
 from vivarium.config_tree import ConfigTree
-from vivarium.interface import (build_simulation_configuration, setup_simulation,
```
collijk: Good bye!

src/vivarium/framework/configuration.py
```diff
@@ -9,24 +9,31 @@
...
+from typing import Union, Dict
 
 import yaml
 
 from vivarium.config_tree import ConfigTree, ConfigurationError
+from vivarium.framework.plugins import DEFAULT_PLUGINS
 
-from .plugins import DEFAULT_PLUGINS
 
+def build_model_specification(model_specification: Union[str, Path, ConfigTree] = None,
```
collijk: **START HERE**

This is the magic.  Well, some work on the configtree makes the magic work.  But what this does is creates the full model specification from the wide ranges of arguments we might receive on initialization.

Recall that the simulation configuration has four levels (in increasing priority level):
- `base` - lowest level defaults are set here
- `component_configs` - where component configuration defaults are set
- `model_override` - where information from a model specification is set
- `override` - Highest priority, for command line args and interactive setting

So this function does:

1. `_get_default_specification` creates the config tree, fills in the default plugins and any configuration in the user `vivarium.yaml` file into the `base` layer of the configuration.
2. We then set any information in the model specification into the `model_override` layer
3. Additional arguments from the component_configuration, configuration, and plugin_configuration are put in as `override`, allowing someone to easily override a file on disk by handing this function a dictionary of info.

src/vivarium/framework/configuration.py
```diff
@@ -63,6 +68,6 @@ def _get_default_specification():
 
     user_config_path = Path('~/vivarium.yaml').expanduser()
     if user_config_path.exists():
-        model_specification.configuration.update(user_config_path, layer='component_configs')
+        model_specification.configuration.update(user_config_path, layer='base')
```
collijk: We may need to think about this more closely later, but setting at the component_configs level didn't make sense.  From a users perspective, every simulation they run has these values set.  We perhaps need a fifth `user_defaults` level.

src/vivarium/framework/engine.py
```diff
@@ -35,107 +33,139 @@
...
-    def __init__(self, configuration, components, plugin_manager=None):
-        plugin_manager = plugin_manager if plugin_manager else PluginManager()
 
-        self.configuration = configuration
+    def __init__(self, model_specification: Union[str, Path, ConfigTree] = None,
+                 components: Union[List, Dict[str, str], ConfigTree] = None,
+                 configuration: Union[Dict, ConfigTree] = None,
+                 plugin_configuration: Union[Dict[str, str], ConfigTree] = None):
+        # Bootstrap phase: Parse arguments, make private managers
+        component_configuration = components if isinstance(components, (dict, ConfigTree)) else None
```
collijk: **Look here 2nd**

The only extra bit we have to do on the parsing is to potentially handle an incoming list of initialized components.

src/vivarium/framework/engine.py
```diff
@@ -35,107 +33,139 @@
...
+        # life-cycle manager).
+        self._lifecycle = self._plugin_manager.get_plugin('lifecycle')
+        self._lifecycle.add_phase('setup', ['setup', 'post_setup', 'population_creation'])
+        self._lifecycle.add_phase(
+            'main_loop', ['time_step__prepare', 'time_step', 'time_step__cleanup', 'collect_metrics'], loop=True
+        )
+        self._lifecycle.add_phase('simulation_end', ['simulation_end', 'report'])
 
-        self.builder = Builder(configuration, plugin_manager)
+        self._component_manager = self._plugin_manager.get_plugin('component_manager')
```
collijk: made all the manager private.  Otherwise this code is essentially the same as before.

src/vivarium/framework/engine.py
```diff
@@ -35,107 +33,139 @@
...
-            [self.clock, self.lifecycle, self.population, self.randomness,
-             self.values, self.events, self.tables, self.data]
-        )
-        self.component_manager.add_managers(list(plugin_manager.get_optional_controllers().values()))
-        self.component_manager.add_components(components + [Metrics()])
+        managers = [self._clock, self._lifecycle, self._population, self._randomness, self._values, self._events,
+                    self._tables, self._data] + list(self._plugin_manager.get_optional_controllers().values())
+        self._component_manager.add_managers(managers)
+
+        component_config_parser = self._plugin_manager.get_plugin('component_configuration_parser')
```
collijk: Pushed component parsing way back into the process.  The above setup is all super snappy, so we'll still fail fast *enough*.

src/vivarium/framework/engine.py
```diff
@@ -35,107 +33,139 @@
...
         for event in self.time_step_events:
-            self.lifecycle.set_state(event)
-            self.time_step_emitters[event](self.population.get_population(True).index)
-        self.clock.step_forward()
+            self._lifecycle.set_state(event)
+            self.time_step_emitters[event](self._population.get_population(True).index)
+        self._clock.step_forward()
+
+    def run(self):
+        while self._clock.time < self._clock.stop_time:
```
collijk: Run has been added (it was already in the interactive context), finalize and report have been slightly expanded to incorporate some of the things that were in the pure functions in this module.

src/vivarium/framework/engine.py
```diff
@@ -164,50 +194,13 @@ def __repr__(self):
...
-    plugin_config = model_specification.plugins
-    component_config = model_specification.components
-    simulation_config = model_specification.configuration
-
-    plugin_manager = PluginManager(plugin_config)
-    component_config_parser = plugin_manager.get_plugin('component_configuration_parser')
-    components = component_config_parser.get_components(component_config)
-
-    simulation = SimulationContext(simulation_config, components, plugin_manager)
+def run_simulation(model_specification: Union[str, Path, ConfigTree] = None,
```
collijk: The main entry point now, for use by all CLIs.

src/vivarium/interface/cli.py
```diff
@@ -75,22 +77,22 @@ def run(model_specification, results_directory, verbose, log, with_debugger):
     is provided, a subdirectory will be created with the same name as the
     MODEL_SPECIFICATION if one does not exist. Results will be written to a
     further subdirectory named after the start time of the simulation run."""
+    start = time()
```
collijk: Pushed the results writing behavior out here until we get to our results feature.

src/vivarium/interface/cli.py
```diff
@@ -99,24 +101,21 @@ def test():
...
-        logging.exception("Uncaught exception {}".format(e))
+        _log.exception("Uncaught exception {}".format(e))
         raise
-    finally:
-        shutil.rmtree(results_directory, ignore_errors=True)
 
 
 @simulate.command()
-@click.argument('model_specification', callback=verify_yaml,
+@click.argument('model_specification',
```
collijk: These callbacks are not useful.  They were wrappers around model specification validation, which is already happening when the sim is initialized.

src/vivarium/interface/interactive.py
```diff
@@ -18,18 +18,15 @@
...
 
 import pandas as pd
 
-from vivarium.framework.configuration import build_simulation_configuration, build_model_specification, ConfigTree
 from vivarium.framework.engine import SimulationContext
-from vivarium.framework.plugins import PluginManager
 from vivarium.framework.time import Timedelta, Time
 from vivarium.framework.values import Pipeline
 
-from .utilities import run_from_ipython, log_progress, raise_if_not_setup
```
collijk: Replacing the raise_if_not_setup behavior with lifecycle restrictions in the next pr.

src/vivarium/interface/interactive.py
```diff
@@ -40,35 +37,23 @@ class InteractiveContext(SimulationContext):
...
+        ----------
+        step_size
+            An optional size of step to take. Must be the same type as the
+            simulation clock's step size (usually a pandas.Timedelta).
         """
-        super().initialize_simulants()
-        self._initial_population = self.population.get_population(True)
-
-    @raise_if_not_setup(system_type='run')
-    def reset(self):
```
collijk: Will reimplement this feature when we can make guarantees about it working.

src/vivarium/interface/interactive.py
```diff
@@ -258,117 +214,10 @@ def get_component(self, name: str) -> Any:
             A component that has the name ``name`` else None.
 
         """
-        return self.component_manager.get_component(name)
-
-
-def initialize_simulation(components: List, input_config: Mapping = None,
```
collijk: :wave:

src/vivarium/interface/interactive.py
```diff
@@ -40,35 +37,23 @@ class InteractiveContext(SimulationContext):
...
 
-    def __init__(self, configuration: ConfigTree, components: List, plugin_manager: PluginManager = None):
-        super().__init__(configuration, components, plugin_manager)
-        self._initial_population = None
-
-    def setup(self):
-        """Setup the simulation and initialize its population."""
-        super().setup()
-        self._start_time = self.clock.time
-        self.initialize_simulants()
```
collijk: I'm going to make the interactive context smart shortly so there will never be a need to call setup.

tests/framework/test_lookup.py
```diff
@@ -2,11 +2,19 @@
 import pandas as pd
 
 from vivarium.testing_utilities import build_table, TestPopulation
-from vivarium.interface.interactive import setup_simulation
+from vivarium.framework.engine import SimulationContext
 from vivarium.framework.lookup import validate_parameters
 import pytest
 
 
+def get_context(components, configuration):
```
collijk: To replace with the interactive context once it's smart

tests/framework/test_state_machine.py
```diff
@@ -40,14 +39,22 @@ def inner(self, pop_data):
     return pop_fixture()
 
 
+def get_context(components, configuration=None):
+    simulation = SimulationContext(components=components,
```
collijk: To replace with the interactive context once it's smart.

kfwilson: and you're leaving this same function in multiple places because it's shortly going to be replaced right?

src/vivarium/framework/engine.py
```diff
@@ -35,107 +33,136 @@
...
+        return metrics
 
     def add_components(self, component_list):
-        if self._setup:
-            raise VivariumError('Components cannot be added to a simulation once it is setup.')
-        self.component_manager.add_components(component_list)
+        """Adds new components to the simulation."""
+        # TODO: Set with life cycle restrictions.
+        if self._lifecycle.current_state.name != 'initialization':
+            raise LifeCycleError
```
kfwilson: It seems a little odd not to tell the user in the error message what they did wrong. Is that intentional? Are they supposed to know that they can't add components except in initialization?

src/vivarium/framework/engine.py
```diff
@@ -35,107 +33,136 @@
...
+        self._component_manager.add_components(component_list)
+
+    def get_population(self, untracked: bool = True):
+        # TODO: Set with life cycle restrictions.
+        if self._lifecycle.current_state.name in ['bootstrap', 'initialization', 'setup', 'post_setup']:
+            raise LifeCycleError
```
kfwilson: same comment re above. where are users supposed to go to know when they can call things since we're not telling them here

tests/framework/test_state_machine.py
collijk: Yep

src/vivarium/framework/configuration.py
```diff
@@ -9,24 +9,31 @@
...
+                              component_configuration: Union[Dict, ConfigTree] = None,
+                              configuration: Union[Dict, ConfigTree] = None,
+                              plugin_configuration: Union[Dict, ConfigTree] = None) -> ConfigTree:
+    if isinstance(model_specification, (str, Path)):
+        validate_model_specification_file(model_specification)
 
-def build_model_specification(model_specification_file_path: str) -> ConfigTree:
-    model_specification_file_path = validate_model_specification_file(model_specification_file_path)
+    output_spec = _get_default_specification()
+    output_spec.update(model_specification, layer='model_override', source=str(model_specification))
```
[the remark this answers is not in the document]
collijk: yes

PR #108, opened 2019-09-02
Make the InteractiveContext smarter

collijk:
Only a couple of code changes.  The bulk of this is updating docs/tests to correctly use the new interface.

### DIFF
--- a/src/vivarium/framework/lifecycle.py
+++ b/src/vivarium/framework/lifecycle.py
@@ -184,7 +184,7 @@ class LifeCycleManager:
             new_state.enter()
             self._current_state = new_state
         else:
-            raise LifeCycleError(f'Invalid transition from {self._current_state.name} to {new_state.name} requested.')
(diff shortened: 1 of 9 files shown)


src/vivarium/framework/lifecycle.py
```diff
@@ -165,25 +165,29 @@ class LifeCycleManager:
...
     def set_state(self, state):
         new_state = self.lifecycle.get_state(state)
-        if self.current_state.valid_next_state(new_state):
+        if self._current_state.valid_next_state(new_state):
             new_state.enter()
-            self.current_state = new_state
+            self._current_state = new_state
         else:
-            raise LifeCycleError(f'Invalid transition from {self.current_state.name} to {new_state.name} requested.')
+            raise LifeCycleError(f'Invalid transition from {self._current_state.name} to {new_state.name} requested.')
```
kfwilson: you could just use `self.current_state` here instead of `self._current_state.name`

PR #109, opened 2019-09-02
lifecycle restrictions!

collijk:


src/vivarium/framework/values.py
```diff
@@ -145,19 +147,32 @@ def name(self):
...
+        self.add_constraint = builder.lifecycle.add_constraint
+
+        builder.lifecycle.add_constraint(self.register_value_producer, allow_during=['setup'])
+        builder.lifecycle.add_constraint(self.register_rate_producer, allow_during=['setup'])
+        builder.lifecycle.add_constraint(self.register_value_modifier, allow_during=['setup'])
+        builder.lifecycle.add_constraint(self.get_value, allow_during=['setup', 'post_setup', 'population_creation',
+                                                                       'simulation_end', 'report'])
 
     def on_post_setup(self, event):
         # FIXME: This should raise an error, but can't due to risk effect generation
```
kfwilson: is this fixme still true since we switched to explicit inclusion of risk effects?

PR #113, opened 2019-09-05
Resource Manager implementation

collijk:


src/vivarium/framework/resource.py
```diff
@@ -0,0 +1,123 @@
...
+    def _to_graph(self):
+        g = nx.DiGraph()
+        g.add_nodes_from(set([r for r in self.resources.values() if r.resource_type == 'column']))
+
+        def _add_in_edges_to(node, from_keys):
+            for dependency_key in from_keys:
+                d = self.resources[dependency_key]
+                if d.resource_type == 'column':
+                    try:
+                        g.add_edge(d, node)
```
kfwilson: I think networkx will add the nodes if they're not already in the graph so I'm not sure you'll ever trigger the except here

https://networkx.github.io/documentation/stable/reference/classes/generated/networkx.DiGraph.add_edge.html#networkx.DiGraph.add_edge

src/vivarium/framework/resource.py
```diff
@@ -0,0 +1,123 @@
...
+                    except:  # Some kind of node doesn't exist error
+                        raise ResourceError # Resource r depends on d but d has no producer.
+                else:
+                    _add_in_edges_to(node, from_keys=d.dependencies)
+
+        for r in set(self.resources.values()):
+            _add_in_edges_to(r, r.dependencies)
+
+        return nx.algorithms.topological_sort(g)
```
kfwilson: commenting just to remind myself: this will throw an error if there are cycles in the graph that you'll catch later

src/vivarium/framework/resource.py
```diff
@@ -0,0 +1,123 @@
...
+
+from vivarium.exceptions import VivariumError
+
+
+class ResourceError(VivariumError):
+    """Error raised when a dependency requirement is violated"""
+    pass
+
+
+RESOURCE_TYPES = {'value', 'value_source', 'value_modifier', 'column'}
```
halfhorst: are you planning to check against this set when ResourceProducers are made?

src/vivarium/framework/resource.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class ResourceProducer:
+
+    def __init__(self, resource_type, resource_names, producer, dependencies):
```
halfhorst: some type hints would be nice here. Woulda helped me reason about how this is expected to be used

src/vivarium/framework/resource.py
```diff
@@ -0,0 +1,123 @@
...
+    def __init__(self):
+        self._resource_groups = {}
+
+    @property
+    def name(self):
+        return "resource_manager"
+
+    def add_group(self, phase, single_producer=False):
+        if phase in self._resource_groups:
+            raise  ResourceError  # One resource group per phase
```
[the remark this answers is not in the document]
collijk: yeah, I'll bring them in with the docs.

src/vivarium/framework/lookup.py
```diff
@@ -257,8 +263,11 @@ def build_table(self, data, key_columns, parameter_columns, value_columns) -> Ca
         return table
 
     def _build_table(self, data, key_columns, parameter_columns, value_columns):
-        return LookupTable(data, self._pop_view_builder, key_columns, parameter_columns,
-                           value_columns, self._interpolation_order, self.clock, self._extrapolate)
+        table_number = len(self.tables)
```
collijk: New.  Giving generic names to tables so I can do something reasonable with them when constructing the dependency graph.

src/vivarium/framework/population.py
```diff
@@ -237,8 +238,16 @@ def _get_view(self, columns: Sequence[str], query: str = None):
...
-                                      creates_columns: Sequence[str] = (), requires_columns: Sequence[str] = ()):
-        self.register_population_initializer((initializer, creates_columns, tuple(requires_columns) + ('tracked',)))
+                                      creates_columns: List[str] = (),
+                                      requires_columns: List[str] = (),
+                                      requires_values: List[str] = (),
+                                      requires_streams: List[str] = ()):
+        dependencies = ([f'column.{name}' for name in requires_columns]
+                        + [f'value.{name}' for name in requires_values]
+                        + [f'stream.{name}' for name in requires_streams])
+        if creates_columns != ['tracked']:
```
collijk: New:  The tracked column is a dependency of everything since it's used by the population view directly.

src/vivarium/framework/randomness.py
```diff
@@ -649,6 +650,9 @@ def get_randomness_stream(self, decision_point: str, for_initialization: bool =
             with the same identifier.
         """
         stream = self._get_randomness_stream(decision_point, for_initialization)
+        if not for_initialization:  # We need the key columns to be created before this stream can be called.
```
collijk: New

src/vivarium/framework/resource.py
```diff
@@ -0,0 +1,147 @@
...
+
+    @property
+    def graph(self):
+        if self._graph is None:
+            self._graph = self._to_graph()
+        return self._graph
+
+    def add_resources(self, resource_type: str, resource_names: List[str],
+                      producer: MethodType, dependencies: List[str]):
+        if resource_type not in RESOURCE_TYPES:
```
collijk: Update

src/vivarium/framework/resource.py
```diff
@@ -0,0 +1,147 @@
...
+    def __iter__(self) -> Iterable[MethodType]:
+        try:
+            sorted_nodes = nx.algorithms.topological_sort(self.graph)
+        except nx.NetworkXUnfeasible:
+            raise ResourceError(f'The resource group {self.phase} contains at least one cycle.')
+
+        return iter([r.producer for r in sorted_nodes if r.resource_type == 'column'])
+
+    def _to_graph(self) -> Iterable[ResourceProducer]:
+        g = nx.DiGraph()
```
collijk: Restructured this to build the full dependency graph and have the iterator filter to column type things.  This makes much more compelling and informative dependency plots.

src/vivarium/framework/resource.py
```diff
@@ -0,0 +1,147 @@
...
+                    warnings.warn(f'Resource {dependency_key} is not provided by any component but is needed to '
+                                  f'compute {r.resource_names}.')
+                    continue
+
+                d = self.resources[dependency_key]
+                g.add_edge(d, r)
+        return g
+
+    def __repr__(self):
+        out = {}
```
collijk: New.  Useful in debugging.  Will print 
```
resource group : dependencies
resource group : dependencies
... 
```

src/vivarium/framework/values.py
```diff
@@ -192,15 +212,52 @@ def keys(self):
     def items(self):
         return self._pipelines.items()
 
+    @staticmethod
+    def _convert_dependencies(func, requires_columns, requires_values, requires_streams):
+        # If declaring a pipeline as a value source or modifier, columns and
```
collijk: New

src/vivarium/framework/values.py
```diff
@@ -192,15 +212,52 @@ def keys(self):
...
+            dependencies = [f'value.{func.name}']
+        else:
+            dependencies = ([f'column.{name}' for name in requires_columns]
+                            + [f'value.{name}' for name in requires_values]
+                            + [f'stream.{name}' for name in requires_streams])
+        return dependencies
+
+    @staticmethod
+    def _get_modifier_name(modifier):
+        if hasattr(modifier, 'name'):  # This is Pipeline or lookup table or something similar
```
collijk: New

src/vivarium/framework/population.py
```diff
@@ -237,8 +238,18 @@ def _get_view(self, columns: Sequence[str], query: str = None):
...
+            # The population view itself uses the tracked column, so include
+            # to be safe.
+            dependencies += ['column.tracked']
```
kfwilson: will it matter if tracked is already in the dependencies list?

src/vivarium/framework/resource.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def __contains__(self, item: Any) -> bool:
+        return False
+
+
+class ResourceGroup:
+
+    def __init__(self, phase: str, single_producer: bool = False):
+        self.phase = phase
+        # One initializer per component, maybe multiple value producers
```
kfwilson: maybe update this comment to include streams

src/vivarium/framework/resource.py
```diff
@@ -0,0 +1,147 @@
...
+    def graph(self):
+        if self._graph is None:
+            self._graph = self._to_graph()
+        return self._graph
+
+    def add_resources(self, resource_type: str, resource_names: List[str],
+                      producer: MethodType, dependencies: List[str]):
+        if resource_type not in RESOURCE_TYPES:
+            raise ResourceError(f'Unknown resource type {resource_type}.  Permitted types are {RESOURCE_TYPES}.')
+        if resource_type == 'column':
```
kfwilson: why are you only checking if resource_type == 'column'? Doesn't the EmptySet handle this so you can always do what's in this if block?

src/vivarium/framework/resource.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def __init__(self, phase: str, single_producer: bool = False):
+        self.phase = phase
+        # One initializer per component, maybe multiple value producers
+        self.producer_components = set() if single_producer else EmptySet()
+        self.resources = {}
+        self._graph = None
+
+    @property
+    def graph(self):
```
kfwilson: hmm maybe this is an unlikely to the point of impossible situation, but since there's no limits on when I could call add_resources, couldn't I conceivably add a bunch of resources, call graph, add a bunch more resources, and call graph again and get back a graph without all the new resources I added?

src/vivarium/framework/population.py
collijk: No.  We'll just end up adding the tracked dependency edge twice, which networkx ignores.

src/vivarium/framework/resource.py
collijk: This checks by resource type.  This is maybe a weird place to enforce this constraint, but the idea is that a single component can only provide one method that produces columns (population initializers) but may supply many methods that produce value modifiers/sources.

The same would be true if we were doing, e.g., time step handlers.  Each component should have one method that updates columns, but still has all the methods that source pipelines.

src/vivarium/framework/resource.py
collijk: yes.  Don't do that.

PR #117, opened 2019-09-09
Feature/loguru

kfwilson:
example terminal logs from running a single sim: 
image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,6 +23,7 @@ if __name__ == "__main__":
         'scipy',
         'click',
         'tables',
+        'loguru',
     ]
 
     interactive_requirements = [
--- a/src/vivarium/interface/utilities.py
+++ b/src/vivarium/interface/utilities.py
@@ -10,6 +10,9 @@ interfaces for ``vivarium``.
 from datetime import datetime
 import functools
 from pathlib import Path
+import sys
+
+from loguru import logger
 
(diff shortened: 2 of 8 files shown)


setup.py
```diff
@@ -23,6 +23,7 @@
         'scipy',
         'click',
         'tables',
+        'loguru',
```
kfwilson: @collijk should I also put loguru in the requirements.txt? Is that even used?

collijk: It is not.  We can get rid of it.

src/vivarium/interface/utilities.py
```diff
@@ -115,3 +118,23 @@ def get_output_root(results_directory, model_specification_file):
...
+                      '<cyan>{name}</cyan>:<cyan>{function}</cyan>:<cyan>{line}</cyan> '
+                      '- <level>{message}</level>')
+    if verbose:
+        logger.add(sink, colorize=colorize, level="DEBUG", format=message_format, serialize=serialize)
+    else:
+        logger.add(sink, colorize=colorize, level="ERROR", format=message_format, serialize=serialize)
+
+
+def configure_logging_to_terminal(verbose):
+    logger.remove()  # Clear default configuration
```
collijk: Use logger.remove(0).  I think we just want to remove the default logger.  Otherwise I think this mucks up log handlers in the distributed tools.

PR #119, opened 2019-09-18
add values docs

kfwilson:
ok here's a concept note + updates to the values docstrings. I started writing an example at the end with multiple components composing a value but a) it doesn't seem like concept notes are the place for what is essentially a tutorial and b) it was really hard to write something that didn't get into the messiness of other systems (e.g., initializing sims, randomness, etc.). So it seemed better to just point people back to the tutorials where they can see examples but let me know if you disagree and I can try to put together something simple (but not trivial)


docs/source/concepts/values.rst
```diff
@@ -9,8 +9,106 @@ The Values System
...
- - The difference between
- - Specifying defaults.
+Why the values system
+---------------------
+The values system provides an interface to an alternative representation of
+state in the simulation: pipelines. Pipelines are dynamically calculated values
+that can be constructed across multiple components. This ability for multiple
+components to together compose a single value is the biggest advantage pipelines
+provide over the standard state representation of the population state table.
+You should use the values system when you have a value that must be composed
```
collijk: Make this part big and bold in some way.

PR #121, opened 2019-09-20
Docs/population system

halfhorst:
Mostly docstrings and a concept note. The pop file is quite old so I removed some super dated things. I also added more info to an exception message. I didn't hand draw concept note figures, just annotated where I think there is potential for a figure. I'd like to use illustrator instead of drawing. 

A lot of the linking and cross-ref remains to be done after the text is settled. I think there are a few glossary terms in here, too.


src/vivarium/framework/population.py
```diff
@@ -247,8 +243,8 @@ def _order_initializers(self) -> None:
 
         self._validate_no_missing_initializers(unordered_initializers)
 
-        # This is the brute force N! way because constructing a dependency graph is work
-        # and in practice this should run in about order N time due to the way dependencies are
+        # This is the brute force N^2 way because constructing a dependency graph is work.
```
halfhorst: I think I was wrong the first time, it's n**2. but it doesn't matter anymore :grin:

docs/source/concepts/population.rst
```diff
@@ -9,8 +9,77 @@ Population Management
    :local:
    :backlinks: none
 
-Outline
--------
- - Initialization of simulants and the dependency problem.
- - Managing access with population views.
- -
+Since ``Vivarium`` is an agent-based simulation framework, managing a group of
```
collijk: Internally, refer to vivarium as :mod:`vivarium`.  Get's us nice cross-link functionality.  

Also, vivarium itself is not agent-based (!).  There are two ways this is false:
1. vivarium itself doesn't enforce anything about the rows in the state table representing individuals.  The whole MSLT model works by treating each group as an age/sex cohort and has columns representing population counts/proportions in those groups.  This doesn't matter at all for our researchers (possibly, unless they want to do other kinds of modeling), but matters significantly for how we build the managers/engine/other abstractions in the framework itself.
2. for experts in the field there's a big distinction between agent-based modeling which focuses on individual interactions (think infectious disease dynamics) and microsimulation which models individuals but mainly cares about the impact of population heterogeneity.

docs/source/concepts/population.rst
```diff
@@ -9,8 +9,77 @@ Population Management
...
+:term:`simulants` and their attributes is a critical task. Fundamentally, to run
+a simulation we need to be able to create new simulates, update their state
+attributes, and facilitate access to their state so that :term:`components` in
+the simulation can do interesting things based on it. The tooling to support
+working with our simulant population is called the population management system.
```
collijk: I've been putting the intro section above the table of contents because I like the rendered formatting better.  Thoughts?

docs/source/concepts/population.rst
```diff
@@ -9,8 +9,77 @@ Population Management
...
+
+The State Table
+---------------
+
```
collijk: I think the best way to think about this document is to take it a section at a time.  I think talking about the simulation state is the right first thing to do.  We want to try to separate the talk about the intention/theory, and then illustrate the implementation with code examples.  A bunch of the images I made for the presentations might be useful here.  They're all at https://github.com/collijk/vivarium-demo/tree/master/2019_07_25_ihme/images

I would talk about the simulation state a bit abstractly at first.  E.g. it's the set of all information about the whole population *at a particular point in time*.  Here's a picture of a group of people.  Think about what attributes they might have (blood pressure, salary, ate a hamburger today).  

Then move on to the representation, organizing all the state info into rows and columns.

You can then use code examples from the examples sub-package to show the actual state table (doctest blocks will both display code and the outputs if I remember correctly).

docs/source/concepts/population.rst
```diff
@@ -9,8 +9,77 @@ Population Management
...
+representation rows represent simulants while columns correspond to state
+attributes like age, sex or systolic blood pressure. These columns represent one
+of several important resources within ``Vivarium`` that other components can
+draw on. Each of the actions we need to be able to take correspond to a
+manipulation of this state table. The addition of new simulants is the creation
+of rows, the creation of new state attributes is the creation of columns, and
+the reading and updating of state is reading and updating the dataframe itself.
+
+<<TODO: image of state table w/ expanding rows and columns>>
+
```
collijk: Talk about simulant creation as the next section.  Mention how this is distinct from the initialization of population attributes so you can put off talking about population views for a second.  Play up the distinction between the initial population creation and what happens during the main loop more.  Remember, audience is engineers, so what's happening to the state table in these two cases is important.  initial creation builds an index out of nothing and adds columns piecemeal.  main loop creation builds out a totally separate table before appending.  Pictures (hand-drawn is fine, or get fancy if you like) are super useful.  Then code examples of getting the simulant creator and how to use it (and when you can).  In particular, that initial population creation is done by the framework itself (with a warning that this behavior is expected to change).

Do this as a separate PR from the state table bit

docs/source/concepts/population.rst
```diff
@@ -9,8 +9,77 @@ Population Management
...
+Population Views
+----------------
```
collijk: Instead of a section about views, have a section about accessing and updating the population.  Start out with why it's important framed in the context of time steps.  A picture of two state tables would be good.  E.g. I need to read these columns of X_t to update these columns of X_t+1. Then talk about the implementation (and how it relates to the idea that components map to sets of attributes/columns).  Code examples of how to get one.  How to get subviews.  How to query.  I think relating them to accessing SQL tables might be helpful (SELECT and WHERE map reasonably well to subview and get).  Talk about the Markov property (we shouldn't be using the state table to communicate between components during the same time step and warn that this will be enforced).  Something brief and reference to a new concept stub.

Again separate PR.

src/vivarium/framework/population.py
```diff
@@ -4,10 +4,10 @@
 ================================
```
collijk: Open a docstring only PR for everyone to review now. This is much easier than the concept stuff.

PR #123, opened 2019-09-24
switch to not requiring specifying start/end column names for param c…

kfwilson:
don't require specification of start/end column names since they're standard now

### DIFF
--- a/src/vivarium/framework/lookup.py
+++ b/src/vivarium/framework/lookup.py
@@ -20,7 +20,7 @@ from typing import Union, List, Tuple, Callable, TypeVar
 
 import pandas as pd
 
-from vivarium.interpolation import Interpolation, ParameterType
+from vivarium.interpolation import Interpolation
 from vivarium.framework.population import PopulationView
 
 ScalarValue = TypeVar('ScalarValue', Number, timedelta, datetime)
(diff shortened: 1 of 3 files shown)


src/vivarium/interpolation.py
```diff
@@ -28,9 +28,7 @@ class Interpolation:
         continuous_parameters :
             Column names to be used as continuous parameters in Interpolation. If
             bin edges, should be of the form (column name used in call, column name
-            for left bin edge, column name for right bin edge). If order is 0 and
-            any continuous parameter is given without bin edges, continuous bins
```
kfwilson: In going through, realized this was still hanging around but is not true - if any continuous parameter is given without bin edges, interpolation will break and yell at you

collijk:
Since we're here, we might as well remove the defaults and make things explicit downstream.

src/vivarium/framework/lookup.py
```diff
@@ -281,9 +285,7 @@ class LookupTableInterface:
     def __init__(self, manager):
         self._lookup_table_manager = manager
 
-    def build_table(self, data, key_columns=('sex',), parameter_columns=(['age', 'age_start', 'age_end'],
```
collijk: :-)

tests/framework/test_lookup.py
```diff
@@ -92,10 +92,7 @@ def test_interpolated_tables__exact_values_at_input_points(base_config):
 
     simulation = InteractiveContext(components=[TestPopulation()], configuration=base_config)
     manager = simulation._tables
-    years = manager._build_table(years, key_columns=('sex',),
```
collijk: Let's be consistent about providing lists of args.

PR #124, opened 2019-09-26
handle resource producers that don't produce anything

kfwilson:
This adds the concept of a null producer (essentially a pure resource consumer instead of a consumer/producer) so that initializers, particularly in observer components, that don't create any state table columns but do have dependencies on other resources still get called

### DIFF
--- a/src/vivarium/framework/resource.py
+++ b/src/vivarium/framework/resource.py
@@ -31,6 +31,7 @@ class ResourceError(VivariumError):
 
 
 RESOURCE_TYPES = {'value', 'value_source', 'value_modifier', 'column', 'stream'}
+NULL_RESOURCE_TYPE = 'null'
 
 
 class ResourceProducer:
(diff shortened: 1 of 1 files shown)


src/vivarium/framework/resource.py
```diff
@@ -83,21 +87,28 @@ def add_resources(self, resource_type: str, resource_names: List[str],
...
-        for resource_name in resource_names:
-            key = f'{resource_type}.{resource_name}'
+        producer = self.get_producer(resource_type, resource_names, producer, dependencies)
+        for resource_name in producer.resource_names:
+            key = f'{producer.resource_type}.{resource_name}'
             if key in self.resources:
                 raise ResourceError  # More than one producer for resource ...
             self.resources[key] = producer
 
+    def get_producer(self, resource_type, resource_names, producer, dependencies):
```
collijk: This should probably be private.  Like lookup tables/pop views/pipelines, this object get's handled directly by other components.

src/vivarium/framework/resource.py
```diff
@@ -83,21 +87,28 @@ def add_resources(self, resource_type: str, resource_names: List[str],
...
+        if not resource_names:  # null producer
+            resource_type = NULL_RESOURCE_TYPE
+            resource_names = [self._null_producer_count]
```
collijk: Coerce the count to a string for type consistency

PR #127, opened 2019-10-03
Feature/api docs

collijk:
- joint value post processor renamed to union post processor.  Reasoning in the doc string.
- Added a 'missing_value_source' type and put pipelines with missing sources explicitly in the dependency graph.  This will be used later for visualization.
- Renamed the interface manager to just '_manager', which i've been doing piecemeal for a while.  I should just do one pr with all of them.  Todo soon.
- Axe the set combiner since it was never used

Everything else is docs

### DIFF
--- a/src/vivarium/framework/resource.py
+++ b/src/vivarium/framework/resource.py
@@ -30,7 +30,7 @@ class ResourceError(VivariumError):
     pass
 
 
-RESOURCE_TYPES = {'value', 'value_source', 'value_modifier', 'column', 'stream'}
+RESOURCE_TYPES = {'value', 'value_source', 'missing_value_source', 'value_modifier', 'column', 'stream'}
 
 
 class ResourceProducer:
--- a/src/vivarium/framework/values.py
+++ b/src/vivarium/framework/values.py
@@ -3,8 +3,8 @@
 The Value Pipeline System
 =========================
 
-The value pipeline system is a vital part of the ``vivarium`` infrastructure.
-It allows for values that determine the behavior of individual
+The value pipeline system is a vital part of the :mod:`vivarium`
+infrastructure. It allows for values that determine the behavior of individual
 :term:`simulants <Simulant>` to be constructed across across multiple
 :ref:`components <components_concept>`.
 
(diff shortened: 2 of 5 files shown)


src/vivarium/framework/lookup.py
```diff
@@ -20,7 +20,7 @@
 
 import pandas as pd
```
collijk: Changes in this file have already been reviewed and approved.

src/vivarium/framework/population.py
```diff
@@ -93,10 +93,11 @@ def get(self, index: pd.Index, query: str = '') -> pd.DataFrame:
         """
         pop = self.manager.get_population(True).loc[index]
 
-        if self._query:
```
collijk: Changes in this file have already been reviewed and approved.

src/vivarium/framework/resource.py
```diff
@@ -30,7 +30,7 @@ class ResourceError(VivariumError):
     pass
 
 
-RESOURCE_TYPES = {'value', 'value_source', 'value_modifier', 'column', 'stream'}
+RESOURCE_TYPES = {'value', 'value_source', 'missing_value_source', 'value_modifier', 'column', 'stream'}
```
collijk: To use later in graph display. We allow missing value sources for pipelines that never get called.

tests/framework/test_engine.py
```diff
@@ -48,7 +48,7 @@ def test_SimulationContext_init_default(components):
     assert isinstance(sim._builder.lookup, LookupTableInterface)
     assert sim._builder.lookup._lookup_table_manager is sim._tables
     assert isinstance(sim._builder.value, ValuesInterface)
-    assert sim._builder.value._value_manager is sim._values
+    assert sim._builder.value._manager is sim._values
```
collijk: Slowly making all the manager names just "manager"

src/vivarium/framework/resource.py
halfhorst: as in the source of the pipeline is "missing value" ? Is that what that means?

src/vivarium/framework/values.py
```diff
@@ -171,46 +301,86 @@ def register_value_producer(self, value_name, source,
...
+        ----------
+        value_name :
+            The name of the dynamic value pipeline to be modified.
+        modifier :
+            A function that modifies the source of the dynamic value pipeline
+            when called. If the pipeline has a ``replace_combiner``, the
+            modifier should accept the same arguments as the pipeline source
+            with an additional last positional argument for the results of the
+            previous stage in the pipeline. For the ``list_combiner`` and
+            ``set_combiner`` strategies, the pipeline modifiers should have
```
halfhorst: set was removed

src/vivarium/framework/resource.py
collijk: This will be a pipeline with no source. Which we allow, I just want to visually represent.

PR #128, opened 2019-10-30
Feature/lifecycle tests

collijk:


src/vivarium/framework/engine.py
```diff
@@ -118,7 +118,7 @@ def setup(self):
 
         self.simulant_creator = self._builder.population.get_simulant_creator()
 
-        self.time_step_events = self._lifecycle.get_states('main_loop')
+        self.time_step_events = self._lifecycle.get_state_names('main_loop')
```
collijk: Changed for type consistency

src/vivarium/framework/lifecycle.py
```diff
@@ -56,13 +56,18 @@ def __init__(self, name: str):
         self._next = None
         self._loop_next = None
         self._entrance_count = 0
-        self.handlers = []
+        self._handlers = []
```
collijk: No reason to be public

PR #138, opened 2019-11-15
Clean up and de-engineer some of the resource manager

collijk:
I was going back to write tests and I realized I severely overengineered the resource manager, which made things way more confusing.  Notes in the code.  Included are docstring and comments in the resource manager code that should make this review less confusing than the initial version.

### DIFF
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -86,8 +86,6 @@ class SimulationContext:
         for name, controller in self._plugin_manager.get_optional_controllers().items():
             setattr(self, f'_{name}', controller)
 
-        self._resource.add_group('initialization')
-
         # The order the managers are added is important.  It represents the
         # order in which they will be set up.  The clock is required by
         # several of the other managers, including the lifecycle manager.  The
@@ -109,6 +107,10 @@ class SimulationContext:
         self._lifecycle.add_constraint(self.get_population, restrict_during=['initialization', 'setup', 'post_setup'])
 
         self.add_components(components)
+
+    @property
+    def name(self):
+        return 'simulation_context'
 
     def setup(self):
         self._lifecycle.set_state('setup')
@@ -182,7 +184,7 @@ class Builder:
         self.value = plugin_manager.get_plugin_interface('value')                   # type: ValuesInterface
         self.event = plugin_manager.get_plugin_interface('event')                   # type: EventInterface
         self.population = plugin_manager.get_plugin_interface('population')         # type: PopulationInterface
-        self.resource = plugin_manager.get_plugin_interface('resource')             # type: ResourceInterface
+        self.resources = plugin_manager.get_plugin_interface('resource')             # type: ResourceInterface
         self.randomness = plugin_manager.get_plugin_interface('randomness')         # type: RandomnessInterface
         self.time = plugin_manager.get_plugin_interface('clock')                    # type: TimeInterface
         self.components = plugin_manager.get_plugin_interface('component_manager')  # type: ComponentInterface
--- a/src/vivarium/framework/event.py
+++ b/src/vivarium/framework/event.py
@@ -35,7 +35,6 @@ from .lifecycle import ConstraintError
 from .time import Time, Timedelta
 
 
-
 class Event(NamedTuple):
     """An Event object represents the context of an event.
 
@@ -90,7 +89,8 @@ class Event(NamedTuple):
 
 class _EventChannel:
     """A named subscription channel that passes events to event listeners."""
-    def __init__(self, manager):
+    def __init__(self, manager, name):
+        self.name = f'event_channel_{name}'
         self.manager = manager
         self.listeners = [[] for _ in range(10)]
 
@@ -137,12 +137,17 @@ class EventManager:
     """
 
     def __init__(self):
-        self._event_types = defaultdict(lambda: _EventChannel(self))
+        self._event_types = {}
 
     @property
     def name(self):
         """The name of this component."""
         return "event_manager"
+
+    def get_channel(self, name):
+        if name not in self._event_types:
+            self._event_types[name] = _EventChannel(self, name)
+        return self._event_types[name]
 
     def setup(self, builder):
         """Performs this component's simulation setup.
@@ -182,7 +187,7 @@ class EventManager:
             listeners
 
         """
-        channel = self._event_types[name]
+        channel = self.get_channel(name)
         try:
             self.add_constraint(channel.emit, allow_during=[name])
         except ConstraintError:
@@ -204,7 +209,7 @@ class EventManager:
             Number in range(10) used to assign the ordering in which listeners
             process the event.
         """
-        self._event_types[name].listeners[priority].append(listener)
+        self.get_channel(name).listeners[priority].append(listener)
 
     def get_listeners(self, name: str) -> Dict[int, List[Callable]]:
         """Get  all listeners registered for the named event.
@@ -219,7 +224,7 @@ class EventManager:
             A dictionary that maps each priority level of the named event's
             listeners to a list of listeners at that level.
         """
-        channel = self._event_types[name]
+        channel = self.get_channel(name)
         return {priority: listeners for priority, listeners in enumerate(channel.listeners) if listeners}
 
     def list_events(self) -> List[Event]:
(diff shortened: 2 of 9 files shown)


src/vivarium/framework/engine.py
```diff
@@ -86,8 +86,6 @@ def __init__(self, model_specification: Union[str, Path, ConfigTree] = None,
         for name, controller in self._plugin_manager.get_optional_controllers().items():
             setattr(self, f'_{name}', controller)
 
-        self._resource.add_group('initialization')
```
collijk: I fused the resource group and resource manager since the only resource group we were creating was this one.  So this goes away.

src/vivarium/framework/engine.py
```diff
@@ -110,6 +108,10 @@ def __init__(self, model_specification: Union[str, Path, ConfigTree] = None,
 
         self.add_components(components)
 
+    @property
+    def name(self):
```
collijk: This, and all other naming fixes, is to make the constraint maker in the lifecycle system happy.  I was still getting random failures in doctest related to identifying constrained methods due to python garbage collecting population views and lookup tables.  Methods that need to be constrained now must be attached to an object with a unique name.  This is already a deal for our components.  It's now part of, e.g., the context and things like lookup tables, randomness streams, and population views.

src/vivarium/framework/event.py
```diff
@@ -90,7 +89,8 @@ def __eq__(self, other):
 
 class _EventChannel:
     """A named subscription channel that passes events to event listeners."""
-    def __init__(self, manager):
+    def __init__(self, manager, name):
+        self.name = f'event_channel_{name}'
```
collijk: See naming note

src/vivarium/framework/event.py
```diff
@@ -137,13 +137,18 @@ class EventManager:
...
     def __init__(self):
-        self._event_types = defaultdict(lambda: _EventChannel(self))
+        self._event_types = {}
 
     @property
     def name(self):
         """The name of this component."""
         return "event_manager"
 
+    def get_channel(self, name):
```
collijk: Jujitsu to add names to event channels

src/vivarium/framework/lifecycle.py
```diff
@@ -376,7 +376,7 @@ def to_guid(method: MethodType) -> str:
         has been constrained before.
 
         """
-        return f'{str(id(method.__self__))}.{method.__name__}'
+        return f'{method.__self__.name}.{method.__name__}'
```
collijk: This is the constraint maker fix.  If something like a population view gets garbage collected, it's id gets repurposed by python.

src/vivarium/framework/population.py
```diff
@@ -114,7 +120,7 @@ def subview(self, columns: Union[List[str], Tuple[str]]) -> 'PopulationView':
         if set(columns) > set(self.columns):
             raise PopulationError(f"Invalid subview requested.  Requested columns must be a subset of this "
                                   f"view's columns.  Requested columns: {columns}, Available columns: {self.columns}")
-        return PopulationView(self._manager, columns, self.query)
+        return self._manager.get_view(columns, self.query)
```
collijk: This was a bug.  All important objects should have one creation path.

src/vivarium/framework/population.py
```diff
@@ -351,7 +358,7 @@ def setup(self, builder):
         """Registers the population manager with other vivarium systems."""
         self.clock = builder.time.clock()
         self.step_size = builder.time.step_size()
-        self.initialization_resources = builder.resource.get_resource_group('initialization')
+        self.resources = builder.resources
```
collijk: No more resource groups, so components deal directly with the resource manager

src/vivarium/framework/randomness.py
```diff
@@ -428,6 +428,10 @@ def __init__(self, key: str, clock: Callable, seed: Any,
         self._manager = manager
         self._for_initialization = for_initialization
 
+    @property
+    def name(self):
```
collijk: naming for contstraint maker

src/vivarium/framework/randomness.py
```diff
@@ -621,7 +625,7 @@ def setup(self, builder):
         pop_size = builder.configuration.population.population_size
         self._key_mapping.map_size = max(map_size, 10*pop_size)
 
-        self.initialization_resources = builder.resource.get_resource_group('initialization')
+        self.resources = builder.resources
```
collijk: no more resource groups

src/vivarium/framework/resource.py
```diff
@@ -11,147 +11,294 @@
...
+that allows other managers to register resources with the resource manager
 and in turn ask for ordered sequences of these resources according to their
 dependencies or raise exceptions if this is not possible.
 
 """
 from types import MethodType
 from typing import List, Any, Iterable
-import warnings
 
+from loguru import logger
```
collijk: We use the logger for warnings everywhere else, so switched.

src/vivarium/framework/resource.py
```diff
@@ -11,147 +11,294 @@
...
+    specified.
+
+    """
 
     def __init__(self, resource_type: str, resource_names: List[str], producer: MethodType, dependencies: List[str]):
-        self.resource_type = resource_type
-        self.resource_names = resource_names
-        self.producer = producer
-        self.dependencies = dependencies
+        self._resource_type = resource_type
```
collijk: Made all this read only

src/vivarium/framework/resource.py
```diff
@@ -11,147 +11,294 @@
...
-        return f'({resources})'
+        Must be one of :data:`RESOURCE_TYPES`.
 
+        """
+        return self._resource_type
 
-class EmptySet:
+    @property
+    def names(self) -> List[str]:
+        """The long names (including type) of all resources in this group."""
```
collijk: Used this to replace like five instances of the same list comprehension

src/vivarium/framework/resource.py
```diff
@@ -11,147 +11,294 @@
...
-        self._graph = None
-        # null producers are those that don't produce any resources externally but still consume
-        # other resources (i.e., have dependencies) - these are only pop initializers as of 9/26/2019
+    def __str__(self) -> str:
+        resources = ', '.join(self)
+        return f'({resources})'
+
+
+class ResourceManager:
+    """Manages all the resources needed for population initialization."""
```
collijk: You can try to grok the diff, or read this as a reimplementation of the original.  The resource manager had almost no responsibilites besides creating resource groups, so I gutted that and moved all functionality to the manager.

src/vivarium/framework/values.py
```diff
@@ -268,7 +268,9 @@ def on_post_setup(self, _):
         """Finalizes dependency structure for the pipelines."""
         # Unsourced pipelines might occur when generic components register
         # modifiers to values that aren't required in a simulation.
-        logger.warning(f"Unsourced pipelines: {[p for p, v in self._pipelines.items() if not v.source]}")
+        unsourced_pipelines = [p for p, v in self._pipelines.items() if not v.source]
```
collijk: We were getting a warning even if there were no unsourced pipelines.

src/vivarium/framework/values.py
```diff
@@ -284,8 +286,8 @@ def on_post_setup(self, _):
                 dependencies += [f'missing_value_source.{name}']
             for i, m in enumerate(pipe.mutators):
                 mutator_name = self._get_modifier_name(m)
-                dependencies.append(f'value_modifier.{name}.{i}.{mutator_name}')
-            self.initialization_resources.add_resources('value', [name], pipe._call, dependencies)
+                dependencies.append(f'value_modifier.{name}.{i+1}.{mutator_name}')
```
collijk: Bug that was causing a ton of spurious warnings.

src/vivarium/framework/values.py
```diff
@@ -365,14 +367,14 @@ def register_value_modifier(self,
 
         """
         modifier_name = self._get_modifier_name(modifier)
-        logger.debug(f"Registering {modifier_name} as modifier to {value_name}")
 
         pipeline = self._pipelines[value_name]  # May create a pipeline
         pipeline.mutators.append(modifier)
 
         name = f'{value_name}.{len(pipeline.mutators)}.{modifier_name}'
+        logger.debug(f"Registering {name} as modifier to {value_name}")
```
collijk: Log the resource name rather than the modifier name.  Resouce name has info about what pipeline it modifies and what order it gets called in.

PR #139, opened 2019-11-15
Basic tests for resource manager

collijk:
Basic test suite for resource manager.  There are a broad battery of tests we could run here, but I'm sort of depending on networkx to have tested all their stuff.

### DIFF
--- a/tests/framework/test_resource.py
+++ b/tests/framework/test_resource.py
@@ -96,7 +96,7 @@ def test_resource_manager_add_resources():
     for r_type in RESOURCE_TYPES:
         old_names = []
         for i in range(5):
-            c = Component(f'r_type_i')
+            c = Component(f'r_type_{i}')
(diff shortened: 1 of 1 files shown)


tests/framework/test_resource.py
```diff
@@ -0,0 +1,155 @@
...
+    with pytest.raises(ResourceError, match='producers for column.1'):
+        rm.add_resources(r_type, r2_names, r2_producer, r_dependencies)
+
+
+def test_resource_manager_add_resources():
+    rm = ResourceManager()
+    for r_type in RESOURCE_TYPES:
+        old_names = []
+        for i in range(5):
+            c = Component(f'r_type_i')
```
KjellSwedin: c = Component(f'r_type_i')
do you want
c = Component(f'r_type_{i}')

collijk: Good catch

PR #143, opened 2019-11-21
Add config default for location

pandersen13:


collijk:
I'll come back and introduce a different fix over the weekend

PR #144, opened 2019-11-27
Bugfix/population view query

pandersen13:


collijk:
I'll work up the location fix as well, merge it with this PR, and then make a new release over the weekend.  So hang on to this for a minute.

PR #145, opened 2019-11-29
Feature/validation config option

pandersen13:


collijk:
In the future, tag @ihmeuw/vivarium-dev when you submit PRs against vivarium and vivarium_public_health.

PR #146, opened 2020-01-10
Fix bug writing invalid artifact hdf keys

rmudambi:
We were adding a new key to the list of keys before validating the key. Key validation happens automatically within the `hdf.write()` method, so moving that before appending the key fixes the issue. 

Underlying `test_hdf` tests all relevant functionality, so merely testing that the call to `hdf.write()` happens before `self._keys.append()` is sufficient.

### DIFF
--- a/tests/framework/artifact/test_artifact.py
+++ b/tests/framework/artifact/test_artifact.py
@@ -182,8 +182,8 @@ def test_artifact_write(hdf_mock, keys_mock):
     assert key in a.keys
     assert key not in a._cache
     expected_call = [call(path, 'metadata.keyspace', ['metadata.keyspace']),
-                     call(path, 'metadata.keyspace', keys_mock + [key]),
-                     call(path, key, 'data')]
+                     call(path, key, 'data'),
+                     call(path, 'metadata.keyspace', keys_mock + [key])]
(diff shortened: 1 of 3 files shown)


tests/framework/artifact/test_hdf.py
```diff
@@ -272,9 +272,12 @@ def test_EntityKey_init_failure():
     bad_keys = ['hello', 'a.b.c.d', '', '.', '.coconut', 'a.', 'a..c']
 
     for k in bad_keys:
-        with pytest.raises(ValueError):
+        with pytest.raises(ValueError) as err_info:
```
collijk: You can use the `match` keyword argument for `pytest.raises` here instead of making a separate assert.  I'd link you to the docs, but I can't google stuff with my current internet.

tests/framework/artifact/test_artifact.py
```diff
@@ -3,6 +3,7 @@
 from pathlib import Path
 
 from vivarium.framework.artifact.artifact import Artifact, ArtifactException, _to_tree, _parse_draw_filters
+from vivarium.framework.artifact.hdf import EntityKey
```
halfhorst: you brought this in but I don't think it's used anywhere

PR #147, opened 2020-01-30
fix artifact contains method bug

rmudambi:


collijk:
This won't fix our bugs.  First you should set up some regression tests.  Things you might check:

- write data to keys provided as EntityKeys and then load them using the string version of the keys
- write data to keys as strings, then try to load them using EntityKeys
- write data to keys as either strings or EntityKeys, make a new artifact over the file, then try to load the data using EntityKeys

I think there's still a one line fix here, but this isn't the right place.

collijk:
There are three classes at work here:
- the Artifact
- the Keys
- the EntityKey

Look through and make sure you understand where all the type coercion is happening currently.

tests/framework/artifact/test_artifact.py
```diff
@@ -126,6 +127,27 @@ def test_artifact_load(hdf_mock, keys_mock):
...
+def test_artifact_contains(hdf_mock, keys_mock):
+    path = Path('/place/with/artifact.hdf')
+    filter_terms = ['location == Global', 'draw == 10']
+    not_a_key = 'not.a.key'
+
+    art = Artifact(path, filter_terms)
+
+    for key in art.keys:
+        assert key in art
+        assert EntityKey(key) in art
```
rmudambi: I don't think these EntityKey checks are necessary due to the fact that I'm explicitly testing how equality works for EntityKey in test_hdf.py, but I've included them here for now in case you want them. Without them, this test didn't fail before the EntityKey equality changes were made (which should be fine?)

tests/framework/artifact/test_hdf.py
```diff
@@ -306,3 +306,32 @@ def test_EntityKey_with_name():
...
+    key = hdf.EntityKey(string)
+
+    class NonString:
+        def __str__(self):
+            return string
+
+    nonstring = NonString()
+
+    assert key == string, 'Comparision using __eq__ between string object and equivalent EntityKey failed'
+    assert not (key != string), 'Comparision using __ne__ between string object and equivalent EntityKey failed'
```
collijk: Does this fail unless you define __ne__?  I though python would infer it from __eq__.

tests/framework/artifact/test_artifact.py
collijk: The failure cases were:

1. open artifact, write with entity key, load with raw string
2. open artifact, write with raw string, load with entity key
3. open artifact, write with raw string or entity key, close artifact, open artifact, load with entity key

collijk: These are good regression tests, if nothing else, and they're cheap to run.

tests/framework/artifact/test_hdf.py
rmudambi: python does **not** infer it from eq. Something I learned today.

collijk: You an me both.  Hmm.

tests/framework/artifact/test_artifact.py
```diff
@@ -188,21 +205,72 @@ def test_artifact_write(hdf_mock, keys_mock):
     assert set(a.keys) == set(initial_keys + [key])
 
 
+def test_artifact_write_and_load_with_different_key_types(hdf_mock, keys_mock):
+    path = Path(Path('/place/with/artifact.hdf'))
```
collijk: Double cast to path?

PR #148, opened 2020-02-10
Update population manager and engine to invert the control of populat…

collijk:
This is a WIP/prototype for delegating the control of population creation to the components.  I feel it's closer to the patterns used elsewhere in the framework for builder interfaces and it solves an issue that has always annoyed me: The population configuration for initial population size has been a part of the manager and not part of the base population generation class.

Tests, docstrings, and updated docs to follow if we like this direction.

Please have a look and drop any questions you have and we'll follow up in person.

### DIFF
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -117,8 +117,6 @@ class SimulationContext:
         self.configuration.freeze()
         self._component_manager.setup_components(self._builder)
 
-        self.simulant_creator = self._builder.population.get_simulant_creator()
-
         self.time_step_events = self._lifecycle.get_state_names('main_loop')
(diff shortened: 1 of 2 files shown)


src/vivarium/framework/engine.py
```diff
@@ -129,18 +127,15 @@ def setup(self):
 
     def initialize_simulants(self):
         self._lifecycle.set_state('population_creation')
-        pop_params = self.configuration.population
-        # Fencepost the creation of the initial population.
```
KjellSwedin: What entity takes care of the "fenceposting" going forward?

collijk:
Ticket for the full implementation, testing, and doc updates (likely to split in planning): https://jira.ihme.washington.edu/browse/MIC-1616

PR #150, opened 2020-02-12
move get_seed to randomness interface

rmudambi:


src/vivarium/framework/randomness.py
```diff
@@ -627,10 +611,10 @@ def setup(self, builder):
 
         self.resources = builder.resources
         self._add_constraint = builder.lifecycle.add_constraint
-        builder.lifecycle.add_constraint(self.get_randomness_stream, allow_during=['setup'])
-        builder.lifecycle.add_constraint(self.register_simulants,
-                                         restrict_during=['initialization', 'setup', 'post_setup',
-                                                          'simulation_end', 'report'])
+        self._add_constraint(self.get_seed, restrict_during=['initialization'])
```
rmudambi: Can we extract these stage names as constants so we don't have strings everywhere?

src/vivarium/framework/randomness.py
```diff
@@ -471,22 +471,6 @@ def get_draw(self, index: Index, additional_key: Any = None) -> pd.Series:
 
         return draw
 
-    def get_seed(self, additional_key: Any = None) -> int:
```
rmudambi: Is it better to deprecate rather than remove this?

collijk: In theory, yes.  We don't have a deprecation strategy (and is one of the reasons we're not at a v1 release).  So don't worry about it for now.

src/vivarium/framework/randomness.py
collijk: Yes, but smartly.  Make a ticket in the next sprint backlog to survey string constants across vivarium and vph and then we'll come up with a strategy for handling categories of string constants.

src/vivarium/framework/randomness.py
collijk: https://jira.ihme.washington.edu/browse/MIC-1614

src/vivarium/framework/randomness.py
collijk: https://jira.ihme.washington.edu/browse/MIC-1615

PR #151, opened 2020-02-18
remove random_seed from idx, add seed and draw to columns

halfhorst:
Local sim results now match the output from psimulate

### DIFF
--- a/src/vivarium/interface/cli.py
+++ b/src/vivarium/interface/cli.py
@@ -87,9 +87,10 @@ def run(model_specification, results_directory, verbose, with_debugger):
     override_configuration = {'output_data': {'results_directory': str(results_root)}}
     finished_sim = main(model_specification, configuration=override_configuration)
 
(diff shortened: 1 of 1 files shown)


src/vivarium/interface/cli.py
```diff
@@ -87,9 +87,10 @@ def run(model_specification, results_directory, verbose, with_debugger):
     override_configuration = {'output_data': {'results_directory': str(results_root)}}
     finished_sim = main(model_specification, configuration=override_configuration)
 
-    idx = pd.Index([finished_sim.configuration.randomness.random_seed], name='random_seed')
-    metrics = pd.DataFrame(finished_sim.report(), index=idx)
+    metrics = pd.DataFrame(finished_sim.report(), index=[0])
     metrics['simulation_run_time'] = time() - start
+    metrics['random_seed'] = finished_sim.configuration.randomness.random_seed
+    metrics['input_draw'] = finished_sim.configuration.input_data.input_draw_number
```
halfhorst: remove

PR #152, opened 2020-02-28
swap try-except for explicit raise in popview get

halfhorst:
The old try-except didn't work because a KeyError was not thrown on missing columns (unless only one column was passed).


src/vivarium/framework/population.py
```diff
@@ -10,6 +10,7 @@
 state safely during runtime.
 
 """
+import warnings
```
collijk: Don't need warnings

PR #155, opened 2020-03-11
Fix typing issues in component manager

collijk:


src/vivarium/framework/components/manager.py
```diff
@@ -296,7 +296,7 @@ def get_component(self, name: str) -> Any:
         """
         return self._manager.get_component(name)
 
-    def get_components_by_type(self, component_type: Type) -> List[Any]:
+    def get_components_by_type(self, component_type: Union[type, Tuple[type, ...]]) -> List[Any]:
```
collijk: The intention is that the argument here is the same as you'd pass as the second argument to `isinstance` cause that's what actually happens.

PR #158, opened 2020-03-20
Add code and tests for mapping strategy

collijk:
Starting to move results system piecemeal into production code.

This PR contains an object representing a mapping strategy, which replaces the tuple representation of the mappers in the original prototype.  It has been expanded to work with mapping multiple columns to a single column.


tests/framework/results/test_strategies.py
```diff
@@ -0,0 +1,70 @@
...
+            return date.dt.year
+        else:
+            return date.year
+
+    strategy = MappingStrategy('design_date', 'design_year', to_year, is_vectorized=is_vectorized)
+    new_data = strategy(data)
+    check_columns(data, new_data, 'design_year')
+    assert data.design_date.dt.year.equals(new_data.design_year.astype(int))
+
+
```
rmudambi: Do we want a multi-column to int test?

PR #159, opened 2020-03-20
Add binning strategy

collijk:
this PR introduces the binning strategy, a special case of a mapping strategy for binning continuous data into categorical data.


src/vivarium/framework/results/strategies.py
```diff
@@ -61,3 +61,41 @@ def __call__(self, population: pd.DataFrame) -> pd.DataFrame:
...
+
+    """
+
+    def __init__(self, target: str, binned_column: str, bins: List[Union[int, float, pd.Timestamp]],
+                 labels: List[str], **cut_kwargs):
+        """
+        Parameters
+        ----------
+        target
+            The name of a true state table column or value pipeline to be
```
KjellSwedin: For both MappingStrategy and BinningStrategy would "source" or "input" be more appropriate than "target"?

PR #162, opened 2020-03-20
Add dict formatting strategy

collijk:
Provides the first (and our only, for now) formatting strategy for results, which process a set of aggregates into our long key_name value pairs where the key names end up as the output.hdf columns.


src/vivarium/framework/results/strategies.py
```diff
@@ -166,3 +166,25 @@ def _broadcast_aggregates(aggregate_data: pd.Series) -> pd.DataFrame:
     @abc.abstractmethod
     def __call__(self, aggregate_data: pd.Series):
         return self._broadcast_aggregates(aggregate_data)
+
+
+class DictFormattingStrategy(FormattingStrategy):
+    """Formatting strategy to produce a dictionary results from aggregates."""
+
+    def __call__(self, aggregate_data: pd.Series) -> Dict[str, float]:
```
collijk: Test for this to come after an implementation of a results producer.

PR #164, opened 2020-03-20
Results producer strategy

collijk:


src/vivarium/framework/results/strategies.py
```diff
@@ -140,6 +141,53 @@ def expand_data(self, data: pd.DataFrame) -> pd.DataFrame:
...
+        """
+        self._measure = measure
+        self._aggregator = aggregator
+        self._additional_keys = additional_keys
+
+    def __call__(self, data: GroupBy) -> Result:
+        """Aggregates the provided data groupby."""
+        result = data.apply(self._aggregator).sort_index()
+        result.name = 'value'
+        return Result(self._measure, result, self._additional_keys)
```
KjellSwedin: "value" shows up a lot. Perhaps it should be a constant/global

PR #183, opened 2021-08-10
fix error in deploy.yml

rmudambi:


.github/workflows/deploy.yml
```diff
@@ -10,14 +10,16 @@ jobs:
...
       - uses: actions/checkout@v2
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install dependencies
+      - name: Print environment values
+        run: |
+          cat $GITHUB_ENV
       - name: Update pip
```
KjellSwedin, line 23: What does this fix? Was having an empty step causing a problem?

rmudambi: Yes, this was the error from the last deploy attempt. 

image

.github/workflows/deploy.yml
```diff
@@ -10,14 +10,16 @@ jobs:
...
       matrix:
-        python-version: [ 3.6, 3.7, 3.8 ]
+        python-version: [3.6, 3.7, 3.8]
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install dependencies
```
collijk, line 20: This whole block should be identical to the build.yml

PR #187, opened 2021-08-17
MIC-2556: Results of running Black

mattkappel:
This PR is simply the results of running [black](https://black.readthedocs.io/en/stable/) on the repository.

### DIFF
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -42,16 +42,24 @@ from .results import ResultsInterface
 
 
 class SimulationContext:
-
-    def __init__(self, model_specification: Union[str, Path, ConfigTree] = None,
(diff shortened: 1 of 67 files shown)


src/vivarium/framework/engine.py
```diff
@@ -223,18 +266,37 @@ class Builder:
     def __init__(self, configuration, plugin_manager):
```
KjellSwedin, line 266: Umm, this is ugly. Maybe the line length specifier would get around this.

KjellSwedin (approved):
I think I could get used to this. Black does allow you to specify line length quite easily and that may have made a difference for some of the changes in this diff.
`-l 120`
would get you 120 char line lengths. I *think* the default is 88

mattkappel:
I aimed this PR against main accidentally.

PR #211, opened 2022-05-21
Feature/empty population

rmudambi:
### Description

- *Category*: feature
- *JIRA issue*: [MIC-3090](https://jira.ihme.washington.edu/browse/MIC-3090)

Allow for an empty initial population

### Testing
Tested running in MNCH Child model
Added unit tests

### DIFF
--- a/src/vivarium/framework/population.py
+++ b/src/vivarium/framework/population.py
@@ -189,7 +189,7 @@ class PopulationView:
             else:
                 return pop.loc[:, columns]
 
-    def update(self, population_update: Union[pd.DataFrame, pd.Series]):
+    def update(self, population_update: Union[pd.DataFrame, pd.Series]) -> None:
         """Updates the state table with the provided data.
 
         Parameters
@@ -211,15 +211,12 @@ class PopulationView:
             type inconsistent with the original population data.
 
         """
-        if population_update.empty:
-            return
-
         # TODO: Cast series to data frame and clean this up.
         if isinstance(population_update, pd.Series):
             if population_update.name in self._columns:
-                affected_columns = [population_update.name]
+                affected_columns = {population_update.name}
             elif len(self._columns) == 1:
-                affected_columns = self._columns
+                affected_columns = set(self._columns)
             else:
                 raise PopulationError(
                     "Cannot update with a pandas series unless the series name is a column "
@@ -234,13 +231,18 @@ class PopulationView:
                 )
             affected_columns = set(population_update.columns)
 
-        affected_columns = set(affected_columns).intersection(self._columns)
+        if population_update.empty and not affected_columns.difference(self._manager.columns):
+            return
+
         state_table = self._manager.get_population(True)
         if not self._manager.growing:
-            affected_columns = set(affected_columns).intersection(state_table.columns)
+            affected_columns = affected_columns.intersection(state_table.columns)
 
         for affected_column in affected_columns:
             if affected_column in state_table:
+                if population_update.empty:
+                    continue
+
                 new_state_table_values = state_table[affected_column].values
                 if isinstance(population_update, pd.Series):
                     update_values = population_update.values
@@ -428,6 +430,13 @@ class PopulationManager:
         metrics["total_population_tracked"] = len(tracked)
         metrics["total_population"] = len(untracked) + len(tracked)
         return metrics
+
+    @property
+    def columns(self) -> List[str]:
+        """The columns that currently exist in the population manager's
+        representation of state.
+        """
+        return list(self._population.columns)
 
     def __repr__(self):
         return "PopulationManager()"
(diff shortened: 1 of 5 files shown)


mattkappel (approved):

albrja (approved):

src/vivarium/framework/randomness.py
```diff
@@ -84,19 +84,20 @@ def update(self, new_keys: Index):
         if not self._map.index.intersection(new_keys).empty:
             raise KeyError("Non-unique keys in index.")
 
-        mapping_update = self.hash_(new_keys)
-        if self._map.empty:
-            self._map = mapping_update.drop_duplicates()
-        else:
-            self._map = pd.concat([self._map, mapping_update]).drop_duplicates()
+        if not new_keys.empty:
```
collijk, line 87: I think it's probably more readable to do 
```
if new_keys.empty:
    return  # Nothing to do
elif not self._map.index.intersection(new_keys).empty:
    raise KeyError("Non-unique keys in index")

...old method body...
```
[`Flat is better than nested`](https://peps.python.org/pep-0020/)

src/vivarium/framework/randomness.py
```diff
@@ -157,7 +158,9 @@ def convert_to_ten_digit_int(self, column: pd.Series) -> pd.Series:
             numeric.
 
         """
-        if isinstance(column.iloc[0], datetime.datetime):
+        if len(column) == 0:
```
collijk, line 161: Under what circumstances do you get here with an empty series?

collijk: The only call is from `update`, which should short-circuit, so we don't need/want the extra branch here. as a non-empty series is a precondition for calling.

src/vivarium/framework/population.py
```diff
@@ -234,13 +230,18 @@ def update(self, population_update: Union[pd.DataFrame, pd.Series]):
                 )
             affected_columns = set(population_update.columns)
 
-        affected_columns = set(affected_columns).intersection(self._columns)
```
collijk, line 237: I think removing this is going to have serious unintended consequences.

src/vivarium/framework/randomness.py
rmudambi: We get that when we have an empty population.

src/vivarium/framework/population.py
rmudambi: It's not possible for affected columns to contain columns not in the view, since an error would have been thrown above. It was possible for this to happen if a Series was passed, but I've closed that loophole now.

src/vivarium/framework/randomness.py
rmudambi: Hmm I was reaching this with the empty population, but now I'm not. Looking at the code I agree it should short-circuit in `update`.

src/vivarium/framework/population.py
```diff
@@ -211,41 +210,41 @@ def update(self, population_update: Union[pd.DataFrame, pd.Series]):
...
+        affected_columns = set(population_update.columns)
+
+        if not affected_columns.issubset(self.columns):
+            raise PopulationError(
+                f"Cannot update with a DataFrame or Series that contains columns the view does "
+                f"not. Dataframe contains the following extra columns: "
+                f"{set(population_update.columns).difference(self.columns)}."
+            )
+
+        if population_update.empty and not affected_columns.difference(self._manager.columns):
```
hussain-jafari, line 234: I'm not sure why we would be passing in an empty population update/if that would have any effect on the population. As a result I'm also not sure why this boolean is making the second check. I read it as checking that there are no columns in population_update that aren't in the state table, but why is this relevant if population_update is empty?

src/vivarium/framework/population.py
```diff
@@ -211,41 +210,41 @@ def update(self, population_update: Union[pd.DataFrame, pd.Series]):
...
-                )
-        else:
-            if not set(population_update.columns).issubset(self._columns):
-                raise PopulationError(
-                    f"Cannot update with a DataFrame that contains columns the view does not. "
-                    f"Dataframe contains the following extra columns: "
-                    f"{set(population_update.columns).difference(self._columns)}."
-                )
-            affected_columns = set(population_update.columns)
+            if population_update.name is None:
```
collijk, line 214: Can you break this out as a `coerce_to_df` (or similar) method?

collijk: Just did it: https://github.com/ihmeuw/vivarium/pull/212

src/vivarium/framework/randomness.py
```diff
@@ -84,6 +84,11 @@ def update(self, new_keys: Index):
         if not self._map.index.intersection(new_keys).empty:
             raise KeyError("Non-unique keys in index.")
 
+        if new_keys.empty:
+            return  # Nothing to do
+        elif not self._map.index.intersection(new_keys).empty:
```
collijk, line 89: This is duplicate with line 84

PR #214, opened 2022-06-02
Clean up population system logic

collijk:
### Description

- *Category*: refactor/test
- *JIRA issue*: [MIC-3135](https://jira.ihme.washington.edu/browse/MIC-3135)

This update tries to clean up logic in the population view to make it more
clear what's going on.  It still kind of sucks.  There's a design flaw in the 
way we do population management that I now have some ideas for, but 
is beyond the scope.  Instead, we try better naming, isolation of 
code into methods, a ton of docstrings/comments, and a boatload of 
new tests.

### Testing
Like a hundred new tests.

### DIFF
--- a/src/vivarium/framework/population/population_view.py
+++ b/src/vivarium/framework/population/population_view.py
@@ -28,6 +28,7 @@ class PopulationView:
     PopulationView can only read and write columns for which it is configured.
     Attempts to update non-existent columns are ignored except during
     simulant creation when new columns are allowed to be created.
+
     Parameters
     ----------
     manager
@@ -38,10 +39,12 @@ class PopulationView:
     query
         A :mod:`pandas`-style filter that will be applied any time this
         view is read from.
+
     Notes
     -----
     By default, this view will filter out ``untracked`` simulants unless
     the ``tracked`` column is specified in the initialization arguments.
+
     """
 
     def __init__(
@@ -63,10 +66,12 @@ class PopulationView:
     @property
     def columns(self) -> List[str]:
         """The columns that the view can read and update.
+
         If the view was created with ``None`` as the columns argument, then
         the view will have access to the full table by default. That case
         should be only be used in situations where the full state table is
         actually needed, like for some metrics collection applications.
+
         """
         if not self._columns:
             return list(self._manager.get_population(True).columns)
@@ -75,27 +80,33 @@ class PopulationView:
     @property
     def query(self) -> Union[str, None]:
         """A :mod:`pandas` style query to filter the rows of this view.
+
         This query will be applied any time the view is read. This query may
         reference columns not in the view's columns.
+
         """
         return self._query
 
     def subview(self, columns: Union[List[str], Tuple[str]]) -> "PopulationView":
         """Retrieves a new view with a subset of this view's columns.
+
         Parameters
         ----------
         columns
             The set of columns to provide access to in the subview. Must be
             a proper subset of this view's columns.
+
         Returns
         -------
         PopulationView
             A new view with access to the requested columns.
+
         Raises
         ------
         PopulationError
             If the requested columns are not a proper subset of this view's
             columns or no columns are requested.
+
         Notes
         -----
         Subviews are useful during population initialization. The original
@@ -104,6 +115,7 @@ class PopulationView:
         requesting a subview, a component can read the sections it needs
         without running the risk of trying to access uncreated columns
         because the component itself has not created them.
+
         """
 
         if not columns or set(columns) - set(self.columns):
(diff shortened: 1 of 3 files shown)


src/vivarium/framework/population/population_view.py
```diff
@@ -208,46 +205,143 @@ def update(self, population_update: Union[pd.DataFrame, pd.Series]) -> None:
...
+        TypeError
+            If the population update is not a :class:`pandas.Series` or a
+            :class:`pandas.DataFrame`.
+        PopulationError
+            If the update violates any preconditions relevant to the context in which
+            the update is provided (initial population creation, population creation on
+            time steps, or population state changes on time steps).
+
+        """
+        assert not (creating_initial_population and not adding_simulants)
```
mattkappel, line 306: Nit: legibility for the Boolean expression.

`not (creating_initial_population and not adding_simulants)`

vs

`not creating_initial_population or adding_simulants`

mattkappel, line 306: Nit: legibility for the Boolean expression.

`not (creating_initial_population and not adding_simulants)`

vs

`not creating_initial_population or adding_simulants`

mattkappel (approved):

src/vivarium/framework/population/population_view.py
```diff
@@ -208,46 +205,143 @@ def update(self, population_update: Union[pd.DataFrame, pd.Series]) -> None:
...
+
+        population_update = PopulationView._coerce_to_dataframe(
+            population_update,
+            view_columns,
+        )
+
+        unknown_simulants = population_update.index.difference(state_table.index)
```
rmudambi, line 313: Could do the len call here, since we're not using the index again.

src/vivarium/framework/population/population_view.py
```diff
@@ -281,16 +384,120 @@ def _coerce_to_dataframe(
...
+                f"A component is missing updates for {missing_pops} simulants."
+            )
+        new_columns = set(population_update).difference(state_table)
+        overlapping_columns = set(population_update).intersection(state_table)
+        if not new_columns:
+            raise PopulationError(
+                f"A component is providing a population update for {list(population_update)} "
+                "but all provided columns are initialized by other components."
+            )
+        for column in overlapping_columns:
```
rmudambi, line 444: Why  do we want to allow this?

ETA: Won't we throw an error in this case later anyway in line 334?

src/vivarium/framework/population/population_view.py
```diff
@@ -281,16 +384,120 @@ def _coerce_to_dataframe(
...
+        if missing_pops:
+            raise PopulationError(
+                f"Components should initialize the same population at the simulation start. "
+                f"A component is missing updates for {missing_pops} simulants."
+            )
+        new_columns = set(population_update).difference(state_table)
+        overlapping_columns = set(population_update).intersection(state_table)
+        if not new_columns:
+            raise PopulationError(
+                f"A component is providing a population update for {list(population_update)} "
```
rmudambi, line 441: Is it possible to get information about what components are trying to set the same columns? This would make debugging much easier.

src/vivarium/framework/population/population_view.py
```diff
@@ -208,46 +205,143 @@ def update(self, population_update: Union[pd.DataFrame, pd.Series]) -> None:
...
+        else:
+            new_columns = list(set(population_update).difference(state_table))
+            if new_columns:
+                raise PopulationError(
+                    f"Attempting to add new columns {new_columns} to the state table "
+                    f"outside the initial population creation phase."
+                )
+
+            if adding_simulants:
+                for column in population_update:
```
rmudambi, line 332: It would be better to loop through and get all columns where the conditional below holds, and write out an error message with all of them, so you can get them all at once if you have multiple, rather than getting them one at a time and having to re-run to learn about the next one.

rmudambi (approved):

src/vivarium/framework/population/population_view.py
collijk: Yes, but not without an amount of introspection/reflection here that would be quite gross (without access to the caller directly, we need to use `inspect` to look at prior stack frames).  And I disagree that it would be *much* easier in a typical model building scenario where you can just grep files for a column name.

src/vivarium/framework/population/population_view.py
collijk: Done, but this helps less than you think it would as the error is triggered by one call from one component (which means we can still get it multiple times)

src/vivarium/framework/population/population_view.py
collijk: I was actually being too restrictive in 334 and have updated the logic to be consistent with what's here.

rmudambi (approved):

PR #220, opened 2022-08-10
Add get_output_model_name_string()

mattkappel:
### Description
- *Category*: feature
- *JIRA issue*:  [MIC-3229](https://jira.ihme.washington.edu/browse/MIC-3229) and [MIC-3091](https://jira.ihme.washington.edu/browse/MIC-3091)

Changes:
- Add `get_output_model_name_string()` to be a standard way for both `psimulate run` and `simulate run` to create the same results paths.
- If an artifact is provided at the CLI, use the location there, if not, check the model specification YAML for an artifact and use its location. If there is no artifact (valid use case!), use the stem of the model specification filename.

### Testing
Ran this VCT and Vivarium against IV Iron with an artifact provided only in the YAML and only at the command and @aflaxman's WA DOH staffing repo, which does not have an artifact and motivated one of these JIRA tickets. In each case, the correct artifact (if any) and results path location were as expected (`"south_asia"` in the former case and `"staffing"` in the latter, from `staffing.yaml`

### DIFF
--- a/src/vivarium/interface/utilities.py
+++ b/src/vivarium/interface/utilities.py
@@ -107,11 +107,11 @@ def raise_if_not_setup(system_type):
     return method_wrapper
 
 
-def get_output_location_like_string(
+def get_output_model_name_string(
     artifact_path: Union[str, Path],
     model_spec_path: Union[str, Path],
 ) -> str:
-    """Find a good string to use as model_name or location in output path creation.
+    """Find a good string to use as model name in output path creation.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


src/vivarium/interface/utilities.py
```diff
@@ -105,9 +107,43 @@ def wrapped_method(*args, **kwargs):
...
+    model_spec_path
+        Path to the model specification file. This must exist.
+
+    Returns
+    -------
+    str
+        A location-like string for use in output labeling.
+    """
+    if artifact_path:
+        location = Path(artifact_path).stem
```
collijk, line 129: There's no reason beyond convention that an artifact is going to be labelled by a location.  Locations are a (GBD) public health specific identifier and have no real place in `vivarium` proper.  What you're creating here is a formal logic for model names which says first use the artifact name and if that doesn't exist, use the model spec name.  None of that is specific to locations.

collijk (changes requested):
Do not reference the GBD/public health specific idea of locations in `vivarium`

src/vivarium/interface/utilities.py
mattkappel: Is the concern with the word "location"? Would the concern be satisfied if I did `s/location/location_like_thing/g`?

collijk: Yes.  "Location" has is not a formal concept in vivarium.  It's a formal concept in GBD.  `vivarium` is a generic simulation framework and we don't want leaky boundaries.

collijk: You want something like `model_name`

albrja (approved):

rmudambi (approved):

PR #223, opened 2022-08-30
Add Stratification class, tests, and docstrings

mattkappel:
### Description
- *Category*: feature
- *JIRA issue*: [MIC-3132](https://jira.ihme.washington.edu/browse/MIC-3132)

#### Changes
- Adds Stratification class and fixes small bug in the non-vectorized flow
- Adds docstrings to document the Stratification class
- Adds 10 pytest tests, for expected outputs and expected exception cases

### Testing
Ran new pytests successfully.

### DIFF
--- a/src/vivarium/framework/results/stratification.py
+++ b/src/vivarium/framework/results/stratification.py
@@ -45,6 +45,10 @@ class Stratification:
                 )
             self.mapper = self._default_mapper
             self.is_vectorized = True
+        if not len(self.categories):
+            raise ValueError("The categories argument must be non-empty.")
+        if not len(self.sources):
+            raise ValueError("The sources argument must be non-empty.")
 
     def __call__(self, population: pd.DataFrame) -> pd.DataFrame:
         if self.is_vectorized:
--- a/src/vivarium/framework/results/stratification.py
+++ b/src/vivarium/framework/results/stratification.py
@@ -1,0 +1,66 @@
+from dataclasses import dataclass
+from typing import Callable, List, Union
+
+import pandas as pd
+from pandas.api.types import CategoricalDtype
+
+
+@dataclass
+class Stratification:
+    """Class for stratifying observed quantities by specified characteristics
+
+    Each Stratification represents a set of mutually exclusive and collectively
+    exhaustive categories into which simulants can be assigned.
+
+    The `Stratification` class has five fields: `name`, `source`, `mapper`,
+    `categories`, and `is_vectorized`. The `name` is the name of the column
+    created by the mapper. The `source` is a list of columns in the extended
+    state table that are the inputs to the mapper function.  Simulants will
+    later be grouped by this column (or these columns) during stratification.
+    `categories` is a set of values that the mapper is allowed to output. The
+    `mapper` is the method that transforms the source to the name column.
+    The method produces an output column by calling the mapper on the source
+    columns. If the mapper is `None`, the default identity mapper is used. If
+    the mapper is not vectorized this is performed by using `pd.apply`.
+    Finally, `is_vectorized` is a boolean parameter that signifies whether
+    mapper function is applied to a single simulant (`False`) or to the whole
+    population (`True`).
+
+    `Stratification` also has a `__call__()` method. The method produces an
+    output column by calling the mapper on the source columns.
+    """
+
+    name: str
+    sources: List[str]
+    categories: List[str]
+    mapper: Callable[[Union[pd.Series, pd.DataFrame]], Union[str, pd.Series]] = None
+    is_vectorized: bool = False
+
+    def __post_init__(self):
+        if self.mapper is None:
+            if len(self.sources) != 1:
+                raise ValueError(
+                    f"No mapper provided for stratification {self.name} with"
+                    f"{len(self.sources)} stratification sources."
+                )
+            self.mapper = self._default_mapper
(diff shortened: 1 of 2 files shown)


albrja (approved):

tests/framework/results/test_stratification.py
```diff
@@ -0,0 +1,52 @@
+import numpy as np
+import pandas as pd
+import pytest
+
+from vivarium.framework.results.stratification import Stratification
+
+##########################
+# Mock data and fixtures #
+##########################
+NAME = "hogwarts_house"
```
stevebachmeier, line 10: I've never seen the movies :( 

(though I have read a few of the first books!)

stevebachmeier (approved):

src/vivarium/framework/results/stratification.py
```diff
@@ -0,0 +1,66 @@
...
+        if self.mapper is None:
+            if len(self.sources) != 1:
+                raise ValueError(
+                    f"No mapper provided for stratification {self.name} with"
+                    f"{len(self.sources)} stratification sources."
+                )
+            self.mapper = self._default_mapper
+            self.is_vectorized = True
+
+    def __call__(self, population: pd.DataFrame) -> pd.DataFrame:
```
rmudambi, line 49: Should variable name be `population` or would somethign like `source_data` be better?

src/vivarium/framework/results/stratification.py
```diff
@@ -0,0 +1,66 @@
...
+        if self.is_vectorized:
+            raw_mapped_column = self.mapper(population[self.sources])
+        else:
+            raw_mapped_column = population[self.sources].apply(self.mapper, axis=1)
+        mapped_column = raw_mapped_column.astype(
+            CategoricalDtype(categories=self.categories, ordered=True)
+        )
+        if mapped_column.isna().any():
+            invalid_categories = set(raw_mapped_column.unique()) - set(self.categories)
+            raise ValueError(f"Invalid values '{invalid_categories}' found in {self.name}.")
```
rmudambi, line 59: Need to test this case

src/vivarium/framework/results/stratification.py
```diff
@@ -0,0 +1,66 @@
...
+    name: str
+    sources: List[str]
+    categories: List[str]
+    mapper: Callable[[Union[pd.Series, pd.DataFrame]], Union[str, pd.Series]] = None
+    is_vectorized: bool = False
+
+    def __post_init__(self):
+        if self.mapper is None:
+            if len(self.sources) != 1:
+                raise ValueError(
```
rmudambi, line 42: Need to test this failure case

src/vivarium/framework/results/stratification.py
```diff
@@ -0,0 +1,66 @@
...
+                    f"No mapper provided for stratification {self.name} with"
+                    f"{len(self.sources)} stratification sources."
+                )
+            self.mapper = self._default_mapper
+            self.is_vectorized = True
+
+    def __call__(self, population: pd.DataFrame) -> pd.DataFrame:
+        if self.is_vectorized:
+            raw_mapped_column = self.mapper(population[self.sources])
+        else:
```
rmudambi, line 52: Need some edge cases:
What if is_vectorized is inconsistent with the mapper?
What if categories is empty?
What if sources is empty?
What if sources is not in population.columns?

Spend some time thinking about other possible edge cases/failure cases

tests/framework/results/test_stratification.py
```diff
@@ -0,0 +1,52 @@
...
+SOURCES = ["first_name", "last_name"]
+CATEGORIES = ["hufflepuff", "ravenclaw", "slytherin", "gryffindor"]
+STUDENT_TABLE = pd.DataFrame(
+    np.array([["harry", "potter"], ["severus", "snape"], ["luna", "lovegood"]]),
+    columns=SOURCES,
+)
+
+
+def sorting_hat_vector(state_table: pd.DataFrame) -> pd.Series:
```
rmudambi, line 19: Need to use the argument in this mapper

src/vivarium/framework/results/stratification.py
mattkappel: We definitely use `population` in a lot of places in vivarium and refer to discrete elements of that population as "simulants". I think we should keep `population` for consistency. After having done a refactor to `source_data`, you've then got a problem of describing what the `source_data` is, and generally, that's a population with simulants (humans, maybe, but could be boids or something else).

src/vivarium/framework/results/stratification.py
mattkappel: Agree will add!

src/vivarium/framework/results/stratification.py
mattkappel: Agree will add!

src/vivarium/framework/results/stratification.py
mattkappel: Agree will add several erroring cases, with possibly adding code to have the errors be handled properly!

mattkappel: > What if is_vectorized is inconsistent with the mapper?

This one is turning out to be a tricky case. It will almost certainly throw an exception, but in a way specific to that `mapper`'s implementation. I'll check for a generic `Exception` here.

PR #226, opened 2022-09-23
Implement register_stratification, underlying functionality, and add tests

mattkappel:
### Description
- *Category*: POC
- *JIRA issue*: [MIC-3133](https://jira.ihme.washington.edu/browse/MIC-3133)

#### Changes
- Implements ResultsInterface and ResultsManager register_stratification along with ResultsContext functionality
- Refactor ResultsInterface to be separate from ResultsManager, using `TYPE_CHECKING` to handle circular imports
- Refactor often-used mocks to its own file
- Adds tests for register_stratification with columns and pipelines as sources
- TODO added for error case (duplicate stratification) pending feedback

### Testing
Four new tests were added, all pass along with the pre-existing ones for the results system.

### DIFF
--- a/src/vivarium/framework/results/context.py
+++ b/src/vivarium/framework/results/context.py
@@ -4,9 +4,9 @@ from typing import Callable, Dict, List, Tuple
 import pandas as pd
 
 from vivarium.framework.results.exceptions import ResultsConfigurationError
+from vivarium.framework.results.stratification import Stratification
 
 
-# TODO: ResultsContext needs tests added to test_engine or elsewhere...
 class ResultsContext:
     """
     Object contained within the ResultsManager organizing observations and the stratifications they require.
@@ -21,6 +21,7 @@ class ResultsContext:
         # values are dicts with key (filter, grouper) value (measure, aggregator, additional_keys)
         self._observations = defaultdict(lambda: defaultdict(list))
 
+    # noinspection PyAttributeOutsideInit
     def set_default_stratifications(self, default_grouping_columns: List[str]):
         if self._default_stratifications:
             raise ResultsConfigurationError(
@@ -42,11 +43,10 @@ class ResultsContext:
         mapper: Callable,
         is_vectorized: bool,
     ):
-        # TODO: implement this with stratifications
-        # self._stratifications.append(
-        #     Stratification(name, sources, categories, mapper, is_vectorized)
-        # )
-        ...
+        stratification = Stratification(name, sources, categories, mapper, is_vectorized)
+        self._stratifications.append(
+            stratification
+        )
 
     def add_observation(
         self,
@@ -58,6 +58,7 @@ class ResultsContext:
         when: str = "collect_metrics",
         **additional_keys: str,
     ):
+        # XXX WTF IS THIS
         groupers = self._get_groupers(additional_stratifications, excluded_stratifications)
         (
             self._producers[when][(pop_filter, groupers)].append(
--- a/src/vivarium/framework/results/context.py
+++ b/src/vivarium/framework/results/context.py
@@ -9,9 +9,11 @@ from vivarium.framework.results.stratification import Stratification
 
 class ResultsContext:
     """
-    Object contained within the ResultsManager organizing observations and the stratifications they require.
+    Manager context for organizing observations and the stratifications they require.
 
-    TODO: add more details when implementing
+    This context object is wholly contained by the manager :class:`vivarium.framework.results.manager.ResultsManger`.
+    Stratifications can be added to the context through the manager via the
+    :meth:`vivarium.framework.results.context.ResultsContext.add_observation` method.
     """
 
     def __init__(self):
@@ -43,10 +45,31 @@ class ResultsContext:
         mapper: Callable,
         is_vectorized: bool,
     ):
+        """Add a stratification to the context.
+
+        Parameters
+        ----------
+        name
+            Name of the of the column created by the stratification.
+        sources
+            A list of the columns and values needed for the mapper to determinate
+            categorization.
+        categories
+            List of string values that the mapper is allowed to output.
+        mapper
+            A callable that emits values in `categories` given inputs from columns
+            and values in the `requires_columns` and `requires_values`, respectively.
+        is_vectorized
+            `True` if the mapper function expects a `DataFrame`, and `False` if it
+            expects a row of the `DataFrame` and should be used by calling :func:`df.apply`.
+
+
+        Returns
+        ------
+        None
+        """
         stratification = Stratification(name, sources, categories, mapper, is_vectorized)
-        self._stratifications.append(
-            stratification
-        )
+        self._stratifications.append(stratification)
 
     def add_observation(
         self,
@@ -58,7 +81,7 @@ class ResultsContext:
         when: str = "collect_metrics",
         **additional_keys: str,
     ):
-        # XXX WTF IS THIS
+        # TODO: _producers doesn't exist, stub code needs bugfix at observation implementation time
         groupers = self._get_groupers(additional_stratifications, excluded_stratifications)
         (
             self._producers[when][(pop_filter, groupers)].append(
--- a/src/vivarium/framework/results/context.py
+++ b/src/vivarium/framework/results/context.py
@@ -17,8 +17,8 @@ class ResultsContext:
     """
 
     def __init__(self):
-        self._default_stratifications = []  # type: List[str]
-        self._stratifications = []  # type: List[Stratification]
+        self._default_stratifications: List[str] = []
+        self._stratifications: List[Stratification] = []
         # keys are event names
         # values are dicts with key (filter, grouper) value (measure, aggregator, additional_keys)
         self._observations = defaultdict(lambda: defaultdict(list))
--- a/src/vivarium/framework/results/context.py
+++ b/src/vivarium/framework/results/context.py
@@ -68,6 +68,8 @@ class ResultsContext:
         ------
         None
         """
+        if len([s.name for s in self._stratifications if s.name == name]):
+            raise ValueError(f"Name `{name}` is already used")
         stratification = Stratification(name, sources, categories, mapper, is_vectorized)
         self._stratifications.append(stratification)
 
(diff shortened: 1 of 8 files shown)


src/vivarium/framework/results/interface.py
```diff
@@ -0,0 +1,122 @@
...
+    stratifications required by the model using :func:`register_default_stratifications`,
+    :func:`register_stratification`, and :func:`register_binned_stratification`
+    as necessary. A “binned stratification” is a stratification special case for
+    the very common situation when a continuous value needs to be binned into
+    categorical bins. The `is_vectorized` argument should be True if the mapper
+    function expects a DataFrame, and False if it expects a row of the DataFrame
+    and should be used by calling df.apply.
+    """
+
+    def __init__(self, manager: "ResultsManager") -> None:
```
stevebachmeier, line 42: I think with TYPE_CHECKING implemented you can lose the quotes? I might be wrong...still kinda confused between using quotes in type-checking and using the TYPE_CHECKING import.

albrja (approved):

albrja:
Nice work on the tests.  Helped my understanding of what was going on here.

src/vivarium/framework/results/interface.py
mattkappel: Good question, I think you need quotes for the `TYPE_CHECKING=False` case. https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

mattkappel:
I've added duplicate Stratification handling in the the context and a test. This PR is ready for full review. I would appreciate in particular review from @rmudambi and @ramittal. 

Please pay special attention to:

- Changes in the ResultsManager src/vivarium/framework/results/manager.py
- Changes in the ResultsContext src/vivarium/framework/results/context.py
- Additions to the test suite in tests/framework/results/test_context.py and tests/framework/results/test_manager.py

Please ignore:

- Refactored out ResultsInterface from the manager file: src/vivarium/framework/results/interface.py
-  Refactored out mocks from pre-existing tests: tests/framework/results/mocks.py and tests/framework/results/test_stratification.py

src/vivarium/framework/results/manager.py
```diff
@@ -38,7 +40,7 @@ def metrics(self):
         return self._metrics.copy()
 
     # noinspection PyAttributeOutsideInit
-    def setup(self, builder):
+    def setup(self, builder: "Builder"):
```
albrja, line 43: Do not need the "s

src/vivarium/framework/results/manager.py
```diff
@@ -130,6 +159,8 @@ def register_observation(
         )
 
     def _add_resources(self, target: List[str], target_type: str):
+        if not len(target):
+            return  # do nothing on empty lists
```
albrja, line 163: Curious on checks for empty lists/dataframes/etc and good practices for that.

src/vivarium/framework/results/manager.py
mattkappel: Yes, you do.

src/vivarium/framework/results/manager.py
mattkappel: Here, it's not strictly necessary, but it's a low-cost check as most `target` lists will be of length 1. Good practice is to consider the tradeoff of the performance hit of the check vs the benefit.

src/vivarium/framework/results/interface.py
```diff
@@ -0,0 +1,122 @@
...
+    backend by means of the builder object. It exposes methods
+    to register stratifications, set default stratifications, and register
+    results producers. There is a special case for stratifications generated
+    by binning continuous data into categories.
+
+    The expected use pattern would be for a single component to register all
+    stratifications required by the model using :func:`register_default_stratifications`,
+    :func:`register_stratification`, and :func:`register_binned_stratification`
+    as necessary. A “binned stratification” is a stratification special case for
+    the very common situation when a continuous value needs to be binned into
```
rmudambi, line 36: Nit: change to "... very common situation when a SINGLE continuous value..."

src/vivarium/framework/results/interface.py
```diff
@@ -0,0 +1,122 @@
...
+    categorical bins. The `is_vectorized` argument should be True if the mapper
+    function expects a DataFrame, and False if it expects a row of the DataFrame
```
rmudambi, line 38: Maybe be a bit more explicit and say "The `is_vectorized` argument should be True if the mapper function expects a DataFrame corresponding to the whole population, and False if it expects a row of the DataFrame corresponding to a single simulant."

src/vivarium/framework/results/interface.py
```diff
@@ -0,0 +1,122 @@
...
+    and should be used by calling df.apply.
+    """
+
+    def __init__(self, manager: "ResultsManager") -> None:
+        self._manager: "ResultsManager" = manager
```
rmudambi, line 43: Above we do typing with comments e.g. `self._default_stratifications = []  # type: List[str]`
We should decide on a consistent pattern.

src/vivarium/framework/results/manager.py
```diff
@@ -82,12 +85,38 @@ def register_stratification(
...
+
+        Returns
+        ------
+        None
+        """
         target_columns = list(requires_columns) + list(requires_values)
         self._results_context.add_stratification(
             name, target_columns, categories, mapper, is_vectorized
         )
+        self._add_resources(requires_columns, "column")
```
rmudambi, line 118: Should we use Enums rather than strings here? I tend not to like naked strings floating around.

rmudambi (commented):
Only looked at the src directory so far. Everything looks good with a few minor nits and some suggestions. Looking at the tests now

tests/framework/results/test_context.py
```diff
@@ -0,0 +1,86 @@
...
+            SOURCES,
+            CATEGORIES,
+            sorting_hat_serial,
+            False,
+        ),
+    ],
+)
+def test_add_stratification(name, sources, categories, mapper, is_vectorized):
+    ctx = ResultsContext()
+    ctx.add_stratification(name, sources, categories, mapper, is_vectorized)
```
rmudambi, line 39: Maybe `assert expected_object not in ctx._stratifications` before adding it?

src/vivarium/framework/results/context.py
```diff
@@ -42,11 +45,33 @@ def add_stratification(
...
+        is_vectorized
+            `True` if the mapper function expects a `DataFrame`, and `False` if it
+            expects a row of the `DataFrame` and should be used by calling :func:`df.apply`.
+
+
+        Returns
+        ------
+        None
+        """
+        if len([s.name for s in self._stratifications if s.name == name]):
```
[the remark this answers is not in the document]
rmudambi: Changing equality has implications for your tests below, so it's probably a bad idea to do this.

tests/framework/results/test_context.py
```diff
@@ -0,0 +1,86 @@
...
+    expected_object = Stratification(name, sources, categories, mapper, is_vectorized)
+    assert expected_object in ctx._stratifications
+
+
+@pytest.mark.parametrize(
+    "name, sources, categories, mapper, is_vectorized, expected_exception",
+    [
+        (  # map to a category that isn't in CATEGORIES
```
rmudambi, line 47: I think each of these cases should be a separate test, since they have nothing in common other than the fact that they should fail

tests/framework/results/test_context.py
```diff
@@ -0,0 +1,86 @@
...
+            NAME,
+            SOURCES,
+            CATEGORIES,
+            sorting_hat_bad_mapping,
+            False,
```
rmudambi, line 52: I'm very surprised this throws an exception. How do we know when adding a stratification that its mapper function can return something not in categories? I'm pleasantly surprised if this is actually what happens.

tests/framework/results/test_manager.py
```diff
@@ -0,0 +1,102 @@
...
+
+from .mocks import CATEGORIES, NAME, SOURCES, sorting_hat_serial, sorting_hat_vector
+
+
+# Mock for get_value call for Pipelines, returns a str instead of a Pipeline
+def mock_get_value(self, name: str):
+    return name
+
+
+def verify_stratification_added(mgr, name, categories, sources, mapper, is_vectorized):
```
rmudambi, line 15: I misread this and thought it was a test. Maybe add a comment to clarify it is not. 

Also, consider whether we want to define the `__eq__` function on `Stratification` to simplify this check. 

In `test_context.py` line 41 you make the assertion `assert expected_object in ctx._stratifications`, and I thought that `in` keyword is implicitly doing an equality check, so check if the default equality function is sufficient for this purpose and that the above test is correct.

tests/framework/results/test_manager.py
```diff
@@ -0,0 +1,102 @@
...
+        (  # expected Stratification for non-vectorized
+            NAME,
+            SOURCES,
+            CATEGORIES,
+            sorting_hat_serial,
+            False,
+        ),
+    ],
+)
+def test_register_stratification_with_pipelines(
```
rmudambi, line 81: Add a test where you register a stratification with both column and pipeline sources.

tests/framework/results/test_context.py
```diff
@@ -0,0 +1,86 @@
...
+        (  # is_vectorized=False with vectorized mapper
+            NAME,
+            SOURCES,
+            CATEGORIES,
+            sorting_hat_vector,
+            False,
+            Exception,
+        ),
+    ],
+)
```
rmudambi, line 80: It occurs to me that it should probably throw an error if someone tries to create a Stratification with duplicate categories - i.e. `categories=['foo', 'bar', 'foo']`. 

Let's create another ticket to handle that edge case and the relevant test case.

rmudambi (approved):

tests/framework/results/test_context.py
mattkappel: Created MIC-3475

tests/framework/results/test_manager.py
mattkappel: I recall why I did it the two different ways, but in retrospect, I don't think my reasoning was right. I'm going to refactor the more deliberate `verify_stratification_added` to be used by both `test_manager.py` and `test_context.py`

tests/framework/results/test_context.py
mattkappel: I removed these tests, as they were nonsensical and replaced it with the only real exception testing I could do for this bit.

tests/framework/results/test_context.py
mattkappel: Yeah, they were nonsense. I don't know why they passed though. Nevertheless, they've been removed and replaced.

PR #231, opened 2022-10-11
Prevent FutureWarnings

hussain-jafari:
### Description
- *Category*: bugfix
- *JIRA issue*: [MIC-3488](https://jira.ihme.washington.edu/browse/MIC-3488)

pandas will show a FutureWarning when passing a list of length 1 to groupby. There is also a FutureWarning about using df.loc[] the way we did in updating the population. 

### Testing
Ran a small simulation all the way through and checked that groupby objects produced the same results using new and old methods. Checked that change to the population view column update produced the same behavior as the old code.

### DIFF
--- a/src/vivarium/framework/state_machine.py
+++ b/src/vivarium/framework/state_machine.py
@@ -83,7 +83,8 @@ def _groupby_new_state(
 
     """
     output_map = {o: i for i, o in enumerate(outputs)}
-    groups = pd.Series(index).groupby([output_map[d] for d in decisions])
+    # Pass grouper as Series to avoid FutureWarning about passing list of length 1
+    groups = pd.Series(index).groupby(pd.Series([output_map[d] for d in decisions]))
     results = [(outputs[i], pd.Index(sub_group.values)) for i, sub_group in groups]
     selected_outputs = [o for o, _ in results]
     for output in outputs:
--- a/src/vivarium/interpolation.py
+++ b/src/vivarium/interpolation.py
@@ -68,7 +68,11 @@ class Interpolation:
         if self.key_columns:
             # Since there are key_columns we need to group the table by those
             # columns to get the sub-tables to fit
-            sub_tables = self.data.groupby(list(self.key_columns))
+            if len(self.key_columns) == 1:
+                # Pass element instead of list of length 1 to avoid FutureWarning
+                sub_tables = self.data.groupby(self.key_columns[0])
+            else:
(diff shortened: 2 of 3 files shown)


src/vivarium/framework/state_machine.py
```diff
@@ -83,7 +83,8 @@ def _groupby_new_state(
 
     """
     output_map = {o: i for i, o in enumerate(outputs)}
-    groups = pd.Series(index).groupby([output_map[d] for d in decisions])
+    # Pass grouper as Series to avoid FutureWarning about passing list of length 1
+    groups = pd.Series(index).groupby(pd.Series([output_map[d] for d in decisions]))
```
albrja, line 87: The index for the series in the groupby call just gets ignored I'm assuming?

hussain-jafari: The groupby documentation says "If a dict or Series is passed, the Series or dict VALUES will be used to determine the groups (the Series’ values are first aligned; see .align() method)".

src/vivarium/interpolation.py
```diff
@@ -68,7 +68,11 @@ def __init__(
         if self.key_columns:
             # Since there are key_columns we need to group the table by those
             # columns to get the sub-tables to fit
-            sub_tables = self.data.groupby(list(self.key_columns))
+            if len(self.key_columns) == 1:
+                # Pass element instead of list of length 1 to avoid FutureWarning
+                sub_tables = self.data.groupby(self.key_columns[0])
+            else:
+                sub_tables = self.data.groupby(list(self.key_columns))
```
albrja, line 75: I would just pass self.key_columns as a pd.Series like you did in the state_machine module or just be consistent with how you are handling this.

albrja (approved):

src/vivarium/framework/state_machine.py
rmudambi: I'm curious to hear how you figured out this would work

rmudambi (approved):

src/vivarium/framework/state_machine.py
hussain-jafari: I was thinking of how I could pass data containing a single element without a list and without passing the element directly, so I looked up the groupby documentation to see what data structures I could use and thought a dictionary and Series were promising. I first tried a dictionary, but that created a different groupby object than the original. A Series reproduced the old behavior.

PR #234, opened 2022-11-03
Implement register_observation

mattkappel:
### Description
- *Category*: POC
- *JIRA issue*: [MIC-3341](https://jira.ihme.washington.edu/browse/MIC-3341)

#### Changes
- Adds docstring to `interface.register_observation`
- Implements `ResultsManager.register_observation` and `ResultsContext.add_observation`
- Integration test added for `ResultsManager.register_observation` and unit test added for `ResultsContext.add_observation`
- Implements and tests `ResultsContext._get_groupers`

### Testing
All tests pass.

### DIFF
--- a/src/vivarium/framework/results/interface.py
+++ b/src/vivarium/framework/results/interface.py
@@ -172,5 +172,13 @@ class ResultsInterface:
         ------
         None
         """
-        self._manager.register_observation(name, pop_filter, aggregator, requires_columns, requires_values,
-                                           additional_stratifications, excluded_stratifications, when)
+        self._manager.register_observation(
+            name,
+            pop_filter,
+            aggregator,
+            requires_columns,
+            requires_values,
+            additional_stratifications,
+            excluded_stratifications,
+            when,
+        )
--- a/src/vivarium/framework/results/interface.py
+++ b/src/vivarium/framework/results/interface.py
@@ -166,7 +166,7 @@ class ResultsInterface:
             names to remove from the observation.
         when
             String name of the phase of a time-step the observation should happen. Valid values are:
-            ["time_step__prepare", "time_step", "time_step__cleanup", "collect_metrics"]
+            `"time_step__prepare"`, `"time_step"`, `"time_step__cleanup"`, `"collect_metrics"`.
 
         Returns
         ------
--- a/src/vivarium/framework/results/interface.py
+++ b/src/vivarium/framework/results/interface.py
@@ -143,4 +143,34 @@ class ResultsInterface:
         excluded_stratifications: List[str] = (),
         when: str = "collect_metrics",
     ) -> None:
-        self._manager.register_observation(...)
+        """Provide the results system all the information it needs to perform the observation.
+
+        Parameters
+        ----------
+        name
+            String name for the observation.
+        pop_filter
+            A Pandas query filter string to filter the population down to the simulants who should
+            be considered for the observation.
+        aggregator
+            A function that computes the quantity for the observation.
+        requires_columns
+            A list of the state table columns that are required by either the pop_filter or the aggregator.
+        requires_values
+            A list of the value pipelines that are required by either the pop_filter or the aggregator.
+        additional_stratifications
+            A list of additional :class:`stratification <vivarium.framework.results.stratification.Stratification>`
+            names by which to stratify.
+        excluded_stratifications
+            A list of default :class:`stratification <vivarium.framework.results.stratification.Stratification>`
+            names to remove from the observation.
+        when
+            String name of the phase of a time-step the observation should happen. Valid values are:
+            ["time_step__prepare", "time_step", "time_step__cleanup", "collect_metrics"]
+
+        Returns
+        ------
+        None
+        """
+        self._manager.register_observation(name, pop_filter, aggregator, requires_columns, requires_values,
+                                           additional_stratifications, excluded_stratifications, when)
--- a/src/vivarium/framework/results/manager.py
+++ b/src/vivarium/framework/results/manager.py
@@ -80,6 +80,7 @@ class ResultsManager:
         #     self._metrics.update(results_group)
 
     def set_default_stratifications(self, default_stratifications: List[str]):
+        # TODO: ought there be default default_stratifications? There is no interface for them in interface.py
         self._results_context.set_default_stratifications(default_stratifications)
 
     def register_stratification(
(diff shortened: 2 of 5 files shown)


src/vivarium/framework/results/interface.py
```diff
@@ -143,4 +143,42 @@ def register_observation(
         excluded_stratifications: List[str] = (),
         when: str = "collect_metrics",
```
stevebachmeier, line 144: The docs mention this could be a list. Would there ever be a desire to register the same stratifications at different times?

src/vivarium/framework/results/manager.py
```diff
@@ -180,8 +180,6 @@ def register_observation(
         excluded_stratifications: List[str] = (),
         when: str = "collect_metrics",
```
stevebachmeier, line 181: I still don't have the historical context here: are we defaulting to "collect_metrics" b/c that's simply the most common time to observe?

tests/framework/results/test_context.py
```diff
@@ -70,3 +67,116 @@ def test_add_stratification_raises(
...
+    [
+        (
+            "living_person_time",
+            'alive == "alive" and undead == False',
+            _aggregate_state_person_time,
+            [],
+            [],
+            "collect_metrics",
+        ),
+        (
```
stevebachmeier, line 88: Why are you doing two pop_filters? Is this stress testing something specific about `add_observation`?

tests/framework/results/test_context.py
```diff
@@ -70,3 +67,116 @@ def test_add_stratification_raises(
...
+            [],
+            "collect_metrics",
+        ),
+    ],
+    ids=["valid_on_collect_metrics"],
+)
+def test_double_add_observation(
+    name, pop_filter, aggregator, additional_stratifications, excluded_stratifications, when
+):
+    """Tests a double add of the same stratification, this should result in one additional observation being added to
```
stevebachmeier, line 133: Nit: This confused me at first b/c I thought you were saying that the second add should result in one additional (ie two total). Maybe something like this would be cleaner? "Tests a double add of the same stratification. Duplicate stratifications should be dropped and only one observation kept"

tests/framework/results/test_interface.py
```diff
@@ -0,0 +1,61 @@
...
+
+@pytest.mark.parametrize(
+    (
+        "name, pop_filter, aggregator, requires_columns, requires_values,"
+        " additional_stratifications, excluded_stratifications, when"
+    ),
+    [
+        (
+            "living_person_time",
+            'alive == "alive" and undead == False',
```
stevebachmeier, line 19: Interesting you're presumably considering undead folks as alive.

tests/framework/results/test_interface.py
```diff
@@ -0,0 +1,61 @@
...
+    # interface.set_default_stratifications(["age", "sex"])
+    assert len(interface._manager._results_context._observations) == 0
+    interface.register_observation(
+        name,
+        pop_filter,
+        aggregator,
+        additional_stratifications,
+        excluded_stratifications,
+    )
+    assert len(interface._manager._results_context._observations) == 1
```
stevebachmeier, line 61: I'm not that familiar yet w/ this design. This is testing that each of the parameterized tests adds an observation, but do you need to test that they're not overwriting each other? ie register two observations and ensure that the total ._observations == 2?

stevebachmeier (approved):
Looks good, though I'm particularly interested in my last comment (should you test that multiple registrations aren't overwriting each other or something like that)

tests/framework/results/test_context.py
albrja: It looks like this is just creating two different observers to test.

albrja (approved):

src/vivarium/framework/results/manager.py
mattkappel: I think so.

tests/framework/results/test_context.py
mattkappel: ^

src/vivarium/framework/results/interface.py
```diff
@@ -143,4 +143,42 @@ def register_observation(
...
-        self._manager.register_observation(...)
+        """Provide the results system all the information it needs to perform the observation.
+
+        Parameters
+        ----------
+        name
+            String name for the observation.
+        pop_filter
+            A Pandas query filter string to filter the population down to the simulants who should
+            be considered for the observation.
```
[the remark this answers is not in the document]
stevebachmeier: I was wondering about that myself but assumed it wasn't THAT different than, say, .loc. The difference was substantial in MNCH?

It seems possible (but rough) with string slicing and or implementing some other data structure sending to .loc or whatever.

tests/framework/results/test_context.py
rmudambi: Or "result in only one observation being added"

src/vivarium/framework/results/interface.py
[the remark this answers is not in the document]
rmudambi: This is a valid concern, and we can definitely consider replacing it with `.loc` calls in a future iteration of this or doing what Steve suggests to convert a `query ` call to a `.loc` call, but I'd consider that out of scope for now.

tests/framework/results/test_context.py
rmudambi: Also, isn't this a duplicate add of the same *observation* not stratification?

tests/framework/results/test_interface.py
rmudambi: I think it would be good to have a test in which you add two different observations, which I think is what Steve is suggesting here.

rmudambi (approved):

tests/framework/results/test_interface.py
mattkappel: I'll add this.

tests/framework/results/test_context.py
```diff
@@ -154,6 +154,33 @@ def test_double_add_observation(
...
+    ids=[
+        "additional_no_operation",
+        "exclude_no_operation",
+        "additional_and_exclude_no_operation",
+    ],
+)
+def test_add_observation_nop_stratifications(default, additional, excluded, match):
+    ctx = ResultsContext()
+    ctx._default_stratifications = default
+    with pytest.warns(UserWarning, match=match):
```
stevebachmeier, line 173: I suspect I just don't know how the pytest.warns context manager works, but what exactly happens if this fails?

mattkappel: It complains that a UserWarning was expected but didn't get made.

stevebachmeier: But don't you need to assert something about this or is that part of pytest.warns job?

mattkappel: Nope. https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.warns

PR #236, opened 2022-12-09
Implement pipeline caching

beatrixkh:
### Description
Adding caching functionality to pipeline calls within a timestep
- *Category*: feature
- *JIRA issue*: [MIC-3616](https://jira.ihme.washington.edu/browse/MIC-3616)
- [design propsal](https://github.com/ihmeuw/vivarium_development/blob/main/docs/source/design/2022_11_28_pipeline_caching.rst)

Added pipeline caching functionality:

- if passed in exactly one arg and no kwargs; or
- if passed in exactly one kwarg with key "index" and no kwargs:
- pipeline caches call in a pd.DataFrame stored in ValuesManager.call_history['pipeline_name']

- if any other combinations are passed in, doesn't cache, just calls pipeline as usual.
- empties out cache on timestep cleanup.

### Testing
stepped through code in a number of different scenarios and edge cases. will include pytests in a separate PR.

### DIFF
--- a/src/vivarium/framework/values.py
+++ b/src/vivarium/framework/values.py
@@ -287,13 +287,13 @@ class ValuesManager:
     def setup(self, builder):
         self.step_size = builder.time.step_size()
         builder.event.register_listener("post_setup", self.on_post_setup)
+        builder.event.register_listener("time_step__cleanup", self.on_timestep_cleanup, priority=9)
 
         self.resources = builder.resources
         self.add_constraint = builder.lifecycle.add_constraint
 
         builder.lifecycle.add_constraint(self.register_value_producer, allow_during=["setup"])
         builder.lifecycle.add_constraint(self.register_value_modifier, allow_during=["setup"])
-        builder.event.register_listener("time_step__cleanup", self.on_timestep_cleanup, priority=9)
 
     def on_post_setup(self, _):
         """Finalizes dependency structure for the pipelines."""
@@ -305,6 +305,7 @@ class ValuesManager:
 
         # register_value_producer and register_value_modifier record the
         # dependency structure for the pipeline source and pipeline modifiers,
+        # respectively.  We don't have enough information to record the
         # respectively.  We don't have enough information to record the
         # dependency structure for the pipeline itself until now, where
         # we say the pipeline value depends on its source and all its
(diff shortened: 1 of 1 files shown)


src/vivarium/framework/values.py
```diff
@@ -227,16 +227,42 @@ def __call__(self, *args, skip_post_processor=False, **kwargs):
...
+                self.manager.call_history[self.name] = self.manager.call_history[self.name].append(value)
+
+            value = self.manager.call_history[self.name].loc[newly_requested]
+
+        else:
+            value = self.source(*args, **kwargs)
+            for mutator in self.mutators:
+                value = self.combiner(value, mutator, *args, **kwargs)
+
         if self.post_processor and not skip_post_processor:
```
beatrixkh, line 266: I left this out of the caching but I think that was probably the wrong choice.

I think I should add it as a layer in the `call_history` dictionary; something like:

self.manager['pipeline_name'][(self.post_processor, self.skip_post_processor)] = pd.DataFrame()

src/vivarium/framework/values.py
```diff
@@ -227,16 +227,43 @@ def __call__(self, *args, skip_post_processor=False, **kwargs):
...
                 f"you are attempting to modify a value that hasn't been created."
             )
 
-        value = self.source(*args, **kwargs)
-        for mutator in self.mutators:
-            value = self.combiner(value, mutator, *args, **kwargs)
+        # if only one kwarg was passed in and it's an index, treat it as the arg
+        if (len(args) == 0) and (len(kwargs) == 1) and (isinstance(kwargs["index"], pd.Index)):
+            args = [kwargs["index"]]
+            kwargs = {}
```
albrja, line 240: Does this just make the logic here and processing easier?

src/vivarium/framework/values.py
```diff
@@ -227,16 +227,43 @@ def __call__(self, *args, skip_post_processor=False, **kwargs):
...
+            else:
+                self.manager.call_history[self.name] = pd.DataFrame()
+                newly_requested = args[0]
+            if newly_requested.any():
+                newly_requested = newly_requested.unique()
+                value = self.source(newly_requested)
+                for mutator in self.mutators:
+                    value = self.combiner(value, mutator, newly_requested)
+
+                self.manager.call_history[self.name] = self.manager.call_history[self.name].append(value)
```
albrja, line 258: What I am curious about is we update previously_called_idx each time we request a new index.  As someone that does not have a great understand of caching, does this mean if we request a subset of that index we would still have a cache for that or would we need to make that call once and then have the cache?

beatrixkh:
i'm sorry --- i found a bug in an assumption i had made.

PR #237, opened 2022-12-14
Feature/mic 3349

mattkappel:
### Description
- *Category*: POC
- *JIRA issue*: [MIC-3349](https://jira.ihme.washington.edu/browse/MIC-3349)

#### Changes
- Implements `gather_results` and `_prepare_population`, metrics count, population view, event listeners, and metric pipeline modifier at the Results Manager level
- Implements ResultsContext `gather_results` and `format_results`.
- Adds an `aggregator_columns` argument to meet the need of `sum` as a valid aggregator
- Adds an integration test for adding observations using the ResultsInterface
- Adds unit tests and exception tests for new methods in ResultsContext
- Adds docstrings where needed.

### Testing
All Results tests pass.

### DIFF
--- a/src/vivarium/framework/results/context.py
+++ b/src/vivarium/framework/results/context.py
@@ -94,19 +94,18 @@ class ResultsContext:
     def gather_results(self, population: pd.DataFrame, event_name: str) -> Dict[str, float]:
         # Optimization: We store all the producers by pop_filter and groupers
         # so that we only have to apply them once each time we compute results.
-        # TODO: uncomment and debug...
-        # for stratification in self._stratifications:
-        #     population = stratification(population)
-        #
-        # for (pop_filter, groupers), observations in self._observations[event_name].items():
-        #     # Results production can be simplified to
-        #     # filter -> groupby -> aggregate in all situations we've seen.
-        #     pop_groups = self.population.query(pop_filter).groupby(list(groupers))
-        #     for measure, aggregator, additional_keys in observers:
-        #         aggregates = pop_groups.apply(aggregator)
-        #         # Keep formatting all in one place.
-        #         yield self._format_results(measure, aggregates, **additional_keys)
-        ...
+        # TODO: XXX
+        for stratification in self._stratifications:
+            population = stratification(population)
+
+        for (pop_filter, groupers), observations in self._observations[event_name].items():
+            # Results production can be simplified to
+            # filter -> groupby -> aggregate in all situations we've seen.
+            pop_groups = population.query(pop_filter).groupby(list(groupers))
+            for measure, aggregator, additional_keys in observations:
+                aggregates = pop_groups.apply(aggregator)
+                # Keep formatting all in one place.
+                yield self._format_results(measure, aggregates, **additional_keys)
 
     def _get_groupers(
         self,
@@ -124,11 +123,14 @@ class ResultsContext:
     def _format_results(
         measure: str, aggregates: pd.DataFrame, **additional_keys: str
     ) -> Dict[str, float]:
+        # TODO: XXX
         results = {}
         # First we expand the categorical index over unobserved pairs.
         # This ensures that the produced results are always the same length.
         idx = pd.MultiIndex.from_product(
-            aggregates.index.levels, names=aggregates.index.names
+            #           aggregates.index.levels, names=aggregates.index.names
+            aggregates.index.categories,
+            names=aggregates.index.names,
         )
         data = pd.Series(data=0, index=idx)
         data.loc[aggregates.index] = aggregates
--- a/src/vivarium/framework/results/context.py
+++ b/src/vivarium/framework/results/context.py
@@ -86,38 +86,38 @@ class ResultsContext:
         **additional_keys: str,
     ):
         self._warn_check_stratifications(additional_stratifications, excluded_stratifications)
-        groupers = self._get_groupers(additional_stratifications, excluded_stratifications)
-        self._observations[when][(pop_filter, groupers)].append(
+        stratifications = self._get_stratifications(additional_stratifications, excluded_stratifications)
+        self._observations[when][(pop_filter, stratifications)].append(
             (name, aggregator, additional_keys)
         )
 
     def gather_results(self, population: pd.DataFrame, event_name: str) -> Dict[str, float]:
-        # Optimization: We store all the producers by pop_filter and groupers
+        # Optimization: We store all the producers by pop_filter and stratifications
         # so that we only have to apply them once each time we compute results.
         # TODO: XXX
         for stratification in self._stratifications:
             population = stratification(population)
 
-        for (pop_filter, groupers), observations in self._observations[event_name].items():
+        for (pop_filter, stratifications), observations in self._observations[event_name].items():
             # Results production can be simplified to
             # filter -> groupby -> aggregate in all situations we've seen.
-            pop_groups = population.query(pop_filter).groupby(list(groupers))
+            pop_groups = population.query(pop_filter).groupby(list(stratifications))
             for measure, aggregator, additional_keys in observations:
                 aggregates = pop_groups.apply(aggregator)
                 # Keep formatting all in one place.
                 yield self._format_results(measure, aggregates, **additional_keys)
 
-    def _get_groupers(
+    def _get_stratifications(
         self,
         additional_stratifications: List[str] = (),
         excluded_stratifications: List[str] = (),
     ) -> Tuple[str, ...]:
-        groupers = list(
+        stratifications = list(
             set(self._default_stratifications) - set(excluded_stratifications)
             | set(additional_stratifications)
         )
         # Makes sure measure identifiers have fields in the same relative order.
-        return tuple(sorted(groupers))
+        return tuple(sorted(stratifications))
 
     @staticmethod
     def _format_results(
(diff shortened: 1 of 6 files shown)


src/vivarium/framework/results/context.py
```diff
@@ -79,57 +79,78 @@ def add_observation(
...
+            event_name
+        ].items():
+            # Results production can be simplified to
+            # filter -> groupby -> aggregate in all situations we've seen.
+            if pop_filter:
+                population = population.query(pop_filter)
+            pop_groups = population.groupby(list(stratifications))
+            for measure, aggregator_sources, aggregator, additional_keys in observations:
+                if aggregator_sources:
+                    aggregates = pop_groups[aggregator_sources].apply(aggregator).fillna(0.0)
```
rmudambi, line 113: Did you actually see nans in any of your tests? I always want to be careful when we fill nans that we are not masking a bug when we do this.

src/vivarium/framework/results/context.py
```diff
@@ -79,57 +79,78 @@ def add_observation(
...
+                    if isinstance(aggregates, pd.DataFrame):
+                        aggregates = aggregates.squeeze(axis=1)
+                else:
+                    aggregates = pop_groups.apply(aggregator)
+                if not isinstance(aggregates, pd.Series):
+                    raise TypeError(
+                        f"aggregator return value is a {type(aggregates)} and could not be "
```
rmudambi, line 120: Nit: capitalize "a" in "aggregator"

rmudambi (commented):
I've only reviewed the framework code so far. Will review the tests tomorrow morning.

src/vivarium/framework/results/context.py
mattkappel: Yes, we see nans if there is no representation for the stratification in the population.

src/vivarium/framework/results/context.py
```diff
@@ -79,57 +79,78 @@ def add_observation(
...
+            if pop_filter:
+                population = population.query(pop_filter)
+            pop_groups = population.groupby(list(stratifications))
+            for measure, aggregator_sources, aggregator, additional_keys in observations:
+                if aggregator_sources:
+                    aggregates = pop_groups[aggregator_sources].apply(aggregator).fillna(0.0)
+                    if isinstance(aggregates, pd.DataFrame):
+                        aggregates = aggregates.squeeze(axis=1)
+                else:
+                    aggregates = pop_groups.apply(aggregator)
```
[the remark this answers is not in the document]
mattkappel: Yeah I think that's true. Good catch, I'll move that conditional squeeze before the type check.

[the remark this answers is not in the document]
albrja: I thought it was doing this.  That was a good catch.

src/vivarium/framework/results/context.py
```diff
@@ -138,14 +159,13 @@ def _format(field, param):
...
-                + [_format(field, measure) for field, param in zip(data.index.names, params)]
+                + [_format(field, param) for field, param in zip(data.index.names, params)]
                 # Sorts additional_keys by the field name.
-                + [
-                    _format(field, measure)
-                    for field, param in sorted(additional_keys.items())
-                ]
+                + [_format(field, param) for field, param in sorted(additional_keys.items())]
             )
             results[key] = val
```
albrja, line 170: I found this section a bit confusing but the takeaway is it is parsing the output data and building the results table?

src/vivarium/framework/results/context.py
```diff
@@ -138,14 +159,13 @@ def _format(field, param):
             return f"{str(field).upper()}_{param}"
 
         for params, val in data.iteritems():
```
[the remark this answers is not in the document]
mattkappel: `values`?

src/vivarium/framework/results/context.py
mattkappel: It's formatting the string for the specific stratification (e.g., `MEASURE_person_time_SEX_female_AGE_25_to_30_year_2045`)

src/vivarium/framework/results/context.py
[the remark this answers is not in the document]
mattkappel: `apply` will only give you a `pd.Series` or `pd.DataFrame`.

src/vivarium/framework/results/context.py
```diff
@@ -138,14 +161,16 @@ def _format(field, param):
...
+            if isinstance(params, str):  # handle single stratification case
```
rmudambi, line 164: Call this `categories` as well

tests/framework/results/test_context.py
```diff
@@ -74,6 +78,11 @@ def _aggregate_state_person_time(self, x: pd.DataFrame) -> float:
     return len(x) * (28 / 365.35)
 
 
+def _aggregate_state_person_time_function(x: pd.DataFrame) -> float:
```
rmudambi, line 81: `_aggregate_state_person_time` and `_aggregate_state_person_time_function` are identical. Is one of them supposed to be returning a callable?

tests/framework/results/test_context.py
```diff
@@ -196,14 +206,224 @@ def test_add_observation_nop_stratifications(default, additional, excluded, matc
...
+        aggregator,
+        stratifications,
+        [],
+        event_name,
+    )
+
+    i = 0
+    for _ in ctx.gather_results(population, event_name):
+        i += 1
+    assert i == 1
```
rmudambi, line 287: I don't understand what this assert is demonstrating. Can you add some comments to explain why this tests that `gather_results` works?

Shouldn't we also test that the observed values are correct?

tests/framework/results/test_context.py
```diff
@@ -196,14 +206,224 @@ def test_add_observation_nop_stratifications(default, additional, excluded, matc
...
+        pop_filter,
+        aggregator_sources,
+        aggregator,
+        stratifications,
+        [],
+        event_name,
+    )
+
+    for r in ctx.gather_results(population, event_name):
+        assert (
```
rmudambi, line 357: Correct me if I'm wrong, but this is testing that the stratifications containing unladen swallow are present in the results here? Shouldn't we also assert that they are 0?

src/vivarium/framework/results/context.py
mattkappel: That doesn't make sense for all the things, does it? (the measure and additional keys)?

tests/framework/results/test_context.py
mattkappel: Good catch, the difference was there for mocking into an object (the `self` param), but there's no apparent reason for that. I'll refactor that out!

tests/framework/results/test_context.py
mattkappel: It's demonstrating that we get a return value. It's not checking for correctness of the applied `aggregator` but that could be added.

tests/framework/results/test_context.py
mattkappel: Correct. We could assert that, probably, but I hesitated because assumption of the `aggregator`.

src/vivarium/framework/results/context.py
```diff
@@ -79,72 +79,100 @@ def add_observation(
...
-        #     pop_groups = self.population.query(pop_filter).groupby(list(groupers))
-        #     for measure, aggregator, additional_keys in observers:
-        #         aggregates = pop_groups.apply(aggregator)
-        #         # Keep formatting all in one place.
-        #         yield self._format_results(measure, aggregates, **additional_keys)
-        ...
-
-    def _get_groupers(
+        for stratification in self._stratifications:
+            population = stratification(population)
```
stevebachmeier, line 101: This is super confusing to me - it looks like you're just overwriting population over and over. Can you explain?

src/vivarium/framework/results/context.py
```diff
@@ -79,72 +79,100 @@ def add_observation(
...
+                population = population.query(pop_filter)
+            pop_groups = population.groupby(list(stratifications))
+            for measure, aggregator_sources, aggregator, additional_keys in observations:
+                if aggregator_sources:
+                    aggregates = pop_groups[aggregator_sources].apply(aggregator).fillna(0.0)
+                else:
+                    aggregates = pop_groups.apply(aggregator)
+
+                # Ensure we are dealing with a single column of formattable results
+                if isinstance(aggregates, pd.DataFrame):
```
stevebachmeier, line 118: nit: I don't think you need this if statement

src/vivarium/framework/results/context.py
mattkappel: It's successively applying the stratification to the population. For this, it's pass through in a way, it's not making a new column, but say we had a stratification of over and under 65 years of age. In that case, the stratification would produce a new column in the population that says over or under 65. Does that help?

src/vivarium/framework/results/context.py
mattkappel: We do I think. I want to squeeze DataFrames but not Series.

stevebachmeier (approved):

src/vivarium/framework/results/context.py
rmudambi: `series.squeeze()` will squeeze a Series of length 1 to a scalar which is not what we want. Also `series.squeeze()` doesn't take the `kwarg` `axis`.

PR #249, opened 2022-12-22
Add method to get sim state

collijk:
### Description

- *Category*: feature
- *JIRA issue*: N/A

Adds a method on the simulation lifecycle interface to provide other components 
and simulation managers a way to access the current simulation state.  In particular,
I'm planning on using this to clean up the way initial population creation vs new births
and migration are handeled.

### Testing
N/A, just an expansion of the interface, but no functionality.


rmudambi (approved):
This will be very helpful! We have resorted to hacks to get precisely this information in several recent simulations

albrja (approved):

PR #258, opened 2022-12-31
ENH: use pop_data.user_data.get pattern in examples.disease_model.BasePopulation

aflaxman:
### Description

- *Category*: feature 
- *JIRA issue*: none [MIC-XYZ](https://jira.ihme.washington.edu/browse/MIC-XYZ)

 
This makes it easier to demonstrate a minimal fertility model

### DIFF
--- a/src/vivarium/examples/disease_model/population.py
+++ b/src/vivarium/examples/disease_model/population.py
@@ -94,8 +94,9 @@ class BasePopulation:
 
         """
 
-        age_start = self.config.population.age_start
-        age_end = self.config.population.age_end
+        age_start = pop_data.user_data.get("age_start", self.config.population.age_start)
+        age_end = pop_data.user_data.get("age_end", self.config.population.age_end)
+
         if age_start == age_end:
             age_window = pop_data.creation_window / pd.Timedelta(days=365)
         else:


src/vivarium/examples/disease_model/population.py
```diff
@@ -94,8 +94,9 @@ def on_initialize_simulants(self, pop_data: SimulantData):
 
         """
 
-        age_start = self.config.population.age_start
-        age_end = self.config.population.age_end
+        age_start = pop_data.user_data.get("age_start", self.config.population.age_start)
```
rmudambi, line 97: Not sure I understand why this is necessary.

rmudambi (approved):

src/vivarium/examples/disease_model/population.py
aflaxman: I had a chance to talk with @rmudambi about this briefly today, and for future reference, this will let the example base population component work with an example fertility component following the pattern in `vivarium_public_health`. See https://github.com/ihmeuw/vivarium_public_health/blob/v0.10.22/src/vivarium_public_health/population/add_new_birth_cohorts.py#L60-L67 for an example of how this approach changes `age_start` and `age_end`, which is effective only with this new version of lines 97 and 98.

PR #260, opened 2023-01-10
Mend get_value unhashable argument for Results Manger add_observation()

mattkappel:
### Description
- *Category*: bugfix
- *JIRA issue*: [MIC-3761](https://jira.ihme.washington.edu/browse/MIC-3761)

#### Changes
- Mends the call to `get_value` to go per item in the list of targets.
- Adds a test case to `test_interface.py` to check for this bug.
- Adds a new test that passes in arguments for the `requires_values` that trigger the error case.
- Refactors a mocked `get_value` call to be in the top-level for results framework tests.
- Mends `add_observation` call with incorrect number of arguments (effect on the actual test was negligible).

### Testing
All tests pass.


src/vivarium/framework/results/manager.py
```diff
@@ -198,7 +198,7 @@ def _add_resources(self, target: List[str], target_type: SourceType):
         if target_type == SourceType.COLUMN:
             self._required_columns.update(target)
         elif target_type == SourceType.VALUE:
-            self._required_values.update(self.get_value(target))
+            self._required_values.update([self.get_value(t) for t in target])
```
rmudambi, line 201: The default value is None which is not iterable, so we will get an error

tests/framework/results/mocks.py
```diff
@@ -76,3 +76,14 @@ def verify_stratification_added(
             matching_stratification_found = True
             break
     return matching_stratification_found
+
+
+# Mock for get_value call for Pipelines, returns a str instead of a Pipeline
+def mock_get_value(self, name: str):
+    try:
+        name.__hash__()
```
rmudambi, line 84: Is it better to try to hash it here or just to check if it's a string?

src/vivarium/framework/results/manager.py
mattkappel: I don't understand what you're saying.

tests/framework/results/mocks.py
mattkappel: The error I'm trying to handle is the TypeError unhashable, but I might as well add another check if it's a string and raise TypeError if not.

src/vivarium/framework/results/manager.py
mattkappel: Understood.

tests/framework/results/mocks.py
```diff
@@ -80,10 +80,8 @@ def verify_stratification_added(
 
 # Mock for get_value call for Pipelines, returns a str instead of a Pipeline
 def mock_get_value(self, name: str):
-    try:
-        name.__hash__()
-    except TypeError:
+    if not isinstance(name, str):
         raise TypeError(
-            "Passed an unhashable type to mock get_value(), check your pipelines."
+            "Passed an non-string type to mock get_value(), check your pipelines."
```
rmudambi, line 85: Nit: should be "a" not "an"

rmudambi (approved):

hussain-jafari (approved):

rmudambi (approved):

PR #264, opened 2023-01-18
Remove copy_with_addtional_key

collijk:
Remove copy_with_addtional_key

### Description

*Category*: Refactor
*JIRA issue*: 

Removes the `copy_with_addtional_key` method from the `RandomnessStream`
as it has been supplanted by:
- Getting a randomness stream for every decision point in a model.
- Or, using the additional_key keyword when using a randomness stream method.

As the `Zen of Python` reminds us: "There should be one-- and preferably only one --obvious way to do it."

Also lets us completely decouple the RandomnessManager from the stream, which is
just good design.

### Testing
CI, no additional tests required.

### DIFF
--- a/src/vivarium/framework/randomness/manager.py
+++ b/src/vivarium/framework/randomness/manager.py
@@ -122,7 +122,6 @@ class RandomnessManager:
             clock=self._clock,
             seed=self._seed,
             index_map=self._key_mapping,
-            manager=self,
             for_initialization=for_initialization,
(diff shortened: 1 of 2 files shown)


rmudambi (approved):
No tests to delete?

hussain-jafari (approved):

collijk:
> No tests to delete?

Coverage on the randomness module is not great.

PR #268, opened 2023-01-20
Add Randomness integration tests

collijk:
Add Randomness integration tests

### Description
*Category*: test
*JIRA issue*

This adds a set of integration tests meant to examine the guarantees the 
CRN system provides (see https://vivarium.readthedocs.io/en/latest/concepts/crn.html 
for a refresher if needed).  I ended up here because I thought I'd found a bug
central to the implementation, but it turns out I instead did something clever and opaque
that makes the system work in most of our use cases. 

I want to fix this so that the actual implementation of CRN is more robust and also 
clearer to understand.  It's currently implemented across a few classes and functions
and should be unified.  Before making changes like that, it seemed prudent to back up
the randomness system with some integration tests at the very least.  In doing so 
I turned up a strict limitation of the system as well as an actual bug that needs to be 
fixed.  In this test suite, they're both marked as `pytest.mark.xfail`, i.e. as expected failures.

**The Limitation**: Our CRN system only works when the `entrance_time` of the simulant
is used as an identifying characteristic. Two examples of where this might matter:

- Example 1: Say we have a model of an intervention on maternal health that effects the gestational
  age of their children (ie, the length of their pregnancy). In this situation, it would be reasonable
  to use something like the conception time and the id of the mother as a way to identify a child
  across multiple simulations.  
- Example 2: Say we have a model of a border policy that affects the speed of immigration. In this
  situation it would be reasonable to say everyone gets the same speedup to their process and 
  we uniquely identify them by the order in which they immigrate (ie, the 10th immigrant is the same
  in both simulations and just enters the simulation on a different date).

The primary reason there's a limitation here is that the current clock time is part of the seed to the
random states used by randomness streams.  So if a person enters on a different date, any randomly
determined CRN attributes will be given a different value.  If the CRN attributes are deterministic, then 
all random numbers drawn for that person will match *when those numbers are drawn for the same 
decision and clock time*.  In particular, though, as the simulant has entered at a different time point 
than they would have without the intervention, any attributes they are initialized with by a draw from 
a random distribution may not match.  This is especially problematic for relatively static attributes (e.g. sex)
which form the basis for conditional distribution draws throughout the rest of the simulation.

There is not an easy solution to this problem, at least that I can think of.  Tagging @aflaxman here
because it's a methodological limitation that we'll have to compensate for in other ways if we decide
not to spend time working on this.

**The Bug**: The randomness stream indexes into the pandas Series underlying the `IndexMap` using
a raw `getitem` call (i.e. series[index]) instead of one of the pandas indexers (`.loc` or `.iloc`). As a reminder,
the pandas Series that underlies the `IndexMap` is a mapping between the CRN attributes in the index and
a fixed position in a randomness stream in the values. This raw `getitem` call works sort of magically due 
to some fallback logic in the pandas Index class *if and only if the first level of the index is not a float or int*. 
We tend to always put the `entrance_time` as the first CRN attribute, so we've never stumbled across it.

### Testing
All tests


tests/framework/randomness/test_crn.py
```diff
@@ -0,0 +1,301 @@
...
+    else:
+        configuration = {}
+
+    sim = InteractiveContext(
+        components=[BrokenPopulation(with_crn=with_crn, sims_to_add=sims_to_add)],
+        configuration=configuration,
+    )
+
+    for i in range(5):
+        sim.step()
```
mattkappel, line 301: How does this test pass? I was expecting an `assert` or `pytest.raises`

collijk: This test fails when the `with_crn` flag is true. But it's marked as `pytest.mark.xfail`, i.e., an expected failure.  `pytest.raises` is the right tool when you are testing error paths in your code, e.g., malformed inputs to a function.  `pytest.mark.xfail` is useful to highlight a bug in the code without an existing patch.  It's essentially a regression test I'm putting in ahead of time.

rmudambi:
Can the dependency on entrance time not be resolved by having another parameter passed to `_key` that is True if this is a call to initialize simulants, in which case the clock isn't used to create the key. If "entrance_time" is a key column for randomness, that will be brought to bear by applying the index map. If it's not, then we don't care about using "entrance_time" as a distinguishing characteristic for differentiating individuals, so it seems fine that it's not accounted for in the key.

tests/framework/randomness/test_crn.py
```diff
@@ -0,0 +1,301 @@
...
+        "clock": lambda: pd.Timestamp("2020-01-01"),
+        "seed": "abc",
+        "index_map": index_map,
+        "for_initialization": for_initialization,
+    }
+
+    stream_base = RandomnessStream(**stream_args)
+    draw_base = stream_base.get_draw(test_idx)
+
+    for arg_permutation in [
```
rmudambi, line 37: Thoughts on this test parameterization being internal to the test rather than in the `@pytest.mark.parameterize` annotation?

tests/framework/randomness/test_crn.py
```diff
@@ -0,0 +1,301 @@
...
+)
+def test_multi_sim_reproducibility_with_different_pop_growth(with_crn, pop_class):
+    if with_crn:
+        configuration = {"randomness": {"key_columns": ["crn_attr1", "crn_attr2"]}}
+    else:
+        configuration = {}
+
+    short, long = 1, 3
+    sim1 = InteractiveContext(
+        components=[pop_class(with_crn=with_crn, sims_to_add=cycle([short, short]))],
```
rmudambi, line 228: Why does this need to to have `cycle([short, short])`. Isn't that no different from `cycle([short])`?

tests/framework/randomness/test_crn.py
```diff
@@ -0,0 +1,301 @@
...
+    pop2 = sim2.get_population().set_index(["crn_attr1", "crn_attr2"])
+    assert_frame_equal(pop1, pop2)
+
+
+@pytest.mark.parametrize(
+    "pop_class, with_crn",
+    [
+        pytest.param(EntranceTimePopulation, True),
+        pytest.param(EntranceTimePopulation, False),
+        pytest.param(SequentialPopulation, True, marks=pytest.mark.xfail),
```
rmudambi, line 216: This is just marking that the test is expected to fail. Is there a way to mark that we expect it to fail on a specific assert statement? By this is mean, we expect the first `assert_frame_equal` call to pass and the later call to fail.

tests/framework/randomness/test_crn.py
```diff
@@ -0,0 +1,301 @@
...
+    if with_crn:
+        overlap = pop1.index.intersection(pop2.index)
+        assert len(overlap) > initial_pop_size
+        assert_frame_equal(pop1.loc[overlap], pop2.loc[overlap])
+    else:
+        overlap = pop1.index[:initial_pop_size]
+        assert_frame_equal(pop1.loc[overlap], pop2.loc[overlap])
+
+
+class BrokenPopulation(BasePopulation):
```
rmudambi, line 258: If you're proposing to add a bug-fix to support this use-case we probably should have a better name for this class.

rmudambi (approved):

tests/framework/randomness/test_crn.py
collijk: This is a mistake.  One should be `cycle([short, long])` and the other should be `cycle([long, short])`.  I committed the wrong thing

tests/framework/randomness/test_crn.py
collijk: I'll roll it into the parameterization of the other population classes in this module with the fix.

tests/framework/randomness/test_crn.py
collijk: Yeah, I can write a more granular test.  I'll think about it.

tests/framework/randomness/test_crn.py
collijk: :shrug:

collijk:
> Can the dependency on entrance time not be resolved by having another parameter passed to `_key` that is True if this is a call to initialize simulants, in which case the clock isn't used to create the key. If "entrance_time" is a key column for randomness, that will be brought to bear by applying the index map. If it's not, then we don't care about using "entrance_time" as a distinguishing characteristic for differentiating individuals, so it seems fine that it's not accounted for in the key.

If we ditch the clock then we need something else.  Otherwise every time we add simulants they'll get the same random numbers.  Without the clock, you'd need a deterministic algorithm to assign the properties that seed CRN for the individuals.  It's a thorny problem.

rmudambi:
> > Can the dependency on entrance time not be resolved by having another parameter passed to `_key` that is True if this is a call to initialize simulants, in which case the clock isn't used to create the key. If "entrance_time" is a key column for randomness, that will be brought to bear by applying the index map. If it's not, then we don't care about using "entrance_time" as a distinguishing characteristic for differentiating individuals, so it seems fine that it's not accounted for in the key.
> 
> If we ditch the clock then we need something else. Otherwise every time we add simulants they'll get the same random numbers. Without the clock, you'd need a deterministic algorithm to assign the properties that seed CRN for the individuals. It's a thorny problem.

I don't mean to ditch the clock. Merely not to use it when initializing simulants when you call `_key`

collijk:
> > > Can the dependency on entrance time not be resolved by having another parameter passed to `_key` that is True if this is a call to initialize simulants, in which case the clock isn't used to create the key. If "entrance_time" is a key column for randomness, that will be brought to bear by applying the index map. If it's not, then we don't care about using "entrance_time" as a distinguishing characteristic for differentiating individuals, so it seems fine that it's not accounted for in the key.
> > 
> > 
> > If we ditch the clock then we need something else. Otherwise every time we add simulants they'll get the same random numbers. Without the clock, you'd need a deterministic algorithm to assign the properties that seed CRN for the individuals. It's a thorny problem.
> 
> I don't mean to ditch the clock. Merely not to use it when initializing simulants when you call `_key`

Without the clock, the key is a combination of the decision point and any additional key information passed by the user.  Let's assume none.  That means the same random numbers are drawn for every call to that particular randomness stream.  So any model with fertility or another way to have people enter the sim is going to reuse the same random numbers.

rmudambi:
> > > > Can the dependency on entrance time not be resolved by having another parameter passed to `_key` that is True if this is a call to initialize simulants, in which case the clock isn't used to create the key. If "entrance_time" is a key column for randomness, that will be brought to bear by applying the index map. If it's not, then we don't care about using "entrance_time" as a distinguishing characteristic for differentiating individuals, so it seems fine that it's not accounted for in the key.
> > > 
> > > 
> > > If we ditch the clock then we need something else. Otherwise every time we add simulants they'll get the same random numbers. Without the clock, you'd need a deterministic algorithm to assign the properties that seed CRN for the individuals. It's a thorny problem.
> > 
> > 
> > I don't mean to ditch the clock. Merely not to use it when initializing simulants when you call `_key`
> 
> Without the clock, the key is a combination of the decision point and any additional key information passed by the user. Let's assume none. That means the same random numbers are drawn for every call to that particular randomness stream. So any model with fertility or another way to have people enter the sim is going to reuse the same random numbers.

If entrance time is still a key column, even if we're still using the "same random numbers" at each time-step, we'll get different random numbers for individual simulants entering at different timesteps even with clock removed from the `_key` hash, since the different entrance time means they'd be treated as a different simulant.

If entrance time is not still a key column, I suggest it's up to the user of the system to make sure they define key columns in such a way as to make sure that individuals are uniquely mappable, whether it's by entrance order, parent and conception date, or whatever. I'm not sure it's vivarium's job to make sure the users use the framework correctly as long as we make it clear to the users that they need to care about this.

mattkappel (approved):

mattkappel (approved):

PR #269, opened 2023-01-24
Fix bug in gather_results, fix typing notation

mattkappel:
### Description
- *Category*: bugfix
- *JIRA issue*: [MIC-3684](https://jira.ihme.washington.edu/browse/MIC-3684)

Mends two bugs found in the development of the `DiseaseObserver` where the population was being improperly refined during `gather_results` and the `ResultsContext` default stratifications were improperly set when used by `ResultsStratifier`. Also updates return type annotation that was incomplete for the Stratification default mapper.

### Testing
`DiseaseObserver` works with this code.

### DIFF
--- a/src/vivarium/framework/results/context.py
+++ b/src/vivarium/framework/results/context.py
@@ -105,10 +105,11 @@ class ResultsContext:
         ].items():
             # Results production can be simplified to
             # filter -> groupby -> aggregate in all situations we've seen.
-            stratified_pop = population.copy()
             if pop_filter:
-                stratified_pop = stratified_pop.query(pop_filter)
-            pop_groups = stratified_pop.groupby(list(stratifications))
+                filtered_pop = population.query(pop_filter)
+            else:
+                filtered_pop = population
+            pop_groups = filtered_pop.groupby(list(stratifications))
             for measure, aggregator_sources, aggregator, additional_keys in observations:
                 if aggregator_sources:
                     aggregates = pop_groups[aggregator_sources].apply(aggregator).fillna(0.0)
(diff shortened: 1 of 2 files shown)


src/vivarium/framework/results/context.py
```diff
@@ -105,9 +105,10 @@ def gather_results(self, population: pd.DataFrame, event_name: str) -> Dict[str,
         ].items():
             # Results production can be simplified to
             # filter -> groupby -> aggregate in all situations we've seen.
+            stratified_pop = population.copy()
             if pop_filter:
```
rmudambi, line 109: Wouldn't something like this be a bit cleaner:
```
if pop_filter:
    filtered_pop = population.query(pop_filter)
else:
    filtered_pop = population
```

That way we don't have to copy the pop table and in the no filter case we don't use any extra memory.

src/vivarium/framework/results/stratification.py
```diff
@@ -66,5 +66,5 @@ def __call__(self, population: pd.DataFrame) -> pd.DataFrame:
         return population
 
     @staticmethod
-    def _default_mapper(pop: pd.DataFrame) -> str:
+    def _default_mapper(pop: pd.DataFrame) -> Union[pd.Series, str]:
```
rmudambi, line 69: This can't ever be `str`. The `axis=1` argument ensures we always return a Series (or a DataFrame if `pop` has more than 1 column). 

I've forgotten if this is enforced, but it's not permissible to use the default mapper if the pop passed in has more than one column. Mappers always have to return a Series.

rmudambi (changes requested):

src/vivarium/framework/results/context.py
mattkappel: That's a good point. That copy could be expensive and is ultimately unnecessary.

rmudambi (approved):

hussain-jafari (approved):

albrja (approved):

rmudambi (approved):

PR #275, opened 2023-02-21
Remove filter_for_probability and reduce argument chaining in choice

collijk:
### Description

*Category*: refactor
*JIRA issue*: 

- Fix typo `sub-population` -> `subpopulation` in randomness docs
- Use `RandomnessStream.get_draw` for all random numbers.  This is done by:
  - folding `filter_for_probability` function into the randomness stream method of the same name.
  - cleaning up the argument chaining in the `choice` calls
- Make `choice` function private

### Testing
CI

### DIFF
--- a/src/vivarium/framework/randomness/stream.py
+++ b/src/vivarium/framework/randomness/stream.py
@@ -152,7 +152,7 @@ class RandomnessStream:
         """Decide an event outcome for each individual from rates.
 
         Given a population or its index and an array of associated rates for
-        some event to happen, we create and return the sub-population for whom
+        some event to happen, we create and return the subpopulation for whom
         the event occurred.
 
         Parameters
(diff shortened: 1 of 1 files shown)


src/vivarium/framework/randomness/stream.py
```diff
@@ -316,12 +321,10 @@ def random(
     return pd.Series(index=index, dtype=float)  # Structured null value
 
 
-def choice(
-    key: str,
-    index: Union[pd.Index, pd.MultiIndex],
+def _choice(
```
rmudambi, line 324: Why does this method even need to exist outside the randomness stream if we're making it private?

rmudambi (approved):

src/vivarium/framework/randomness/stream.py
mattkappel: +1

mattkappel (approved):

stevebachmeier (approved):

src/vivarium/framework/randomness/stream.py
collijk: It does not.  It's a more complicated change and will be addressed later.

PR #277, opened 2023-02-22
Fix bug for non-unique initial IndexMap.update

collijk:
### Description

*Category*: bugfix
*JIRA issue*: 

- Added regression testing for bug where IndexMap.update logic would not treat duplicate initial mapping data as duplicates.
- Fixed bug.
- Removed `"entrance_time"` as a default key_column for the randomness system as it's not generated by anything in base `vivarium`
- Added `"entrance_time"` and `"age"` to the `base_config` test fixture so the `TestPopulation` component will work.

### Testing
New tests plus normal CI


tests/framework/randomness/test_index_map.py
```diff
@@ -164,31 +164,49 @@ def test_hash_uniformity(map_size_and_hashed_values):
...
+
+    index_map.update(keys)
+    with pytest.raises(RandomnessError):
+        index_map.update(keys)
+
+
+def test_update_empty_good_keys(index_map):
+    keys = generate_keys(1000)
+    index_map.update(keys)
+    assert len(index_map._map) == len(keys), "All keys not in mapping"
```
rmudambi, line 198: Should have different error messages for these two error cases. 

Minor point, but I think any two of these asserts is sufficient to test all keys are present and are unique.

tests/framework/randomness/test_index_map.py
```diff
@@ -164,31 +164,49 @@ def test_hash_uniformity(map_size_and_hashed_values):
...
+
+
+def test_update_nonempty_good_keys(index_map):
+    keys = generate_keys(2000)
+    keys1, keys2 = keys[:1000], keys[1000:]
+
+    index_map.update(keys1)
+    index_map.update(keys2)
+
+    assert len(index_map._map) == len(keys), "All keys not in mapping"
```
rmudambi, line 210: Same as above.

rmudambi (approved):

src/vivarium/framework/randomness/index_map.py
```diff
@@ -39,8 +39,10 @@ def update(self, new_keys: pd.Index) -> None:
         """
         if new_keys.empty or not self._use_crn:
             return  # Nothing to do
-        elif not self._map.index.intersection(new_keys).empty:
```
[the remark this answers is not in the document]
collijk: It did not handle duplicates within the new keys only values duplicated between the existing map and the new keys.

tests/framework/randomness/test_index_map.py
collijk: Fixed in next pr

PR #279, opened 2023-02-23
gather_results empty data bugfix

hussain-jafari:
### Description
- *Category*: bugfix
- *JIRA issue*: [MIC-3840](https://jira.ihme.washington.edu/browse/MIC-3840)

When we filter to an empty population in gather_results, the output of our groupby -> aggregate is an empty dataframe. Vivarium tries to convert this to a Series using squeeze and errors out if it's not a Series. However, an empty dataframe that is squeezed is still a dataframe.

### Testing
Added test with filter that would lead to an empty dataframe. Checked that previous code failed test and that updated code passed.

### DIFF
--- a/tests/framework/results/test_context.py
+++ b/tests/framework/results/test_context.py
@@ -259,7 +259,7 @@ def test__get_stratifications(
         "len_aggregator_one_stratification",
         "sum_aggregator_one_stratification",
         "custom_aggregator_one_stratification",
-        "len_aggregator_two_stratifications_empty_filter",
+        "empty_pop_filter",
     ],
 )
 def test_gather_results(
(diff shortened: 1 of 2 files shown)


tests/framework/results/test_context.py
```diff
@@ -250,6 +250,7 @@ def test__get_stratifications(
             ["house"],
             1.53277,
         ),
+        ("wizard_count", "house=='Durmstrang'", None, len, ["house", "familiar"], 4),
```
rmudambi, line 253: This test passes, but it doesn't make any sense. There aren't 4 member of Durmstrang in each stratification group. There are 0, but `gather_results()` returns an empty dict, which means that any assertion will pass. Our test should be that the returned dict is empty.

rmudambi (changes requested):

rmudambi (approved):

mattkappel (approved):

PR #280, opened 2023-02-24
Clean up CRN implementation so it's clearer what's going on

collijk:
### Description

*Category*: refactor
*JIRA issue*: 

Leverages the changes of the previous several PRs to implifies the logic of the 
CRN implementation and group all the code for random number generation
together in a succinct single method.

### Testing
CI (which includes unit and integration tests for CRN)

### DIFF
--- a/src/vivarium/framework/randomness/stream.py
+++ b/src/vivarium/framework/randomness/stream.py
@@ -132,15 +132,54 @@ class RandomnessStream:
         -------
         pandas.Series
             A series of random numbers indexed by the provided `pandas.Index`.
-        """
-        key = self._key(additional_key)
-        if self.initializes_crn_attributes:
-            draw = random(key, pd.Index(range(len(index))))
-            draw.index = index
-        else:
-            draw = random(key, index, self.index_map)
 
-        return draw
+        Note
+        ----
+        This is the core of the CRN implementation, allowing for consistent use of random
+        numbers across simulations with multiple scenarios.
+
+        See Also
+        --------
+        https://en.wikipedia.org/wiki/Variance_reduction and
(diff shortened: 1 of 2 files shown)


mattkappel (approved):

albrja (approved):

src/vivarium/framework/randomness/stream.py
```diff
@@ -132,15 +132,54 @@ def get_draw(self, index: pd.Index, additional_key: Any = None) -> pd.Series:
...
+            return pd.Series(index=index, dtype=float)
+
+        # Initialize a random_state with a seed based on the simulation clock, the
+        # decision_point this stream represents, and any additional user-supplied information.
+        # This is one pre-condition to reproducibility.
+        seed = get_hash(self._key(additional_key))
+        random_state = np.random.RandomState(seed=seed)
+
+        # Second, we need to sample a very large chunk of random numbers. The size of the
+        # index map is set at the simulation start and is at least 10x the size of the
```
rmudambi, line 159: Does it make sense for the size of the index map to be paramterized as multiplicative of of the initial population size rather than as a static value. By that I mean, say we want an index map size of 1 million and an initial population of 100 thousand. Then we set the key `randomness.map_size: 1_000_000` the key `population.population_size: 100_000`. Would it make more sense to replace `randomness.map_size` with something like `randomness.map_multiplier: 10`?

rmudambi (approved):

src/vivarium/framework/randomness/stream.py
collijk: :shrug: We've never had to change the parameter so I don't think the parameterization matters much.

Issue #283, opened 2023-02-25
Logging in interactive settings is verbose and not configurable


collijk:
Working in a notebook or REPL with vivarium simulations is pretty unpleasant.  The logging level is set to `DEBUG` by default, which throws a bunch of not very helpful information at the user.  It's also not configurable in any way.  After tackling https://github.com/ihmeuw/vivarium/issues/282, the next thing to do should be to allow user level configuration of the logging and change the default level in the interactive setting to be much less noisy.

aflaxman:
> Working in a notebook or REPL with vivarium simulations is pretty unpleasant. The logging level is set to `DEBUG` by default, which throws a bunch of not very helpful information at the user. It's also not configurable in any way. After tackling #282, the next thing to do should be to allow user level configuration of the logging and change the default level in the interactive setting to be much less noisy.

In case it is helpful, I have been using the following code snippet to suppress DEBUG messages when I work in a notebook sometimes:

```
# all the debugging logs are annoying... do this to suppress them:
import vivarium.interface.utilities
vivarium.interface.utilities.configure_logging_to_terminal(verbose=False)
```

Issue #282, opened 2023-02-25
Simulation logging can't distinguish between unique simulation contexts


collijk:
`vivarium` is structured to allow multiple simulation contexts to live in a single python process, but logging is done in interactive settings with global loggers.  This means log messages can't distinguish between simulations if they're run from the same Python process.  E.g.
```
>>> from vivarium.examples.disease_model import get_disease_model_simulation
>>> sim1 = get_disease_model_simulation()
2023-02-25 14:54:45.034 | DEBUG    | vivarium.framework.values:register_value_modifier:392 - Registering metrics.1.results_manager.get_results as modifier to metrics
2023-02-25 14:54:45.035 | DEBUG    | vivarium.framework.values:_register_value_producer:338 - Registering value pipeline mortality_rate
2023-02-25 14:54:45.035 | DEBUG    | vivarium.framework.values:_register_value_producer:338 - Registering value pipeline lower_respiratory_infections.cause_specific_mortality_rate
2023-02-25 14:54:45.035 | DEBUG    | vivarium.framework.values:register_value_modifier:392 - Registering mortality_rate.1.diseasemodel.lower_respiratory_infections.delete_cause_specific_mortality as modifier to mortality_rate
2023-02-25 14:54:45.035 | DEBUG    | vivarium.framework.values:register_value_modifier:392 - Registering metrics.2.diseasemodel.lower_respiratory_infections.metrics as modifier to metrics
2023-02-25 14:54:45.035 | DEBUG    | vivarium.framework.values:_register_value_producer:338 - Registering value pipeline susceptible_to_lower_respiratory_infections.excess_mortality_rate
2023-02-25 14:54:45.035 | DEBUG    | vivarium.framework.values:_register_value_producer:338 - Registering value pipeline susceptible_to_lower_respiratory_infections.excess_mortality_rate.population_attributable_fraction
2023-02-25 14:54:45.035 | DEBUG    | vivarium.framework.values:register_value_modifier:392 - Registering mortality_rate.2.diseasestate.susceptible_to_lower_respiratory_infections.add_in_excess_mortality as modifier to mortality_rate
2023-02-25 14:54:45.036 | DEBUG    | vivarium.framework.values:_register_value_producer:338 - Registering value pipeline infected_with_lower_respiratory_infections.incidence_rate
[...]

PR #284, opened 2023-02-25
Add unique simulation names

collijk:
### Description

*Category*: feature
*JIRA issue*: N/A

This is the first step to fixing a couple issues with simulation logging described here: https://github.com/ihmeuw/vivarium/issues/282 and here: https://github.com/ihmeuw/vivarium/issues/283.
It adds an attribute to th simulation context to track all created contexts within a python process which
allows us to uniquely name simulation contexts.  

This implementation requires that SimulationContexts
are created in serial to ensure thread safety, which I think is a reasonable constraint given all historical 
use cases (in which we never do multiprocessing or multithreading to begin with).  

### Testing
CI + new unit tests

### DIFF
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -18,16 +18,16 @@ Finally, there are a handful of wrapper methods that allow a user or user
 tools to easily setup and run a simulation.
 
 """
-import time
 from pathlib import Path
 from pprint import pformat
-from typing import Dict, List, Tuple, Union
+from typing import Dict, List, Set, Union
 
 import numpy as np
 import pandas as pd
 from loguru import logger
 
 from vivarium.config_tree import ConfigTree
+from vivarium.exceptions import VivariumError
 from vivarium.framework.configuration import build_model_specification
 
 from .artifact import ArtifactInterface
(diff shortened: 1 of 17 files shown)


tests/framework/test_engine.py
```diff
@@ -25,6 +26,12 @@ def is_same_object_method(m1, m2):
     return m1.__func__ is m2.__func__ and m1.__self__ is m2.__self__
 
 
+@pytest.fixture()
+def SimulationContext():
+    yield SimulationContext_
+    SimulationContext_._clear_context_cache()
```
mattkappel, line 32: Clever use of yield.

src/vivarium/framework/engine.py
```diff
@@ -46,13 +46,39 @@
 
 
 class SimulationContext:
+
+    _created_simulation_contexts: Set[str] = set()
+
+    @staticmethod
+    def _get_context_name(sim_name: Union[str, None]) -> str:
+        # Not threadsafe, but I don't see why we'd need to be.
```
rmudambi, line 54: You should add docstring making it clear that there is a side-effect here (i.e. adding the simulation to the set of simulation context names

tests/framework/test_engine.py
rmudambi: When does `SimulationContext_._clear_context_cache()` get triggered? I think it'd be helpful to add a comment explaining exactly what is happening here.

src/vivarium/framework/engine.py
```diff
@@ -46,13 +46,39 @@
...
+            if sim_name in SimulationContext._created_simulation_contexts:
+                msg = (
+                    "Attempting to create two SimulationContexts "
+                    f"with the same name {sim_name}"
+                )
+                raise VivariumError(msg)
+
+        if sim_name is None:
+            sim_number = len(SimulationContext._created_simulation_contexts) + 1
+            sim_name = f"simulation_{sim_number}"
```
rmudambi, line 65: What if there was a simulation manually named this? You should move this assignment above the error check.

rmudambi (approved):

src/vivarium/framework/engine.py
collijk: good catch

tests/framework/test_engine.py
collijk: Disagree.  This is the standard teardown procedure for fixtures: https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended

rmudambi (approved):

collijk:
Screwed up the commit history with an errant commit from a later PR.  See #287 for the redo

PR #286, opened 2023-02-27
Add a CLI progress bar

collijk:
### Description

*Category*: feature
*Issues*: N/A

The main thrust of this PR is to add a progress bar to the CLI and to clean up the interactive simulation progress bar implementation.

Example CLI usage:
image

Example interactive sim usage from REPL: 
image

Example interactive sim usage from notebook:
image

**Detailed changes**
- Add `tqdm` to framework dependencies
- Add a `steps_remaining` property to simulation clocks so they can tell us the remaining number of simulation steps.
- Add a private context manager to the `InteractiveContext` to abstract away the behavior of adjusting the simulation step size.
- Add a `take_steps` method to the base `SimulationContext` that is smart about the balance between verbose logging and using a progress bar.
- Make the time stepping methods on the base `SimulationContext` and the `InteractiveContext` more consistent and clean up their implementations a bit.
- Replace custom progress bar code in the `InteractiveContext` with `tqdm`'s implementation so we don't have to think about it when it breaks (as it's already broken).

### Testing
CI + manual testing

### DIFF
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -24,6 +24,7 @@ from typing import Dict, List, Set, Union
 
 import numpy as np
 import pandas as pd
+from tqdm import auto as tqdm
 
 from vivarium.config_tree import ConfigTree
 from vivarium.exceptions import VivariumError
@@ -202,23 +203,34 @@ class SimulationContext:
         )
         self._clock.step_forward()
 
-    def step(self):
-        self._logger.debug(self._clock.time)
+    def step(self, log_step: bool = True):
+        if log_step:
+            self._logger.debug(self._clock.time)
+
         for event in self.time_step_events:
             self._lifecycle.set_state(event)
             self.time_step_emitters[event](self._population.get_population(True).index)
+
         self._clock.step_forward()
 
-    def run(self):
-        while self._clock.time < self._clock.stop_time:
-            self.step()
+    def take_steps(self, number_of_steps: int, progress_bar: bool = False) -> None:
+        with tqdm.tqdm(total=number_of_steps, disable=not progress_bar) as pbar:
+            for _ in range(number_of_steps):
+                pbar.set_description(f"Time step {self._clock.time}")
+                self.step()
+                pbar.update(1)
+
+    def run(self, progress_bar: bool = False) -> int:
+        number_of_steps = self._clock.steps_remaining
+        self.take_steps(number_of_steps, progress_bar=progress_bar)
+        return number_of_steps
 
     def finalize(self):
         self._lifecycle.set_state("simulation_end")
         self.end_emitter(self._population.get_population(True).index)
         unused_config_keys = self.configuration.unused_keys()
         if unused_config_keys:
-            self._logger.warn(
+            self._logger.warning(
                 f"Some configuration keys not used during run: {unused_config_keys}."
             )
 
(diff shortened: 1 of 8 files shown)


stevebachmeier (approved):
nice

src/vivarium/framework/engine.py
```diff
@@ -202,16 +203,28 @@ def initialize_simulants(self):
...
             self.time_step_emitters[event](self._population.get_population(True).index)
+
         self._clock.step_forward()
 
-    def run(self):
-        while self._clock.time < self._clock.stop_time:
-            self.step()
+    def take_steps(self, number_of_steps: int, progress_bar: bool = False) -> None:
+        log_step = not progress_bar
+        with tqdm.tqdm(total=number_of_steps, disable=not progress_bar) as pbar:
```
rmudambi, line 242: Is it not possible for this logic to live inside the logging system? It feels more natural there.

src/vivarium/interface/interactive.py
```diff
@@ -234,6 +259,18 @@ def get_component(self, name: str) -> Any:
         """
         return self._component_manager.get_component(name)
 
+    @contextmanager
+    def _adjust_step_size(self, step_size: Timedelta = None):
+        old_step_size = self._clock.step_size
```
rmudambi, line 264: Can you add a docstring?

src/vivarium/interface/interactive.py
```diff
@@ -118,39 +131,51 @@ def run_until(self, end_time: Time, with_logging: bool = True) -> int:
...
-
-    def take_steps(
-        self, number_of_steps: int = 1, step_size: Timedelta = None, with_logging: bool = True
-    ):
-        """Run the simulation for the given number of steps.
+            msg = f"Provided time must be an instance of {type(self._clock.time)}"
+            raise TypeError(msg)
+
+        step_size = step_size if step_size is not None else self._clock.step_size
+        number_of_steps = int(ceil((end_time - self._clock.time) / self._clock.step_size))
```
rmudambi, line 138: Would it make sense to make `clock.steps_remaining` a function that takes an optional parameter of end time? This is the same logic that you are using there.

src/vivarium/framework/engine.py
collijk: I think it is very not advisable.  The logger is a very thin wrapper around normal logging conventions.  Making good and informative progress bars requires a lot of context-specific state information that I think would be hard to generalize reasonably into the logging system.

On a more theoretical point, I think of progress bars as kind of an animation trick for your terminal that obscures the guarantees of a logging stream (a stream of discrete messages), which make me a bit hesitant to conflate the two.

src/vivarium/interface/cli.py
```diff
@@ -93,6 +93,12 @@ def simulate():
     is_flag=True,
     help="Suppresses all logging except for warnings and errors.",
 )
+@click.option(
+    "--progress-bar",
+    "--pb",
```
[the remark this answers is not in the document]
collijk: We used both in all the covid libraries without much issue.

src/vivarium/interface/interactive.py
collijk: I don't think so.  I think of the clock as a closed system with a current time and step size and some functions for advancing time, and not really a calculator for time-based inquiries.  But I'm not adamant about it.

rmudambi: As a compromise, you could have a utility function in the time module that is called in both locations?

rmudambi (approved):

PR #292, opened 2023-05-16
pin pandas>=2; stop supporting python 3.7; fix failing doctest

stevebachmeier:
### Description
- *Category*: bugfix
- *JIRA issue*: [MIC-3948](https://jira.ihme.washington.edu/browse/MIC-3948)

Doctests started failing with Pandas 2.0 and we temporarily pinned it at <2.0.
This goes all in with Pandas v2.0+.
- Pins Pandas>=2.0
- Removes support of python 3.7 (Pandas 2.0 requires python 3.8+)
- Updates readme 
- Fixes failing doctest

### Testing
CI is passing again

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -19,7 +19,7 @@ Vivarium
 Vivarium is a simulation framework written using standard scientific Python
 tools.
 
-**Vivarium requires Python 3.8-3.10 to run**
+**Vivarium requires Python 3.8-3.11 to run**
 
 You can install ``vivarium`` from PyPI with pip:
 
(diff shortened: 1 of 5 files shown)


albrja (approved):

rmudambi (commented):
Can you see if we can support 3.11?

README.rst
ramittal, line 1: I am guessing, we generate the package version from the git-tag, would suggest to change major version number to signal this is a major change and does not break existing vivarium users.

stevebachmeier: Yes, this is going into develop - we'll release in a separate ticket (and I agree it should be a major version change)

hussain-jafari (approved):

README.rst
rmudambi: We don't need to get into this here, but I don't agree that this should be a major version change.

rmudambi (approved):

PR #296, opened 2023-05-19
Allow zero stratifications

hussain-jafari:
### Description
- *Category*: feature
- *JIRA issue*: [MIC-3774](https://jira.ihme.washington.edu/browse/MIC-3774)

Allow zero stratifications when running gather_results. Add test.

### Testing
Ran CIFF SAM for a few time steps and checked that the categorical risk observer was recording child stunting person time properly without any stratifications. All tests pass and new test was written.

### DIFF
--- a/src/vivarium/framework/results/context.py
+++ b/src/vivarium/framework/results/context.py
@@ -45,11 +45,6 @@ class ResultsContext:
                 "Multiple calls are being made to set default grouping columns "
                 "for results production."
             )
-        if not default_grouping_columns:
-            raise ResultsConfigurationError(
-                "Attempting to set an empty list as the default grouping columns "
-                "for results production."
-            )
         self.default_stratifications = default_grouping_columns
 
     def add_stratification(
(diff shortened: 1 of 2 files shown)


src/vivarium/framework/results/context.py
```diff
@@ -162,38 +166,43 @@ def _get_stratifications(
...
-                ]
-                # Sorts additional_keys by the field name.
-                + [
-                    _format(field, category)
-                    for field, category in sorted(additional_keys.items())
-                ]
-            )
-            results[key] = val
-        return results
+            # First we expand the categorical index over unobserved pairs.
```
mattkappel, line 177: nit: Could we do away with this `else` block and it's indentation? If the above conditional is true, we'll not enter the rest of the method.

mattkappel (approved):
Approved with a suggestion to avoid unnecessary indentation.

rmudambi (approved):

PR #297, opened 2023-05-22
move setting of default stratifications

hussain-jafari:
### Description
- *Category*: refactor
- *JIRA issue*: [MIC-3890](https://jira.ihme.washington.edu/browse/MIC-3890)

We'd like to set default stratifications in the setup of the results manager than than the results stratifier. 

### Testing
Ran CIFF SAM for a few time steps, tests are passing.

### DIFF
--- a/src/vivarium/framework/results/manager.py
+++ b/src/vivarium/framework/results/manager.py
@@ -29,6 +29,12 @@ class ResultsManager:
     `collect_metrics`).
     """
 
+    configuration_defaults = {
+        "stratification": {
+            "default": [],
+        }
+    }
+
     def __init__(self):
         self._metrics = Counter()
         self._results_context = ResultsContext()
@@ -58,6 +64,8 @@ class ResultsManager:
         builder.event.register_listener("collect_metrics", self.on_collect_metrics)
 
         self.get_value = builder.value.get_value
+
+        self.set_default_stratifications(builder.configuration.stratification.default)
 
         builder.value.register_value_modifier("metrics", self.get_results)
 
--- a/src/vivarium/framework/results/manager.py
+++ b/src/vivarium/framework/results/manager.py
@@ -65,7 +65,7 @@ class ResultsManager:
 
         self.get_value = builder.value.get_value
 
-        self.set_default_stratifications(builder.configuration.stratification.default)
+        self.set_default_stratifications(builder)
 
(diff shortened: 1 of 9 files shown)


mattkappel (approved):

src/vivarium/framework/results/manager.py
```diff
@@ -59,6 +65,8 @@ def setup(self, builder: "Builder"):
 
         self.get_value = builder.value.get_value
 
+        self.set_default_stratifications(builder.configuration.stratification.default)
```
rmudambi, line 68: It's more in line with our conventions to define this method like `self.set_default_stratifications(builder)` and parse the configuration inside the method. This also has the benefit of moving all the logic of setting the default stratification inside the method defined to do that rather than having some of it outside in the `setup` function.

tests/framework/results/test_manager.py
```diff
@@ -224,3 +224,17 @@ def test_add_observation_nop_stratifications(
     )
     for m in match:
         assert m in caplog.text
+
+
+def test_setting_default_stratifications(mocker):
```
rmudambi, line 229: This is fine. Do we already have a test that tests that `set_default_stratifications` works as expected?

tests/framework/results/test_stratification.py
```diff
@@ -140,3 +141,18 @@ def test_stratification_call_raises(
     my_stratification = Stratification(name, sources, categories, mapper, is_vectorized)
     with pytest.raises(expected_exception):
         raise my_stratification(STUDENT_TABLE)
+
+
+def test_results_manager_setup(mocker):
```
rmudambi, line 146: Isn't this test identical to the one above in `test_manager`?

rmudambi (changes requested):

src/vivarium/framework/results/manager.py
hussain-jafari: Will update

tests/framework/results/test_manager.py
hussain-jafari: Not directly - I'll add one.

tests/framework/results/test_stratification.py
hussain-jafari: Ah yeah I forgot to clean this one up.

rmudambi (approved):
Looks good!

PR #308, opened 2023-07-26
include missing results

hussain-jafari:
### Description
- *Category*: bugfix
- *JIRA issue*: [MIC-4230](https://jira.ihme.washington.edu/browse/MIC-4230)

Include all keys in _metrics Counter attribute of results manager on post setup. 

### Testing
Ran simulation without bugfix and verified that run with bugfix included results that were missing in the old run.

### DIFF
--- a/src/vivarium/framework/results/manager.py
+++ b/src/vivarium/framework/results/manager.py
@@ -1,8 +1,8 @@
+import itertools
 from collections import Counter
 from enum import Enum
 from typing import TYPE_CHECKING, Callable, List, Union
 
-import itertools
 import pandas as pd
 
 from vivarium.framework.event import Event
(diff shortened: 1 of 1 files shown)


src/vivarium/framework/results/manager.py
```diff
@@ -81,6 +83,31 @@ def on_time_step_cleanup(self, event: Event):
     def on_collect_metrics(self, event: Event):
         self.gather_results("collect_metrics", event)
 
+    def on_post_setup(self, event: Event):
```
stevebachmeier, line 86: I'd move this to before `def on_time_step_prepare()` (also move the post_setup listener registration to before the the "time_step__prepare" registration.

stevebachmeier (approved):

src/vivarium/framework/results/manager.py
```diff
@@ -69,6 +71,33 @@ def setup(self, builder: "Builder"):
...
+                key = measure_str + "_".join(complete_stratifications)
+                self._metrics[key] = 0
+
+        for event in self._results_context.observations:
+            for (_, stratifications), observations in self._results_context.observations[
+                event
+            ].items():
+                for measure, *_ in observations:
+                    create_measure_specific_keys(measure, stratifications)
+
```
mattkappel, line 100: Why did you choose this approach, initializing everything effectively to zero, rather than simply recording 0 when the filtered pop is 0 (i.e., [change this code](https://github.com/ihmeuw/vivarium/blob/9bde61308d205a9f71a16c3d7ac9cd8b9454201b/src/vivarium/framework/results/context.py#L120-L121))? This seems like potentially a waste of computation at the start of the sim.

rmudambi: There are two reasons for this:
- We only have to do this once during setup, rather than during every observation
- It is also possible that the else branch will return some but not all keys, which means we'd need to check there as well

src/vivarium/framework/results/manager.py
```diff
@@ -69,6 +71,33 @@ def setup(self, builder: "Builder"):
...
+    def on_post_setup(self, event: Event):
+        # update self._metrics to have all output keys
+        def create_measure_specific_keys(measure: str, stratifications: List[str]) -> None:
+            measure_str = f"MEASURE_{measure}_"
+            individual_stratification_strings = [
+                [
+                    f"{stratification.name.upper()}_{category}"
+                    for category in stratification.categories
+                ]
+                for stratification in sorted(
```
rmudambi, line 83: Is there a reason this has to be sorted?

rmudambi (approved):

src/vivarium/framework/results/manager.py
hussain-jafari: The stratifications are sorted by the results context so this makes sure that the output keys in post setup and the output keys generated by gather_results match.

PR #311, opened 2023-08-09
create Component

rmudambi:
### Description

- *Category*: refactor/POC
- *JIRA issue*: [MIC-4329](https://jira.ihme.washington.edu/browse/MIC-4329)

Creates a standard base class called `Component` that all components used in vivarium simulations should inherit from.

Features include:
- Ensuring components define required attributes/properties: `__repr__` and `name`
- Defining a default implementation of lifecycle methods: `__init__`, `setup`, and `on_post_setup`
- Defining existing attributes/properties that are not required but are currently being used by the vivarium framework: `configuration_defaults`, `sub_components`
- Defining new properties to standardize setting column requirements: `columns_created`, `columns_required`, `initialization_columns_required`
- Defining new properties to standardize setting priority requirements: `X_priority`
- Defining methods for registering listeners and creating population views: `set_population_view`, `register_simulant_initializer`, and `register_X_listener`
- Defining placeholder methods for the listeners to register (e.g. `on_initialize_simulants`, `on_time_step`)

### Testing
Tests pass locally and I can run a simulation with `BasePopulation` inheriting from `Component`

### DIFF
--- a/src/vivarium/component.py
+++ b/src/vivarium/component.py
@@ -8,7 +8,7 @@ if TYPE_CHECKING:
     from vivarium.framework.population import PopulationView, SimulantData
 
 
-class VivariumComponent(ABC):
+class Component(ABC):
     """A component that can be used in a Vivarium simulation."""
 
     """
@@ -32,7 +32,7 @@ class VivariumComponent(ABC):
         pass
 
     @property
-    def sub_components(self) -> List["VivariumComponent"]:
+    def sub_components(self) -> List["Component"]:
         """A list of the components that are managed by this component."""
         return self._sub_components
 
@@ -99,7 +99,7 @@ class VivariumComponent(ABC):
     #####################
 
     def __init__(self):
-        self._sub_components: List["VivariumComponent"] = []
+        self._sub_components: List["Component"] = []
         self.population_view: Optional[PopulationView] = None
 
     def setup(self, builder: "Builder") -> None:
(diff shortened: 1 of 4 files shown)


src/vivarium/component.py
```diff
@@ -0,0 +1,143 @@
...
+        return []
+
+    @property
+    def initialization_columns_required(self) -> List[str]:
+        """A list of the columns this component requires during simulant initialization."""
+        return []
+
+    @property
+    def time_step_prepare_priority(self) -> int:
+        return 5
```
hussain-jafari, line 52: I just looked this up and in case anyone reading doesn't know - the event manager in vivarium defaults to priority 5 when registering listeners if not specified :)

src/vivarium/component.py
```diff
@@ -0,0 +1,143 @@
...
+    #################
+    # Setup methods #
+    #################
+
+    def get_population_view(self, builder: "Builder") -> Optional["PopulationView"]:
+        population_view_columns = self.columns_created + self.columns_required
+        if population_view_columns:
+            return builder.population.get_view(population_view_columns)
+        else:
+            return None
```
stevebachmeier, line 89: Why does this return None instead of raising?

stevebachmeier (approved):

src/vivarium/component.py
rmudambi: A component is not required to have a population view. 

Now that I think more carefully about this, I think this may have issues in the situation where we want a component to have all columns of the population view, since that is currently done by passing an empty list to the population view. Perhaps the default ought to be `None` instead of `[]`.

albrja (approved):

PR #313, opened 2023-08-22
Make ConfigTree objects pickleable

zmbc:
Currently, this fails:

```python
import pickle

from vivarium.config_tree import ConfigTree

tree = ConfigTree({'foo': 'bar', 'baz': {'quux': 'foo'}})

print(pickle.loads(pickle.dumps(tree)))
```

with `vivarium.config_tree.ConfigurationKeyError: 'No value at name __getstate__.'`

The root of this problem is that the `__getattr__`/`__setattr__` stuff that ConfigTree does to support dot-syntax (e.g. tree.foo.bar) is also applying to dunder attributes. pickle automatically checks for a `__getstate__` attribute, and only catches AttributeError, so the ConfigurationKeyError is unhandled.

This PR fixes this issue by defining `__getstate__` and `__setstate__`.

### DIFF
--- a/src/vivarium/config_tree.py
+++ b/src/vivarium/config_tree.py
@@ -589,3 +589,6 @@ class ConfigTree:
                 for name, c in self._children.items()
             ]
         )
+
+    def __eq__(self, other):
+        raise NotImplementedError
--- a/src/vivarium/config_tree.py
+++ b/src/vivarium/config_tree.py
@@ -513,11 +513,16 @@ class ConfigTree:
 
     def __setattr__(self, name, value):
         """Set a value on the outermost layer."""
-        # https://stackoverflow.com/a/50888571/
-        if name.startswith("__") and name.endswith("__"):
-            raise AttributeError
-
         if name not in self:
+            if name.startswith("__") and name.endswith("__"):
+                # Note that we allow keys that look like dunder attributes,
+                # they just raise a different error when you try to access ones
+                # that don't exist (as attributes).
+                # This is important because it is expected by some modules,
+                # in particular the pickle module.
+                # https://stackoverflow.com/a/50888571/
+                raise AttributeError
+
             raise ConfigurationKeyError(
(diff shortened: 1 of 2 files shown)


src/vivarium/config_tree.py
```diff
@@ -513,6 +513,10 @@ def _set_with_metadata(
 
     def __setattr__(self, name, value):
         """Set a value on the outermost layer."""
+        # https://stackoverflow.com/a/50888571/
+        if name.startswith("__") and name.endswith("__"):
```
rmudambi, line 517: Shouldn't this be inside the `if name not in self:` block?

zmbc: Hm... I didn't fully think this through. My initial idea was that it should be consistent between getting and setting. The way I've written it here, we allow ConfigTree to hold keys that look like dunders, but they can't be get/set with dot syntax, only with bracket syntax.

If you'd prefer that they can be get/set with dot syntax _once they exist_ but can only be _created_ with bracket syntax (I think this is how pandas columns work), I could do what you suggest, and make a change to the `__getattr__` as well to make that work if the key already exists.

There is probably some argument for not allowing them at all, but that seems like a bigger change.

rmudambi (approved):

mattkappel (approved):

tests/config_tree/test_basic_functionality.py
```diff
@@ -439,6 +440,31 @@ def test_to_dict_yaml(test_spec):
     assert yaml_config == config.to_dict()
 
 
+@pytest.mark.xfail(reason="Not yet implemented")
+def test_equals():
```
rmudambi, line 444: Rather than xfailing, if we don't support equality operations on ConfigTrees we should override the `__eq__` function and throw a `NotImplementedError`.

rmudambi (changes requested):

rmudambi (approved):

PR #314, opened 2023-08-24
add tests of Component

rmudambi:
### Description

- *Category*: test
- *JIRA issue*: [MIC-4419](https://jira.ihme.washington.edu/browse/MIC-4419)

- Added tests of the newly implemented Component class
- Fixed typing on a function of InteractiveContext

### Testing
Unit tests pass

### DIFF
--- a/src/vivarium/interface/interactive.py
+++ b/src/vivarium/interface/interactive.py
@@ -176,7 +176,7 @@ class InteractiveContext(SimulationContext):
         """List all event types registered with the simulation."""
         return self._events.list_events()
 
-    def get_listeners(self, event_type: str) -> List[Callable]:
+    def get_listeners(self, event_type: str) -> Dict[int, List[Callable]]:
         """Get all listeners of a particular type of event.
 
         Available event types can be found by calling
--- a/tests/framework/components/test_component.py
+++ b/tests/framework/components/test_component.py
@@ -1,0 +1,252 @@
+from typing import List
+
+import pandas as pd
+
+from vivarium import Component, InteractiveContext
+from vivarium.framework.event import Event
+from vivarium.framework.population import SimulantData
+
+
+class ColumnCreator(Component):
+    def __repr__(self):
+        return "ColumnCreator()"
+
+    @property
+    def name(self) -> str:
+        return "column_creator"
+
+    @property
+    def columns_created(self) -> List[str]:
+        return ["test_column_1", "test_column_2", "test_column_3"]
+
+    def on_initialize_simulants(self, pop_data: SimulantData) -> None:
+        initialization_data = pd.DataFrame(
+            {column: 9 for column in self.columns_created}, index=pop_data.index
+        )
+        self.population_view.update(initialization_data)
+
+
+class ColumnRequirer(Component):
+    def __repr__(self):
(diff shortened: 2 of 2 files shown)


tests/framework/components/test_component.py
```diff
@@ -0,0 +1,252 @@
...
+    def __repr__(self):
+        return "AllColumnsRequirer()"
+
+    @property
+    def name(self) -> str:
+        return "all_columns_requirer"
+
+    @property
+    def columns_required(self) -> List[str]:
+        return []
```
albrja, line 73: test_column_1-4 because of empty list correct?

tests/framework/components/test_component.py
```diff
@@ -0,0 +1,252 @@
...
+    component = NoPopulationView()
+    simulation = InteractiveContext(components=[component])
+
+    time_step_prepare_methods = simulation.get_listeners("time_step__prepare")
+    time_step_methods = simulation.get_listeners("time_step")
+    time_step_cleanup_methods = simulation.get_listeners("time_step__cleanup")
+    collect_metrics_methods = simulation.get_listeners("collect_metrics")
+
+    for i in range(10):
+        assert component.on_time_step_prepare not in set(time_step_prepare_methods.get(i, []))
```
albrja, line 214: Here and all the instances below I am a bit confused with how you know what integer key to call in the dict. Can you explain that a bit because I see above you did change the typing for the method to be a Dict[int, list]?

src/vivarium/interface/interactive.py
```diff
@@ -176,7 +176,7 @@ def list_events(self) -> List[str]:
         """List all event types registered with the simulation."""
         return self._events.list_events()
 
-    def get_listeners(self, event_type: str) -> List[Callable]:
+    def get_listeners(self, event_type: str) -> Dict[int, List[Callable]]:
```
mattkappel, line 179: Good find.

tests/framework/components/test_component.py
rmudambi: I'm not sure I understand your question, but the empty list here means that the component requires all columns.

tests/framework/components/test_component.py
rmudambi: The integer key here is the priority of the call. By default calls are registered to be at priority 5. Here I'm iterating through all the possible priorities to demonstrate that this call wasn't registered on any of them. Below, I know that the default priority is 5, so I'm checking there for `DefaultPriorities`, since that component doesn't specify a priority. For `CustomPriorities`, I specified those particular values, so I want to be sure that they have been registered in the right place.

See the comments in the definition of `register_listener` here fore more details: https://github.com/ihmeuw/vivarium/blob/main/src/vivarium/framework/event.py#L286

mattkappel (approved):

tests/framework/components/test_component.py
albrja: Yes was verifying that.

tests/framework/components/test_component.py
albrja: Oh ok that makes sense.

albrja (approved):

PR #319, opened 2023-08-25
properly register listeners for post-setup and simulation-end

rmudambi:
### Description

- *Category*: bugfix
- *JIRA issue*: [MIC-4429](https://jira.ihme.washington.edu/browse/MIC-4429)

Implements:
- Add post-setup and simulant_end priorities
- Register post-setup and simulant_end listeners
- Define default post-setup and simulant_end functions
- Improve docstrings

### Testing
Updated unit tests


src/vivarium/component.py
```diff
@@ -92,34 +92,41 @@ def initialization_columns_required(self) -> List[str]:
         """
         return []
 
+    @property
+    def post_setup_priority(self) -> int:
+        """The priority of this component's post_setup listener if it exists."""
+        return 5
```
mattkappel, line 98: Should we continue using bald integers here? Like would it be better to have this defined as a constant elsewhere?

mattkappel (approved):
Approved with a question.

src/vivarium/component.py
rmudambi: Good idea

stevebachmeier (approved):

PR #321, opened 2023-08-26
Feature/mic 4441 machine

rmudambi:
### Description

- *Category*: refactor
- *JIRA issue*: [MIC-4441](https://jira.ihme.washington.edu/browse/MIC-4441)

Changes:

- Updates the `Machine` component to inherit from `Component
- Updates the example disease model component so that doctests pass
- Updates default `__repr__` to use kwarg-style arguments
- Updates default `__repr__` and `name` to ignore arguments to `__init__` that don't match an attribute (rather than erroring)

### Testing
Automated tests pass. Example disease model doc tests pass

Need to test using US CVD

### DIFF
--- a/tests/framework/components/test_component.py
+++ b/tests/framework/components/test_component.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import List, Optional
 
 import pandas as pd
 
@@ -43,6 +43,12 @@ class AllColumnsRequirer(Component):
     @property
     def columns_required(self) -> List[str]:
         return []
+
+
+class FilteredPopulationView(ColumnRequirer):
+    @property
+    def population_view_query(self) -> Optional[str]:
+        return "test_column_1 == 5"
 
 
 class NoPopulationView(Component):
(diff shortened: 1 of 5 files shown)


tests/framework/components/test_component.py
```diff
@@ -55,14 +55,18 @@ class NoPopulationView(Component):
...
+    def __init__(self, p_one: str, p_two: int, p_three: str):
+        super().__init__()
+        self.p_one = p_one
+        self.p_two = p_two
+        self._p_three = p_three
+
 
-    def __init__(self, parameter_one: str, parameter_two: int):
+class ParameterizedByComponent(Component):
+    def __init__(self, param: Parameterized):
```
albrja, line 67: I feel like I'm missing the whole point of this. Both ParameterizedByComponent and Parameterized are subclasses of component. Based on the implementation for these two classes for this test, it just seems like an extra unnecessary step.  What would the use case be for this instead of just implementing Parameterized?

albrja (approved):

tests/framework/components/test_component.py
rmudambi: The difference is that `ParameterizedByComponent` takes a Component as an argument for its constructor. This is being used in the tests `test_component_by_other_component_has_repr_that_incorporates_arguments` and test_component_by_other_component_has_name_that_incorporates_arguments` which are testing how `__repr__` and `name` are generated.

PR #335, opened 2023-09-05
Make ConfigTree objects pickleable

rmudambi:
### Description

- *Category*: feature
- *JIRA issue*: N/A

Changes:
- ConfigTree objects are now pickleable
- Equality doesn't work on ConfigTrees. Rather than just always returning False, we now throw a `NotImplementedError` 

### Testing
Automated testing added.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,7 @@
+**1.2.5 - 09/05/23**
+
+ - Update ConfigTree to make it pickleable; raise NotImplementedError on equality calls
+
 **1.2.4 - 09/01/23**
 
  - Create LookupTableData type alias for the source data to LookupTables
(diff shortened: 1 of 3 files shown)


zmbc:
This change was motivated by https://github.com/ihmeuw/pseudopeople/pull/265. While that PR is only a spike of a potential feature, it is clear that if we want to distribute pseudopeople noising between processes/machines, it will be much more convenient if we can pass the noising config object (which is a Vivarium ConfigTree) between those processes/machines.

Before this PR, attempting to do so resulted in the error `vivarium.config_tree.ConfigurationKeyError: 'No value at name __getstate__.'` which was encountered when the pickle library tried to serialize the ConfigTree, as described in #313.

After this PR, the code in that pseudopeople branch succeeds, serializing and de-serializing the config so that it can be accessed within each worker process.

albrja (approved):

patricktnast (approved):

PR #338, opened 2023-09-13
Address Pandas >2.0.0 FutureWarnings

patricktnast:
### Description

- *Category*: 
feature

- *JIRA issue*: [MIC-4076](https://jira.ihme.washington.edu/browse/MIC-4076)

 
This fixes an error that arises from a change to how groupby objects return group labels in pandas > 2.0.  From the 
pandas deprecation warning: "A length 1 tuple will be returned when iterating over a groupby with a grouper equal to a list of length 1" We map possible transition states onto integers, group by that mapping, and then try to use it as a list index, which fails if groupby returns a tuple.

It seemed preferable to group more directly, but naive methods of doing so didn't seem to work for one reason or another. However, by grouping by a `pd.Categorical` object (as described in https://pandas.pydata.org/pandas-docs/stable/user_guide/groupby.html#handling-of-un-observed-categorical-values), we can avoid the list comprehension, as well as obviating the rest of the logic in the method, which is doing the same thing that `observed=False` does. 

It's probably possible to slim this down even more and just get rid of the internal method, but I thought it might be best to leave the structure intact.

### Testing

I checked manually in the CVD repository that the method returns identical results (apart from the case where the existing behavior fails), and existing tests for state_machine.py pass.

I added (more like copied) a test for the specific situation where the existing behavior will fail with pandas 2.0 in CVD, that is, if the index passed into `_groupby_new_state` contains a single simulant, the groupby object will return a tuple of length 1 from the output map instead of an integer. It might be better to integrate this into the existing tests somehow, since it is very similar; I'm happy to take opinions/suggestions.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,7 @@
+**1.2.6 - 09/14/23**
+
+ - Update state machine to prepare for pandas 2.0
+
 **1.2.5 - 09/05/23**
 
  - Update ConfigTree to make it pickleable; raise NotImplementedError on equality calls
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ if __name__ == "__main__":
 
     install_requirements = [
         "numpy",
-        "pandas",
+        "pandas<2.0.0",
         "pyyaml>=5.1",
         "scipy",
         "click",
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ if __name__ == "__main__":
 
     install_requirements = [
         "numpy",
-        "pandas<2.0.0",
+        "pandas",
         "pyyaml>=5.1",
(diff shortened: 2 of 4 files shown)


rmudambi (approved):
Looks great!

Does this remove all of our FutureWarnings?

patricktnast:
> Looks great!
> 
> Does this remove all of our FutureWarnings?

Running in CVD, I do not see any other futurewarnings with pandas 2.0.3, but there is a new deprecation in pandas 2.1.0 that we run into quite a bit:
https://pandas.pydata.org/docs/dev/whatsnew/v2.1.0.html#deprecated-silent-upcasting-in-setitem-like-series-operations

hussain-jafari (approved):

PR #340, opened 2023-09-15
Unpin Pandas

patricktnast:
### Description

- *Category*: 
misc
- *JIRA issue*: [MIC-4545](https://jira.ihme.washington.edu/browse/MIC-4545)

 
Pandas 2.1.0 features two new deprecations (at least, two that we must address):
https://pandas.pydata.org/docs/dev/whatsnew/v2.1.0.html#deprecations

One is silent upcasting, for which we just need to define pd.Series(0, ...) or DataFrame with a value of 0.0 instead, so it starts out with float data

The second is that for categorical groupers, `groupby()` is changing the default of `observed` from `False` to `True`. I think the appropriate thing to do here is preserve the existing behavior and set observed=False, so that the full cartesian product of categories is included. 
 
### Testing

Tested in US CVD; no futurewarnings (from vivarium specifically, others to be addressed in other PRs). Sim runs, CI passes; by inspection, changes are not behavior-changing.


setup.py
```diff
@@ -32,7 +32,7 @@
 
     install_requirements = [
         "numpy",
-        "pandas<2.1.0",
+        "pandas",
```
albrja, line 35: Would it hurt to make this >2.1.0?

albrja (approved):

setup.py
patricktnast: I was a little wary to do this, since `pip` was having trouble in US CVD with resolving all the dependencies out of the box with pandas bounded from below (in CVD's `setup.py`, not `vivarium`), and I didn't want to break anything for anybody. But I guess with vivarium's pandas bounded from below, it's still fine, since pip can just install the release before this.

rmudambi: Correct me if I'm wrong, but I think the way you have it implemented, things should still work in `pandas<2.1.0`. In that case pinning a lower bound is not correct (i.e. it is correct as you have it).

rmudambi (approved):

setup.py
patricktnast: yes, it is backwards compatible

PR #341, opened 2023-09-16
Remove the need for all setup methods to call super

rmudambi:
### Description

- *Category*: bugfix
- *JIRA issue*: [MIC-4543](https://jira.ihme.washington.edu/browse/MIC-4543)

### Changes
The real changes are all in 3 files: 
- `src/vivarium/component.py`
- `src/vivarium/framework/components/manager.py`
- `src/vivarium/manager.py`

The rest is just removing calls to `super().setup(builder)` and updating typing in one place.

In `component.py`, rename `setup` to `setup_component` and call the new `setup` method which is empty. This has the added benefit of ordering calls better. Previously, we had to make the `super` call later in `setup` for some components, because we needed some code to have run in the component's setup before we could create a population view or register a listener properly. Now we can define the logger first and do the population view creation and listener registration last, like we usually want to.

In `src/vivarium/framework/components/manager.py` we update component setup to call `setup_component` for all components, while continuing to call `setup` for managers (and non-component components, though that will soon not be possible to exist).

In `manager.py` we define an empty `setup` function just so that all managers will have one for the above method.

### Testing
Automated tests pass. Ran US CVD simulation for a few steps

### DIFF
--- a/src/vivarium/component.py
+++ b/src/vivarium/component.py
@@ -181,9 +181,10 @@ class Component(ABC):
         self.logger = None
         self.population_view: Optional[PopulationView] = None
 
-    def setup(self, builder: "Builder") -> None:
+    def setup_component(self, builder: "Builder") -> None:
         """Method that vivarium will run during the setup phase."""
         self.logger = builder.logging.get_logger(self.name)
+        self.setup(builder)
         self.set_population_view(builder)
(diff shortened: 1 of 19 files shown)


stevebachmeier (approved):

patricktnast (approved):
LGTM

PR #343, opened 2023-09-18
datetime and Timestamp are incompatible in run_until

patricktnast:
##datetime and Timestamp are incompatible in run_until

### Description

- *Category*: 
documentation
- *JIRA issue*: [MIC-2706](https://jira.ihme.washington.edu/browse/MIC-2706)

 
The bug as described in MIC-2706 is explicitly coded:

```
if not isinstance(end_time, type(self._clock.time)):            
     raise ValueError(f"Provided time must be an instance of {type(self._clock.time)}") 
```
the clock is a pd.Timestamp, and datetime is not a subclass of Timestamp (namely, the reverse is true)

I see three options, in order of increasing effort:

1. This is working as intended, but needs a simple addition to the docstring in line with elsewhere e.g. ln 88 (
Must be the same type as the simulation clock's step size (usually a pandas.Timedelta)
.)
2. datetime and timestamp are interoperable, so we should just allow for this flexibility (as well as in the other places in `interactive.py` where we enforce the clock type explicitly)
3. We could go even further than (1) and just decide to not support non-pandas datetime clocks and remove datetime/timedelta from Time and Timedelta types. It doesn't seem to me like it would cause any problems, since it should be easy to convert datetime to pandas.Timestamp, and vivarium requires pandas anyway.

My preference would be to just do (1), and this PR is made in that spirit. Would be interested to hear opinions in the comments, if you disagree.
### Testing

Made sure that Timestamp and datetime work together, but an integer passed in with a timestamp clock doesn't.

### DIFF
--- a/src/vivarium/interface/interactive.py
+++ b/src/vivarium/interface/interactive.py
@@ -48,14 +48,17 @@ class InteractiveContext(SimulationContext):
         Parameters
         ----------
         step_size
-            An optional size of step to take. Must be the same type as the
+            An optional size of step to take. Must be compatible with the
             simulation clock's step size (usually a pandas.Timedelta).
         """
         old_step_size = self._clock.step_size
         if step_size is not None:
-            if not isinstance(step_size, type(self._clock.step_size)):
+            if not (
+                isinstance(step_size, type(self._clock.step_size))
+                or isinstance(self._clock.step_size, type(step_size))
+            ):
                 raise ValueError(
-                    f"Provided time must be an instance of {type(self._clock.step_size)}"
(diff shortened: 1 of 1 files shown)


albrja (approved):

albrja:
I agree with option 1. I think option 3 is also a really good idea.

stevebachmeier (approved):
I personally like the idea of supporting all reasonable inputs a user might provide.

rmudambi (approved):
We don't want to do 3, because while we always use the DateTimeClock, it is reasonable that someone might want to create a simulation that is not operating using datetime and as some other means of measuring "time".

rmudambi (approved):

PR #351, opened 2023-09-22
Mic-4381/Run all tests on deploy

albrja:
### Runs all tests on deploy
- *Category*: Release
- *JIRA issue*: [MIC-4381](https://jira.ihme.washington.edu/browse/MIC-4381)

-runs full test suite on release

### Testing
All tests pass

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -18,6 +18,10 @@ jobs:
           python --version
           python -m pip install --upgrade pip
           pip install setuptools wheel twine
+      - name: Test
+        run: |
+          pip install .[test]
+          pytest --runslow ./tests
       - name: Build and publish
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -18,10 +18,15 @@ jobs:
           python --version
           python -m pip install --upgrade pip
           pip install setuptools wheel twine
+      - name: Test
```
albrja, line 21: I figured we should run this before build and publish step. Im do not believe the pip install will affect the follow step wheere we publish.

rmudambi (approved):

.github/workflows/deploy.yml
rmudambi: Let's check after I release vivarium later today to see if this affects the published version of vivarium

stevebachmeier (approved):

PR #356, opened 2023-10-12
Refactor ComponentConfigurationParser creating Components as they are parsed

rmudambi:
### Description

- *Category*: refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4597

### Changes and notes
- Refactored all functions to be instance or static method of `ComponentConfigurationParser` (first commit)
- Refactored parser to create components iteratively as they are parsed. 
- Updated unit tests to match new usage

**Note: it may be easier to understand the functionality changes if you exclude the first commit while reviewing.**

### Testing
Automated tests pass. Ran CVD sim with it.

### DIFF
--- a/src/vivarium/framework/components/parser.py
+++ b/src/vivarium/framework/components/parser.py
@@ -36,15 +36,16 @@ class ParsingError(ComponentConfigError):
 
 
 class ComponentConfigurationParser:
-    """Parses component configuration from model specification and initializes
+    # todo update docstring
+    """
+    Parses component configuration from model specification and initializes
     components.
 
     To define your own set of parsing rules, you should write a parser class
     that inherits from this class and overrides ``parse_component_config``.
     You can then define a set of parsing rules that turn component configuration
-    information into a list of strings where each string is the full python
-    import path to the class followed by a set of parentheses containing
-    initialization arguments.
+    information into a string which is the full python import path to the class
+    followed by a set of parentheses containing initialization arguments.
 
     For example, say your component configuration specifies ``FancyClass`` from
     the ``important_module`` of your ``made_up_package`` and that ``FancyClass``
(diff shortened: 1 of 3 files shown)


src/vivarium/framework/components/parser.py
```diff
@@ -180,85 +152,84 @@ def _process_level(level, prefix):
...
-    and required arguments.
+        Returns
+        -------
+        List[Tuple[str, Tuple]]
+            List of component/argument tuples.
+        """
+        components = []
+        for c in component_list:
+            path, args_plus = c.split("(")
+            cleaned_args = self.clean_args(args_plus[:-1].split(","), path)
```
stevebachmeier: Does `self.clean_args()` work when clean_args is static?

src/vivarium/framework/components/parser.py
```diff
@@ -180,85 +152,84 @@ def _process_level(level, prefix):
...
+            components.append((path, cleaned_args))
+        return components
+
+    @staticmethod
```
stevebachmeier, line 224: Also...why use a static method? I'm still uncomfortable when/where to use these. In this case I see how the cleaning method should not be specific to each instance, but it also seems to be the case for `prep_components` above which you did not specify as static.

src/vivarium/framework/components/parser.py
```diff
@@ -87,23 +88,48 @@ def get_components(
...
+        component_config
+            A ConfigTree representing a hierarchical component specification blob.
+
+        Returns
+        -------
+        List[Component]
+            A flat list of Components
+        """
+        if not component_config:
+            return []
```
stevebachmeier, line 118: When would component_config ever be empty?

src/vivarium/framework/components/parser.py
```diff
@@ -128,60 +154,77 @@ def parse_component_config(
...
-        if isinstance(level, list):
-            return [".".join(prefix + [child]) for child in level]
-
         component_list = []
-        for name, child in level.items():
-            component_list.extend(self._process_level(child, prefix + [name]))
+        if isinstance(level, dict):
+            for name, child in level.items():
+                component_list.extend(self.process_level(child, prefix + [name]))
+        else:
```
stevebachmeier, line 177: Any way an improperly-formatted yaml would result in this not being a list?

src/vivarium/framework/components/parser.py
rmudambi: Yes, you can call a static method either off the class itself or off an instance of it:  i.e. `self.static_method` or `Class.static_method`

stevebachmeier (approved):

src/vivarium/framework/components/parser.py
rmudambi: I believe it would be something like:
```yaml
components:

```

which would create a simulation with no components. I'm not sure if that's a desirable input, but in any case, this is not a change from the way it previously worked (see `parse_component_config_to_list` in the previous version).

src/vivarium/framework/components/parser.py
rmudambi: Yes that is possible (and was possible before as well). I can make a change to catch this case, though I'm not sure how important it is

src/vivarium/framework/components/parser.py
rmudambi: This article does a good job of explaining when to use them (and class methods): https://realpython.com/instance-class-and-static-methods-demystified/

src/vivarium/framework/components/parser.py
rmudambi: Done this now and added test coverage for it @stevebachmeier

mattkappel (approved):

PR #359, opened 2023-10-19
create categorical lookup table

rmudambi:
### Description

- *Category*: feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4642

### Changes and notes
- Created `CategoricalTable` for LookupTables that have only categorical variables

### Testing
Created new automated tests to cover new use case. New test failed before the fix and now passes.

### DIFF
--- a/src/vivarium/framework/lookup/table.py
+++ b/src/vivarium/framework/lookup/table.py
@@ -124,18 +124,14 @@ class CategoricalTable:
     Attributes
     ----------
     data
-        The data from which to build the interpolation.
+        The categorical data from which to select values.
     population_view
         View of the population to be used when the table is called with an
         index.
     key_columns
-        Column names to be used as categorical parameters in Interpolation
-        to select between interpolation functions.
+        Column names to be used as categorical parameters.
     value_columns
-        Names of value columns to be interpolated over. All non parameter- and
-        key- columns in data.
-    validate
-        Whether to validate the data and column names.
+        Names of value columns to be retrived. All non key-columns in data.
 
     Notes
     -----
@@ -148,15 +144,12 @@ class CategoricalTable:
         self,
         data: pd.DataFrame,
         population_view: PopulationView,
-        key_columns: Union[List[str], Tuple[str]],
-        value_columns: Union[List[str], Tuple[str]],
-        validate: bool,
+        key_columns: Union[List[str], Tuple[str]]
     ):
         self.data = data
         self.population_view = population_view
         self.key_columns = key_columns
-        self.value_columns = value_columns
-        self.validate = validate
+        self.value_columns = self.data.columns.difference(set(self.key_columns))
 
     def __call__(self, index: pd.Index) -> pd.DataFrame:
         """Get the mapped values for the rows in ``index``.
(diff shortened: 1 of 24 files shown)


hussain-jafari (approved):

src/vivarium/framework/lookup/table.py
```diff
@@ -209,6 +291,11 @@ def __init__(
         # Note datetime catches pandas timestamps
         if isinstance(data, (Number, datetime, timedelta, list, tuple)):
             self._table = ScalarTable(data, value_columns)
+        elif parameter_columns is None:
```
stevebachmeier, line 294: if parameter_columns can now be None, change type hint above to Optional

stevebachmeier, line 294: I'm kinda confused by this. Can you explain the logic in this if block? How is it that if `data` is not Number, datetime, timedelta, list, or tuple and `parameter_columns` is None means that this is a categorical table?

src/vivarium/framework/lookup/table.py
```diff
@@ -115,6 +116,87 @@ def __repr__(self) -> str:
...
 
+class CategoricalTable:
+    """
+    A callable that selects values from a table based on categorical parameters
+    across an index.
+
+    Attributes
+    ----------
+    data
+        The data from which to build the interpolation.
```
stevebachmeier, line 127: No interpolation here

src/vivarium/framework/lookup/table.py
```diff
@@ -115,6 +116,87 @@ def __repr__(self) -> str:
...
+    population_view
+        View of the population to be used when the table is called with an
+        index.
+    key_columns
+        Column names to be used as categorical parameters in Interpolation
```
stevebachmeier, line 132: Or here

src/vivarium/framework/lookup/table.py
```diff
@@ -115,6 +116,87 @@ def __repr__(self) -> str:
...
+        Returns
+        -------
+            A table with the interpolated values for the population requested.
+        """
+        pop = self.population_view.get(index)
+        del pop["tracked"]
+
+        # specify some numeric type for columns, so they won't be objects but
+        # will be updated with whatever column type it actually is
+        result = pd.DataFrame(index=pop.index, columns=self.value_columns, dtype=np.float64)
```
stevebachmeier, line 178: Won't this just get changed to object anyway? So if result["category"] is initialized as float64 but then down below you add non-floag64 objects it will just become object?

src/vivarium/framework/lookup/table.py
hussain-jafari: If the data is a Number, datetime, timedelta (ScalarValue types in lookup.py) or a list or tuple of these types, it can be built using a ScalarTable. Otherwise, it's a DataFrame (the type checking for data in the init of LookupTable allows for Union[ScalarValue, pd.DataFrame, List[ScalarValue], Tuple[ScalarValue]]). 

The DataFrame will have parameter columns or key columns (or both). Parameter columns contain continuous values and key columns contain categorical values. So if we have no parameter columns we only have key columns aka categorical values.

src/vivarium/framework/lookup/table.py
rmudambi: My understanding is that it will remain float64 as long as all data is of a numeric dtype.

albrja (approved):

PR #363, opened 2023-10-23
enable LookupTables with DataFrame source data to specify value columns

rmudambi:
### Description

- *Category*: feature/refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4649

### Changes and notes
-  Enabled `LookupTables` with DataFrame source data to specify value columns
- Renamed `Interpolation` key and parameter column attributes 
- Added value_columns as an argument to the `Interpolation` constructor
- Moved test_lookup and test_interpolation into a lookup package
- Fixed some typing
- Introduced an error if user attempts to set key or parameter columns for a scalar table
- Gave the LookupTable constructor sensible default values so they can be created with kwargs

Note: this PR looks big, but most of it consists of non-substantial changes. Nothing of note has changed in `test_interpolation` except adding the value columns argument to all `__init__` calls.

### Testing
All automated tests pass. Wrote new tests.


tests/framework/lookup/test_lookup.py
```diff
@@ -253,53 +246,59 @@ def test_validate_parameters_fail_other_data(data):
...
+    [
+        (["a"], ["b"], ["c"]),
+        ([], ["b"], ["c", "a"]),
+        ([], ["b"], ["a", "c"]),
+        ([], ["b"], ["c"]),
+        (["a"], [], ["c"]),
+        (["a"], ["b"], []),
+        (["a"], [], []),
+        ([], ["b"], []),
+    ],
```
albrja, line 281: So in the following test if `validate_build_table_parameters` raises an error the test fails and if it doesn't it then passes correct?

albrja (approved):

tests/framework/lookup/test_lookup.py
rmudambi: That's correct!

mattkappel (approved):

hussain-jafari (approved):

PR #370, opened 2023-10-26
Individiual Clocks integration

patricktnast:
### Description

- *Category*: 
feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4637

### Changes and notes
 
Three primary changes:
exclude simulants from off-time events by filtering the time-aligned population
use the simulant step sizes in place of the global step in rescale post-processor
rename results manager "step size", since that is now the name of the local step column.

I also added an "event time" method in the time manager that just returns time + step_size; it's convenient in a number of places where i was otherwise calling X._clock.time + X._clock.step_size. Not sure if calling it "event time" is confusing, though.

### Testing

example sim runs without issue
Added tests to show that a simulant does get excluded from events (if they're supposed to be) and that the rescale post-processor properly uses the simulant step size when calculating rates. 
I've just changed step sizes by altering them directly and running a `.step()` to update the value (so that they will be excluded on the next step); it might be a bit nicer to fake a component to update the step size but this is more direct IMO

### DIFF
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -237,7 +237,7 @@ class SimulationContext:
         self._logger.debug(self._clock.time)
         for event in self.time_step_events:
             self._lifecycle.set_state(event)
-            pop_to_update = self._clock.aligned_pop(
+            pop_to_update = self._clock.get_active_population(
                 self._population.get_population(True).index,
                 self._clock.event_time,
             )
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -237,7 +237,10 @@ class SimulationContext:
         self._logger.debug(self._clock.time)
         for event in self.time_step_events:
             self._lifecycle.set_state(event)
-            self.time_step_emitters[event](self._population.get_population(True).index)
+            pop_to_update = self._clock.aligned_pop(
+                self._population.get_population(True).index, event.time
+            )
+            self.time_step_emitters[event](pop_to_update.index)
         self._clock.step_forward(self._population.get_population(True).index)
 
     def run(self) -> None:
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -238,7 +238,8 @@ class SimulationContext:
         for event in self.time_step_events:
             self._lifecycle.set_state(event)
             pop_to_update = self._clock.aligned_pop(
-                self._population.get_population(True).index, self._clock.time + self._clock.step_size
+                self._population.get_population(True).index,
+                self._clock.time + self._clock.step_size,
             )
             self.time_step_emitters[event](pop_to_update.index)
         self._clock.step_forward(self._population.get_population(True).index)
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -239,7 +239,7 @@ class SimulationContext:
             self._lifecycle.set_state(event)
             pop_to_update = self._clock.aligned_pop(
                 self._population.get_population(True).index,
-                self._clock.time + self._clock.step_size,
+                self._clock.event_time,
             )
             self.time_step_emitters[event](pop_to_update.index)
         self._clock.step_forward(self._population.get_population(True).index)
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -238,7 +238,7 @@ class SimulationContext:
         for event in self.time_step_events:
             self._lifecycle.set_state(event)
             pop_to_update = self._clock.aligned_pop(
-                self._population.get_population(True).index, event.time
+                self._population.get_population(True).index, self._clock.time + self._clock.step_size
             )
             self.time_step_emitters[event](pop_to_update.index)
         self._clock.step_forward(self._population.get_population(True).index)
--- a/src/vivarium/framework/results/manager.py
+++ b/src/vivarium/framework/results/manager.py
@@ -236,7 +236,7 @@ class ResultsManager(Manager):
     def _add_resources(self, target: List[str], target_type: SourceType):
         if not len(target):
             return  # do nothing on empty lists
-        target = set(target) - {"event_time", "current_time", "step_size"}
+        target = set(target) - {"event_time", "current_time", "event_step_size"}
         if target_type == SourceType.COLUMN:
             self._required_columns.update(target)
         elif target_type == SourceType.VALUE:
@@ -247,7 +247,7 @@ class ResultsManager(Manager):
             event.index
         )
         population["current_time"] = self.clock()
-        population["step_size"] = event.step_size
+        population["event_step_size"] = event.step_size
         population["event_time"] = self.clock() + event.step_size
         for k, v in event.user_data.items():
             population[k] = v
--- a/src/vivarium/framework/time.py
+++ b/src/vivarium/framework/time.py
@@ -56,7 +56,7 @@ class SimulationClock(Manager):
         if not self._clock_step_size:
             raise ValueError("No step size provided")
         return self._clock_step_size
-    
+
     @property
     def event_time(self) -> Time:
         "Convenience method for event time, or clock + step"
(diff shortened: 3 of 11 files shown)


tests/framework/components/mocks.py
```diff
@@ -92,20 +92,29 @@ def __init__(self, *args, name="test_listener"):
         self.collect_metrics_called = False
         self.simulation_end_called = False
 
+        self.time_step_prepare_index = None
+        self.time_step_index = None
+        self.time_step_cleanup_index = None
+        self.collect_metrics_index = None
```
mattkappel, line 98: Should these like objects be contained in something, dict where keys are phases?

mattkappel (approved):

src/vivarium/framework/engine.py
```diff
@@ -237,7 +237,11 @@ def step(self) -> None:
         self._logger.debug(self._clock.time)
         for event in self.time_step_events:
             self._lifecycle.set_state(event)
-            self.time_step_emitters[event](self._population.get_population(True).index)
+            pop_to_update = self._clock.aligned_pop(
```
rmudambi, line 240: I don't love the name `aligned_pop`. What about something like `get_active_index` or something like that?

src/vivarium/framework/results/manager.py
```diff
@@ -236,7 +236,7 @@ def register_observation(
     def _add_resources(self, target: List[str], target_type: SourceType):
         if not len(target):
             return  # do nothing on empty lists
-        target = set(target) - {"event_time", "current_time", "step_size"}
+        target = set(target) - {"event_time", "current_time", "event_step_size"}
```
rmudambi, line 239: Is this the "global" step-size? If so, is it used by any existing observers? Is there a compelling reason that an observer would need this?

src/vivarium/framework/time.py
```diff
@@ -84,21 +89,21 @@ def simulant_next_event_times(self, index: pd.Index) -> pd.Series:
         """The next time each simulant will be updated."""
         if not self.population_view:
             raise ValueError("No population view defined")
-        return self.population_view.subview(["next_event_time"]).get(index)
+        return self.population_view.subview(["next_event_time"]).get(index).squeeze()
```
rmudambi, line 92: Shouldn't this be `.squeeze{axis=1)` here to ensure we don't squeeze to a scalar if there is only one row?

src/vivarium/framework/time.py
```diff
@@ -84,21 +89,21 @@ def simulant_next_event_times(self, index: pd.Index) -> pd.Series:
...
 
     def simulant_step_sizes(self, index: pd.Index) -> pd.Series:
         """The step size for each simulant."""
         if not self.population_view:
             raise ValueError("No population view defined")
-        return self.population_view.subview(["step_size"]).get(index)
+        return self.population_view.subview(["step_size"]).get(index).squeeze()
```
rmudambi, line 98: Same here.

src/vivarium/framework/values.py
```diff
@@ -261,7 +271,7 @@ def name(self):
 
     def setup(self, builder):
         self.logger = builder.logging.get_logger(self.name)
-        self.step_size = builder.time.step_size()
+        self.step_size = builder.time.simulant_step_sizes
```
rmudambi, line 274: I think I missed this in the previous PR (or maybe I just said the wrong thing entirely), but this should be 
```python
self.step_sizes = builder.time.simulant_step_sizes()
```
and in the `TimeInterface` we should have 
```python
def simulant_step_sizes(self) -> Callable:
    return lambda: self._manager.simulant_step_sizes
```
and the same for `simulant_next_event_times`.

src/vivarium/framework/values.py
```diff
@@ -110,6 +110,16 @@ def rescale_post_processor(value: NumberLike, time_step: pd.Timedelta):
         The annual rates rescaled to the size of the current time step size.
 
     """
+    if isinstance(time_step, Callable):
```
rmudambi, line 113: I think I'm missing something, but how would `time_step` ever be anything but a `Callable`?

tests/framework/test_time.py
```diff
@@ -24,13 +24,13 @@ def test_align_times(SimulationContext, base_config, components):
     sim = SimulationContext(base_config, components)
     sim.setup()
     sim.initialize_simulants()
-    pop_size = len(sim._population.get_population(True))
+    pop_size = len(sim.get_population())
```
rmudambi, line 27: Why did you drop `True` here?

tests/framework/test_time.py
```diff
@@ -70,45 +68,66 @@ def test_align_times(SimulationContext, base_config, components):
...
             sim._clock.aligned_pop(
-                sim._population.get_population(True).index,
-                sim._clock.time + sim._clock.step_size,
+                sim.get_population().index,
+                sim._clock.event_time,
             )
         )
         == pop_size
     )
+    # Check that we actually include 7 in events
```
rmudambi, line 128: What's the significance of 7?

tests/framework/test_values.py
```diff
@@ -2,14 +2,21 @@
...
 
 
 @pytest.fixture
 def manager(mocker):
     manager = ValuesManager()
     builder = mocker.MagicMock()
-    builder.step_size = lambda: lambda: pd.Timedelta(days=1)
+    builder.time.simulant_step_sizes = lambda: lambda idx: pd.Series(
+        pd.Timedelta(days=3), index=idx
+    )
```
rmudambi, line 19: It'd be a better test if the time-deltas were not all the same

rmudambi (changes requested):
Mostly looks good, but there are a few minor things that I think we should address.

src/vivarium/framework/values.py
stevebachmeier: @rmudambi Why? Just to keep builder's interface consistent w/ callables?

stevebachmeier (approved):

src/vivarium/framework/results/manager.py
patricktnast: It's the step size that comes from the event (which is defined by the global time ofc). I don't know of any observations/stratifications that use it, but "event_time" is in a commented out piece of stratification code
https://github.com/ihmeuw/vivarium_public_health/blob/0703d0964d6b92d99bb818db25802e8364fcd266/src/vivarium_public_health/metrics/stratification.py#L54
I preserved it just to be on the safe side, but happy to drop it altogether if it is not needed

src/vivarium/framework/values.py
patricktnast: I haven't added (or really fully thought through I think) the managing of swapping between local-functionality clocks and the old clocks without individual clocks, but the idea was that in the base case the `step_size` propery defined in `value` becomes the old `builder.time.step_size` which, when passed into the post-processor, is a `pd.Timedelta`.

tests/framework/test_time.py
patricktnast: `SimulationContext`'s get_population method defaults to True, as opposed to the population managers, which does not specify a default for `untracked`.

tests/framework/test_time.py
patricktnast: It's just a randomly chosen (small but nonzero) simulant index to operate the unit test on.

src/vivarium/framework/values.py
patricktnast: done

rmudambi: @stevebachmeier yes this is for making our interactions with builder interfaces consistent

src/vivarium/framework/engine.py
patricktnast: renamed to `get_active_population`

src/vivarium/framework/time.py
patricktnast: done

tests/framework/components/mocks.py
patricktnast: yeah that is better

tests/framework/test_values.py
patricktnast: how is this?

rmudambi: Just because this means that we can test that it is in fact acting on each simulant individually.

rmudambi (approved):

tests/framework/test_values.py
patricktnast: Oh, I agree, I meant how did the addition I made (to use 3 days for even and 5 days for odd) look, to be responsive to your comment

PR #371, opened 2023-10-31
Implement Step Size as Pipeline

patricktnast:
### Description

- *Category*: 
feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4639

### Changes and notes
 
In the minimum implementation, we make local step a state column. Let's upgrade it to a pipeline (that also outputs a state table column)
Also add a step size post-processor, which chooses the maximum of (global step size, min(pipeline modifiers))
### Testing

added tests to make sure that pipeline gets updated appropriately
I wasn't sure if it was a good idea to basically merge the new test_time.py into the old ones, where instead of putting in a component that modifies the step size, I just manually update the step size in the pop table. I can see an argument for integrating them, or for keeping them separate. Let me know what you think!


src/vivarium/framework/values.py
```diff
@@ -170,6 +172,29 @@ def union_post_processor(values: List[NumberLike], _) -> NumberLike:
     return joint_value
 
 
+def step_size_post_processor(values: List[NumberLike], _) -> pd.Series:
```
rmudambi, line 175: This post-processor is very specific to step-size calculation (especially in its docstring), so it doesn't seem likely to be of general utility. This suggests to me that the post-processor should be defined in the `SimulationClock`, not here in `values.py`.

tests/framework/test_time.py
```diff
@@ -20,52 +29,44 @@ def components():
...
-            sim._clock.get_active_population(
-                sim.get_population().index,
-                sim._clock.event_time,
-            )
-        )
-        == pop_size
+    active_simulants = lambda: sim._clock.get_active_population(
+        sim.get_population().index,
+        sim._clock.event_time,
     )
```
rmudambi, line 53: Since you're calling this method in multiple tests, you should probably bring it outside the test:
```python
def active_simulants(sim):
    return sim._clock.get_active_population(
        sim.get_population().index, sim._clock.event_time
    )
```

tests/framework/test_time.py
```diff
@@ -114,16 +104,66 @@ def test_unequal_steps(SimulationContext, base_config, components):
...
+    )
+
+    pipeline = lambda: sim._values.get_value("simulant_step_size")(
+        sim._population.get_population(True).index
+    )
+
+    column = lambda: sim._population._population.step_size
+
+    assert np.all(pipeline() == column())
+    assert len(active_simulants()) == pop_size
```
rmudambi, line 131: Do we need these `len` asserts? Are they adding any value not present in the above test?

tests/framework/test_values.py
```diff
@@ -86,3 +87,35 @@ def test_rescale_postprocessor(manager):
     odds = value.iloc[lambda x: x.index % 2 == 1]
     assert np.all(evens == from_yearly(0.5, pd.Timedelta(days=3)))
     assert np.all(odds == from_yearly(0.5, pd.Timedelta(days=5)))
+
+
+def test_step_size_post_processor(manager):
```
rmudambi, line 92: If we move this post processor inside the clock as I suggested above, this test should move inside `test_time.py`

tests/framework/test_time.py
```diff
@@ -114,16 +104,66 @@ def test_unequal_steps(SimulationContext, base_config, components):
...
+    )
+    column = lambda: sim._population._population.step_size
+
+    ## Everyone starts active
+    assert np.all(pipeline() == column())
+    assert len(active_simulants()) == pop_size
+
+    ## Go through a couple simulant step cycles
+    for _ in range(2):
+        for _ in range(math.ceil(step_modifier) - 1):
```
rmudambi, line 159: Is 3.5 effectively the same thing as 4?

rmudambi (approved):

src/vivarium/framework/time.py
```diff
@@ -69,6 +71,12 @@ def __init__(self):
         self.population_view = None
 
     def setup(self, builder: "Builder"):
+        self.step_size_pipeline = builder.value.register_value_producer(
```
stevebachmeier, line 74: Can you remind me what the motivation for promoting this to a pipeline is?

src/vivarium/framework/values.py
```diff
@@ -170,6 +172,29 @@ def union_post_processor(values: List[NumberLike], _) -> NumberLike:
...
+    Returns
+    -------
+    pandas.Series
+        The largest feasible step size for each simulant
+
+
+    """
+    if len(values) == 1:
+        return values[0]
+    min_modified = pd.DataFrame(values[1:]).min(axis=0).astype("timedelta64[ns]")
```
stevebachmeier, line 194: Can you explain the indexing here? What is special about the 0th item?

src/vivarium/framework/values.py
```diff
@@ -170,6 +172,29 @@ def union_post_processor(values: List[NumberLike], _) -> NumberLike:
...
+    return pd.DataFrame([values[0], min_modified]).max(axis=0)
```
stevebachmeier, line 195: I'm kinda confused why you want the max of the two. It seems like it would be the min?

src/vivarium/framework/time.py
```diff
@@ -69,6 +71,12 @@ def __init__(self):
...
+            "simulant_step_size",
+            source=lambda idx: [pd.Series(self.step_size, index=idx)],
+            preferred_combiner=list_combiner,
```
hussain-jafari, line 77: I forget - what does this list combiner do?

tests/framework/test_values.py
```diff
@@ -86,3 +87,35 @@ def test_rescale_postprocessor(manager):
...
+        preferred_post_processor=step_size_post_processor,
+    )
+    manager.register_value_modifier(
+        "test",
+        modifier=lambda idx: pd.Series(
+            [pd.Timedelta(days=7) if i % 2 == 0 else pd.Timedelta(days=5) for i in idx],
+            index=idx,
+        ),
+    )
+    manager.register_value_modifier(
```
hussain-jafari, line 108: Might be worth adding a comment explicitly saying what we're testing with this modifier.

tests/framework/test_values.py
```diff
@@ -86,3 +87,35 @@ def test_rescale_postprocessor(manager):
...
+        "test", modifier=lambda idx: pd.Series(pd.Timedelta(days=9), index=idx)
+    )
+    value = pipeline(index)
+    evens = value.iloc[lambda x: x.index % 2 == 0]
+    odds = value.iloc[lambda x: x.index % 2 == 1]
+    assert np.all(evens == pd.Timedelta(days=7))
+    assert np.all(odds == pd.Timedelta(days=5))
+
+    manager.register_value_modifier(
```
hussain-jafari, line 117: And this one.

hussain-jafari (approved):

src/vivarium/framework/values.py
patricktnast: values[0] in a `list_combiner` is always the source of the pipeline. We're taking out the source, which defined by the global step size, and all the component modifications to it. I'll explain why below

src/vivarium/framework/values.py
patricktnast: In the current formulation, we don't actually want a local step that is smaller than the global step. So we're taking the smallest of the modifications, and taking the larger between that and the global step. This may possibly change a bit when we use the simulant step sizes themselves to modify the global step (MIC-4538)

src/vivarium/framework/time.py
patricktnast: It returns a list of Series: first the source, and then any mutated versions of the pipeline (which doesn't necessarily have to be the same size as the source)

src/vivarium/framework/values.py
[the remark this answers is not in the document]
patricktnast: I think this doesn't quite work, since `values` is a list of `pd.Series` -- we'd need something like an iteration over the items in the series somewhere to get a row-wise minimum. I think it's fairly natural to think of a `List[pd.Series]` as basically a `DataFrame`, which we are squeezing down using `.min()` but YMMV. I will add a comment that makes this a bit more explicit.

tests/framework/test_time.py
patricktnast: Yeah, this is something that needs to get dealt with in the next ticket (MIC-4638), since we update steps based on the global clock, but update pipeline values based on the simulant step size, even if that doesn't conform to to the _actual_ number of steps that we took. I didn't totally appreciate that until this comment, so thanks!

tests/framework/test_time.py
patricktnast: No, i suppose not

PR #372, opened 2023-11-07
Align Step Sizes

patricktnast:
### Description

- *Category*: 
feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4638

### Changes and notes
 
I aligned the clocks in the following manner:
add a "min step size" property, so the global step size can change dynamically
adjust the step size post processor to rectify to multiples of the minimum step size
modify the clock step so that the next time step will always go to the smallest next event time.

I also moved around the step postprocessor to be a static method, which i neglected to do in the previous ticket.
### Testing

Added (and somewhat condensed) test coverage to show that pipeline calls have correct values, steps are skipped as appropriate, and uneven step modifiers produce the correct multiple of steps.

### DIFF
--- a/tests/framework/test_time.py
+++ b/tests/framework/test_time.py
@@ -48,18 +48,26 @@ def components():
     ]
 
 
-def active_simulants(sim):
-    return sim._clock.get_active_population(sim.get_population().index, sim._clock.event_time)
-
-
-def step_pipeline(sim):
-    return sim._values.get_value("simulant_step_size")(
+def validate_index_aligned(sim, expected_active_simulants):
+    """Ensure that the pipeline and column step sizes are aligned, and that the
+    active simulants are as expected BEFORE a step"""
+    step_pipeline = sim._values.get_value("simulant_step_size")(
         sim._population.get_population(True).index
     )
+    step_column = sim._population._population.step_size
+    active_simulants = sim._clock.get_active_population(
+        sim.get_population().index, sim._clock.event_time
+    )
+
+    assert np.all(step_pipeline == step_column)
+    assert active_simulants.index.equals(expected_active_simulants)
+    assert active_simulants.index.difference(expected_active_simulants).empty
 
 
-def step_column(sim):
-    return sim._population._population.step_size
+def validate_event_indexes(listener, expected_simulants):
+    """Make sure AFTER a step, that simulants were included in the right events"""
+    for index in listener.event_indexes.values():
+        assert index.equals(expected_simulants)
 
 
 def take_step(sim):
@@ -67,6 +75,16 @@ def take_step(sim):
     sim.step()
     new_time = sim._clock.time
     return new_time - old_time
+
+
+def take_step_and_validate(sim, listener, expected_simulants, expected_step_size_days):
+    """Take a step, and ensure that we included the right simulants, with the right step size"""
+    ## Check Before Timestep
+    validate_index_aligned(sim, expected_simulants)
+    ## Check Timestep
+    assert take_step(sim) == pd.Timedelta(days=expected_step_size_days)
+    ## Check After
+    validate_event_indexes(listener, expected_simulants)
 
 
 class StepModifier(MockGenericComponent):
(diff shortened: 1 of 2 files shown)


tests/framework/test_time.py
```diff
@@ -56,94 +62,106 @@ def step_column(sim):
...
+        )
 
-    def modify_step(self, index):
-        return pd.Series(pd.Timedelta(days=self.step_modifier), index=index)
-
+    def on_time_step(self, event: Event) -> None:
+        self.ts_pipeline_value = self.rate_pipeline(event.index)
 
-def test_align_times(SimulationContext, base_config, components):
+    def modify_step(self, index):
```
stevebachmeier, line 91: These tests are getting intensely detailed and I think some brief comments would go a long way

tests/framework/test_time.py
```diff
@@ -56,94 +62,106 @@ def step_column(sim):
...
-    # No simulants should be aligned after a step size adjustment
-    assert active_simulants(sim).empty
-
-    sim.step()
-    # Now they should be aligned again
-    assert len(active_simulants(sim)) == pop_size
+    for _ in range(2):
+        taken_step_size = take_step(sim)
+        # After a step (and no step adjustments, simulants should still be aligned)
+        assert taken_step_size == pd.Timedelta(days=1)
```
stevebachmeier, line 119: You might be able to move these assertions into a sort of `validate_aligned` helper method.

tests/framework/test_time.py
```diff
@@ -56,94 +62,106 @@ def step_column(sim):
...
     for index in listener.event_indexes.values():
-        assert 0 not in index
+        assert index.empty
 
-    # Check that everybody will step forward next step
     assert len(active_simulants(sim)) == pop_size
-
-    # Check that they are actually included in events
-    sim.step()
+    taken_step_size = take_step(sim)
```
stevebachmeier, line 146: Shouldn't active_simulants be empty here on the line above? You previously took a time step and `assert index.empty`.

patricktnast: The `active_simulants` method reutrns the indices of the simulants that are supposed to be included *before* timestep/lifecycle events are emitted. The `Listener` mock component stores the index acted upon *most recently* by timestep/liecycle events that were previously emitted. So here we are checking that before an event, we're "loading" the right simulants, as it were, and then, after doing the timestep, we verify that the same simulants were in fact updated. After the timestep has been taken, the `active_simulants` is now the group of simulants that will be updated upon the next iteration of events.

hussain-jafari (approved):

tests/framework/test_time.py
```diff
@@ -56,94 +62,123 @@ def step_column(sim):
...
+    SimulationContext, base_config, step_modifier_even, step_modifier_odd
+):
+    """Test that if everyone has some (non-minimum) step size, the global step adjusts to match"""
+    base_config["time"]["step_size"] = 1
+    listener = Listener("listener")
+    sim = SimulationContext(
+        base_config,
+        [StepModifier("step_modifier", step_modifier_even, step_modifier_odd), listener],
+    )
+    quantized_step = math.ceil(step_modifier_even)
```
hussain-jafari, line 181: Is this still what we want to do?

albrja (approved):
Those are some nice tests.

tests/framework/test_time.py
patricktnast: changed to `floor`, thanks

tests/framework/test_time.py
```diff
@@ -56,94 +62,123 @@ def step_column(sim):
...
+        # After a step (and no step adjustments, simulants should still be aligned)
+        assert taken_step_size == pd.Timedelta(days=1)
+        assert len(active_simulants(sim)) == pop_size
+        assert np.all(step_pipeline(sim) == step_column(sim))
+        for index in listener.event_indexes.values():
+            assert np.all(index == sim.get_population().index)
 
 
-def test_unequal_steps(SimulationContext, base_config, components):
+def test_empty_active_pop(SimulationContext, base_config, components):
```
rmudambi, line 139: I thought it was impossible to have an empty active population, since we'll always set the next time to the smallest next event time. That being said, we probably do need to test that the clock will still function if the population is empty or if there are no tracked simulants remaining in the population.

tests/framework/test_time.py
```diff
@@ -56,94 +62,123 @@ def step_column(sim):
...
+    SimulationContext, base_config, step_modifier_even, step_modifier_odd
+):
+    """Test that if everyone has some (non-minimum) step size, the global step adjusts to match"""
+    base_config["time"]["step_size"] = 1
+    listener = Listener("listener")
+    sim = SimulationContext(
+        base_config,
+        [StepModifier("step_modifier", step_modifier_even, step_modifier_odd), listener],
+    )
+    quantized_step = math.floor(max([step_modifier_even, 1]))
```
rmudambi, line 181: This isn't the right name for this variable imo, since the quantized value is the minimum possible step size (i.e. 1). Better might be `expected_step_size`?

tests/framework/test_time.py
rmudambi: +1

rmudambi (approved):

tests/framework/test_time.py
patricktnast: This usage is common in physics--to "quantize" some continuous thing is to discretize it in terms of the quantum, so the "quantized" value is not the quantum itself, but the discretized representation of the thing in terms of the quantum. That said, I'm not really picky about the language and arguably it's better to avoid the somewhat inaccessible analogy to quantum mechanics in the first place.

tests/framework/test_time.py
patricktnast: I managed to refactor and clean this up quite a bit, I think

tests/framework/test_time.py
patricktnast: Yeah, that's the intent behind this test--it shouldn't ever "naturally" be the case that we end up with an empty population, but I wanted to be sure that even in this case, we just take a minimum step size.

PR #373, opened 2023-11-10
Make Variable Clocks Backwards Compatible

patricktnast:
### Description

- *Category*: 
feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4665

### Changes and notes
 
--Added default step size, which defaults to "minimum step size" if not set. This is so that we can have, say, a minimum varied step size of 1 day, but "most" people get 16 days, for example.
I don't like the name at all; the quandary is that `step_size` in the default configuration is (and probably should be) the minimum step size, but we also need a default step size, which is different. Perhaps something like "default_varied_step_size"?? or something?? or maybe it is better after all to make `step_size` the _default_ size, and optionally specify a separate minimum step size in the configuration.

--Made it so that if no components modify the step size pipeline, revert to a global clock.

--did some light refactoring to keep everything sensible

### Testing

Added tests to make sure we can accomodate either static or variable time stepping

### DIFF
--- a/src/vivarium/framework/time.py
+++ b/src/vivarium/framework/time.py
@@ -191,6 +191,7 @@ class SimpleClock(SimulationClock):
             "start": 0,
             "end": 100,
             "step_size": 1,
+            "default_step_size": None,
         }
     }
 
--- a/src/vivarium/framework/time.py
+++ b/src/vivarium/framework/time.py
@@ -64,7 +64,7 @@ class SimulationClock(Manager):
 
     @property
     def default_step_size(self) -> Timedelta:
-        """The minimum step size."""
+        """The default varied step size."""
         if not self._default_step_size:
             raise ValueError("No default step size provided")
         return self._default_step_size
--- a/src/vivarium/framework/time.py
+++ b/src/vivarium/framework/time.py
@@ -21,9 +21,11 @@ if TYPE_CHECKING:
     from vivarium.framework.engine import Builder
     from vivarium.framework.population.population_view import PopulationView
     from vivarium.framework.event import Event
+    from vivarium.framework.population import SimulantData
 
 from vivarium.framework.values import list_combiner
 from vivarium.manager import Manager
+
 
 Time = Union[pd.Timestamp, datetime, Number]
 Timedelta = Union[pd.Timedelta, timedelta, Number]
@@ -65,9 +67,9 @@ class SimulationClock(Manager):
     @property
     def default_step_size(self) -> Timedelta:
         """The default varied step size."""
-        if not self._default_step_size:
+        if not self._standard_step_size:
             raise ValueError("No default step size provided")
-        return self._default_step_size
+        return self._standard_step_size
 
     @property
     def step_size(self) -> Timedelta:
@@ -85,12 +87,12 @@ class SimulationClock(Manager):
         self._clock_time: Time = None
         self._stop_time: Time = None
         self._minimum_step_size: Timedelta = None
-        self._default_step_size: Timedelta = None
+        self._standard_step_size: Timedelta = None
         self._clock_step_size: Timedelta = None
-        self.individual_clocks: PopulationView = None
+        self._individual_clocks: PopulationView = None
 
     def setup(self, builder: "Builder"):
-        self.step_size_pipeline = builder.value.register_value_producer(
+        self._step_size_pipeline = builder.value.register_value_producer(
             "simulant_step_size",
             source=lambda idx: [pd.Series(np.nan, index=idx).astype("timedelta64[ns]")],
             preferred_combiner=list_combiner,
@@ -100,17 +102,17 @@ class SimulationClock(Manager):
             self.on_initialize_simulants, creates_columns=self.columns_created
         )
         builder.event.register_listener("post_setup", self.on_post_setup)
-        self.individual_clocks = builder.population.get_view(columns=self.columns_created)
+        self._individual_clocks = builder.population.get_view(columns=self.columns_created)
 
-    def on_post_setup(self, event: "Event"):
-        if not self.step_size_pipeline.mutators:
+    def on_post_setup(self, event: "Event") -> None:
+        if not self._step_size_pipeline.mutators:
             ## No components modify the step size, so we use the default
             ## and remove the population view
-            self.individual_clocks = None
+            self._individual_clocks = None
 
-    def on_initialize_simulants(self, pop_data):
+    def on_initialize_simulants(self, pop_data: "SimulantData") -> None:
         """Sets the next_event_time and step_size columns for each simulant"""
-        if self.individual_clocks:
+        if self._individual_clocks:
             clocks_to_initialize = pd.DataFrame(
                 {
                     "next_event_time": [self.event_time] * len(pop_data.index),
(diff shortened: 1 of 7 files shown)


src/vivarium/framework/time.py
```diff
@@ -122,20 +139,23 @@ def step_backward(self) -> None:
...
-        if not pop_to_update.empty:
-            pop_to_update["step_size"] = self.step_size_pipeline(pop_to_update.index)
-            pop_to_update["next_event_time"] = self.time + pop_to_update["step_size"]
-            self.population_view.update(pop_to_update)
-        self._clock_step_size = self.simulant_next_event_times(index).min() - self.time
-
-    def get_active_population(self, index: pd.Index, time: Time):
+        if self.individual_clocks:
+            update_index = self.get_active_population(index, self.time)
+            pop_to_update = self.individual_clocks.get(update_index)
```
rmudambi, line 144: Nit: `pop_to_update` as a variable name seems like an artifact of when `self.individualized_clocks` was called `self.population_view`.

src/vivarium/framework/time.py
```diff
@@ -77,43 +85,52 @@ def __init__(self):
         self._clock_time: Time = None
         self._stop_time: Time = None
         self._minimum_step_size: Timedelta = None
+        self._default_step_size: Timedelta = None
         self._clock_step_size: Timedelta = None
-        self.population_view: PopulationView = None
+        self.individual_clocks: PopulationView = None
```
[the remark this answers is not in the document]
patricktnast: done

src/vivarium/framework/time.py
```diff
@@ -77,43 +85,52 @@ def __init__(self):
...
 
     def setup(self, builder: "Builder"):
         self.step_size_pipeline = builder.value.register_value_producer(
             "simulant_step_size",
-            source=lambda idx: [pd.Series(self.minimum_step_size, index=idx)],
+            source=lambda idx: [pd.Series(np.nan, index=idx).astype("timedelta64[ns]")],
```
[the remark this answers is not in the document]
patricktnast: done

src/vivarium/framework/time.py
```diff
@@ -77,43 +85,52 @@ def __init__(self):
...
+        builder.event.register_listener("post_setup", self.on_post_setup)
+        self.individual_clocks = builder.population.get_view(columns=self.columns_created)
+
+    def on_post_setup(self, event: "Event"):
+        if not self.step_size_pipeline.mutators:
+            ## No components modify the step size, so we use the default
+            ## and remove the population view
+            self.individual_clocks = None
 
     def on_initialize_simulants(self, pop_data):
```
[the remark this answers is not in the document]
patricktnast: done

src/vivarium/framework/time.py
patricktnast: renamed to get_active_simulants and clocks_to_update

src/vivarium/framework/time.py
```diff
@@ -239,8 +269,8 @@ def step_size(self) -> Callable[[], Timedelta]:
 
     def simulant_next_event_times(self) -> Callable[[pd.Index], pd.Series]:
         """Gets a callable that returns the next event times for simulants."""
-        return lambda: self._manager.simulant_next_event_times
+        return self._manager.simulant_next_event_times
 
     def simulant_step_sizes(self) -> Callable[[pd.Index], pd.Series]:
         """Gets a callable that returns the simulant step sizes."""
-        return lambda: self._manager.simulant_step_sizes
+        return self._manager.simulant_step_sizes
```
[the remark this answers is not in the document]
patricktnast: filed

src/vivarium/framework/time.py
```diff
@@ -65,9 +66,9 @@ def minimum_step_size(self) -> Timedelta:
     @property
     def default_step_size(self) -> Timedelta:
```
rmudambi, line 67: shouldn't this `standard_step_size`?

src/vivarium/framework/time.py
```diff
@@ -139,18 +140,22 @@ def step_backward(self) -> None:
...
-        if self.individual_clocks:
-            update_index = self.get_active_population(index, self.time)
-            pop_to_update = self.individual_clocks.get(update_index)
-            if not pop_to_update.empty:
-                pop_to_update["step_size"] = self.step_size_pipeline(update_index)
-                pop_to_update["next_event_time"] = self.time + pop_to_update["step_size"]
-                self.individual_clocks.update(pop_to_update)
+        if self._individual_clocks:
+            update_index = self.get_active_simulants(index, self.time)
+            simulant_clocks_to_update = self._individual_clocks.get(update_index)
```
rmudambi, line 145: Nit: My OCD would drive me to come up with a variable name that would enable these to stay on one line, but it's obviously not necessary.

tests/framework/test_time.py
```diff
@@ -246,43 +283,106 @@ def test_uneven_steps(SimulationContext, base_config):
...
+        assert sample_pipeline.index.equals(get_pop_by_parity(sim, group).index)
+        assert np.all(sample_pipeline == pipeline_by_parity(sim, step_modifiers, group))
+
+
+def test_partial_modification(SimulationContext, base_config):
+    """Test that if we have one modifier that doesn't apply to all simulants,
+    we choose the default value for unmodified simulants.
+    """
+
+    base_config["time"]["step_size"] = 1
```
rmudambi, line 307: In this case we don't set "standard_step_size" so fall back to the "step_size" configuration. Do we have a test where we actually set the "standard_step_size"?

tests/framework/test_time.py
```diff
@@ -246,43 +283,106 @@ def test_uneven_steps(SimulationContext, base_config):
...
         sample_pipeline = step_modifier_component.ts_pipeline_value
-        assert sample_pipeline.index.equals(test[group])
-        assert np.all(sample_pipeline == pipeline_values[group])
+        assert sample_pipeline.index.equals(get_pop_by_parity(sim, group).index)
+        assert np.all(sample_pipeline == pipeline_by_parity(sim, step_modifiers, group))
+
+
+def test_partial_modification(SimulationContext, base_config):
+    """Test that if we have one modifier that doesn't apply to all simulants,
+    we choose the default value for unmodified simulants.
```
rmudambi, line 304:
```suggestion
    we choose the standard value for unmodified simulants.
```

tests/framework/test_time.py
```diff
@@ -246,43 +283,106 @@ def test_uneven_steps(SimulationContext, base_config):
...
+    to reach all simulants' next event times in the fewest steps.
+    """
+
+    base_config["time"]["step_size"] = 1
+    listener = Listener("listener")
+    step_modifiers = {"evens": 3, "odds": 7}
+    step_modifier_A = StepModifier(
+        "step_modifier_A", step_modifiers["evens"], modified_simulants="evens"
+    )
+    step_modifier_B = StepModifier(
```
rmudambi, line 351: These two modifiers are going to modify disjoint sets of simulants, so we have no simulants who have a modifier from both of them. Is that what we want? Shouldn't we test that we get the expected (minimum) value if two different modifiers modify the same simulant?

rmudambi (approved):

tests/framework/test_time.py
patricktnast: This does get tested in the test of the step size post-processor, though one might argue that we don't quite test the full integration. My concern in including this and the partial modification test is just to ensure that we are doing proper handling of NaNs when we have modifiers that don't return the full population index.

stevebachmeier (approved):

PR #374, opened 2023-11-15
Add Time Interface Step Size Modifier

patricktnast:
### Description

- *Category*: 
refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4702

### Changes and notes
 
Replace the builder.value.register_value_modifier for step size modification with a time interface method (just wrapping the value method) to highlight the special status of this particular pipeline.

I also adjusted the post processor test, to rely less on the values manager.

### Testing

modified test coverage still passes.

### DIFF
--- a/src/vivarium/framework/time.py
+++ b/src/vivarium/framework/time.py
@@ -11,6 +11,7 @@ For more information about time in the simulation, see the associated
 
 """
 from datetime import datetime, timedelta
+from functools import partial
 from numbers import Number
 from typing import TYPE_CHECKING, Callable, List, Union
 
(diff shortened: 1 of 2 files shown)


src/vivarium/framework/time.py
```diff
@@ -281,3 +286,31 @@ def simulant_next_event_times(self) -> Callable[[pd.Index], pd.Series]:
...
+            A list of the state table columns that already need to be present
+            and populated in the state table before the modifier
+            is called.
+        requires_values
+            A list of the value pipelines that need to be properly sourced
+            before the  modifier is called.
+        requires_streams
+            A list of the randomness streams that need to be properly sourced
+            before the modifier is called."""
+        return self._manager.register_step_modifier(
```
stevebachmeier, line 314: Just adding here what @hussain-jafari recommended: we should warn or maybe even error if the user tries to use `register_value_modifier("simulant_step_size", ...)` directly.

My opinion is to make a ticket for that since it seemed like we didn't have any great ideas on how to do that offhand.

stevebachmeier (approved):

src/vivarium/framework/time.py
albrja: + 1

albrja (approved):

src/vivarium/framework/time.py
patricktnast: Filed at [MIC-4710](https://jira.ihme.washington.edu/browse/MIC-4710)

PR #375, opened 2023-11-23
Update code in docs to match current implementation

aflaxman:
This was updated in some places, but we missed a spot.


albrja (approved):

patricktnast (approved):

rmudambi (commented):
When merging to main we must always have a changelog update, since a merge to main is always accompanied by a release.

PR #377, opened 2023-11-29
Integrate MNCH implementation learnings

patricktnast:
### Description

- *Category*: 
misc
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4719

### Changes and notes
 
1. Add some checks for empty indices
2. Ensure that rescale pipeline doesn't blow up
3. Figure out how to properly deal with Tracked / Untracked / Dead/ Alive simulants. Including untracked or dead simulants in the step_forward() call will cause a lot of erroneous timesteps

(3) has a basically two options. Either we sort of commit to not including tracked simulants in event indexes, in which case we basically lose the ability to retrack simulants (to the extent that that matters)
Or, we propagate untracked simulants to events and figure out an appropriate (separate) way to deal with their next event times. 
I opted for the former for the time being, under the assumption that we'll want to revamp the way tracked/untracked works at some point in the near-ish future anyway. 
### Testing

Combo of tested in MNCH and new test coverage added

### DIFF
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -238,7 +238,7 @@ class SimulationContext:
         for event in self.time_step_events:
             self._lifecycle.set_state(event)
             pop_to_update = self._clock.get_active_simulants(
-                self._population.get_population(False).index, self._clock.event_time
+                self._population.get_population(True).index, self._clock.event_time
             )
             self.time_step_emitters[event](pop_to_update)
         self._clock.step_forward(self._population.get_population(False).index)
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -237,8 +237,7 @@ class SimulationContext:
         self._logger.debug(self._clock.time)
         for event in self.time_step_events:
             pop_to_update = self._clock.get_active_simulants(
-                self._population.get_population(False).index,
-                self._clock.event_time
+                self._population.get_population(False).index, self._clock.event_time
             )
             self._logger.debug("Event: %s" % event)
             self._logger.debug("Updating population: %s" % len(pop_to_update))
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -235,12 +235,13 @@ class SimulationContext:
 
     def step(self) -> None:
         self._logger.debug(self._clock.time)
-        for event in self.time_step_events:
-            self._lifecycle.set_state(event)
-            pop_to_update = self._clock.get_active_simulants(
+        pop_to_update = self._clock.get_active_simulants(
                 self._population.get_population(True).index,
                 self._clock.event_time,
             )
+        self._logger.debug("Updating population: %s" % len(pop_to_update))
+        for event in self.time_step_events:
+            self._lifecycle.set_state(event)
             self.time_step_emitters[event](pop_to_update)
         self._clock.step_forward(self._population.get_population(False).index)
 
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -242,7 +242,7 @@ class SimulationContext:
                 self._clock.event_time,
             )
             self.time_step_emitters[event](pop_to_update)
-        self._clock.step_forward(self._population.get_population(True).index)
+        self._clock.step_forward(self._population.get_population(False).index)
 
     def run(self) -> None:
         while self._clock.time < self._clock.stop_time:
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -236,11 +236,13 @@ class SimulationContext:
     def step(self) -> None:
         self._logger.debug(self._clock.time)
         for event in self.time_step_events:
+            self._logger.debug(f"Event: {event}")
             self._lifecycle.set_state(event)
             pop_to_update = self._clock.get_active_simulants(
                 self.get_population().index,
                 self._clock.event_time,
             )
+            self._logger.debug(f"Updating: {len(pop_to_update)}")
             self.time_step_emitters[event](pop_to_update)
         self._clock.step_forward(self.get_population().index)
 
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -231,17 +231,18 @@ class SimulationContext:
         self._clock.step_backward()
         population_size = pop_params.population_size
         self.simulant_creator(population_size, {"sim_state": "setup"})
-        self._clock.step_forward(self._population.get_population(True).index)
+        self._clock.step_forward(self.get_population().index)
 
     def step(self) -> None:
         self._logger.debug(self._clock.time)
         for event in self.time_step_events:
             self._lifecycle.set_state(event)
             pop_to_update = self._clock.get_active_simulants(
-                self._population.get_population(True).index, self._clock.event_time
+                self.get_population().index, 
+                self._clock.event_time,
             )
             self.time_step_emitters[event](pop_to_update)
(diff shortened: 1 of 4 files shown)


src/vivarium/framework/values.py
```diff
@@ -111,10 +111,15 @@ def rescale_post_processor(value: NumberLike, manager: "ValuesManager") -> Numbe
 
     """
     if hasattr(value, "index"):
-        time_step = manager.simulant_step_sizes(value.index).dt
+        return value.mul(
```
patricktnast, line 114: The mortality pipeline is a dataframe, and unfortunately, the default behavior of `mul` is to multiply  dataframe x series across the columns, not the index.

src/vivarium/framework/time.py
```diff
@@ -145,11 +145,14 @@ def step_backward(self) -> None:
...
         """Advances the clock by the current step size, and updates aligned simulant clocks."""
         self._clock_time += self.step_size
-        if self._individual_clocks:
+        if self._individual_clocks and index.any():
             update_index = self.get_active_simulants(index, self.time)
             clocks_to_update = self._individual_clocks.get(update_index)
             if not clocks_to_update.empty:
                 clocks_to_update["step_size"] = self._step_size_pipeline(update_index)
+                clocks_to_update.loc[
+                    ~clocks_to_update["tracked"], "step_size"
```
patricktnast, line 154: the point here is to preserve the existing behavior of propagating untracked simulants to events (in case, for example, we want to *re-track* them later on), while excluding them from the logic that determines what the next time to update a simulant is.

src/vivarium/framework/engine.py
```diff
@@ -231,26 +231,26 @@ def initialize_simulants(self) -> None:
         self._clock.step_backward()
         population_size = pop_params.population_size
         self.simulant_creator(population_size, {"sim_state": "setup"})
-        self._clock.step_forward(self._population.get_population(True).index)
+        self._clock.step_forward(self.get_population().index)
```
patricktnast, line 234: I did a refactor here, since sim has a `get_population()` method, and it seemed weird not to use it.

src/vivarium/framework/time.py
```diff
@@ -145,7 +145,7 @@ def step_backward(self) -> None:
     def step_forward(self, index: pd.Index) -> None:
         """Advances the clock by the current step size, and updates aligned simulant clocks."""
         self._clock_time += self.step_size
-        if self._individual_clocks:
+        if self._individual_clocks and index.any():
```
patricktnast, line 148: the .min() can be problematic with null index (helpfully discovered on first timestep of MNCH child)

tests/framework/test_time.py
```diff
@@ -408,6 +433,47 @@ def test_multiple_modifiers(SimulationContext, base_config):
         )
 
 
+def test_untracked_simulants(SimulationContext, base_config):
```
patricktnast, line 436: This addition is the main component of the tracked/untracked issue. Explained in more detail in the description, this checks that we don't include untracked simulants in the event indices and step size calculations, the default behavior with current functionality.

src/vivarium/framework/engine.py
stevebachmeier: is the `get_population` method something you added somewhere along the way as part of this variable time clock work? If not, are you very sure you understand why this wasn't already using it?

src/vivarium/framework/time.py
```diff
@@ -261,7 +261,7 @@ def setup(self, builder):
             if time.standard_step_size
             else self._minimum_step_size
         )
-        self._clock_step_size = self._standard_step_size
+        self._clock_step_size = self._minimum_step_size
```
stevebachmeier, line 264: This seems like major change in logic - was this just a proper bug?

src/vivarium/framework/values.py
```diff
@@ -111,10 +111,15 @@ def rescale_post_processor(value: NumberLike, manager: "ValuesManager") -> Numbe
...
+            manager.simulant_step_sizes(value.index)
+            .astype("timedelta64[ns]")
+            .dt.total_seconds()
+            / (60 * 60 * 24 * 365.0),
```
stevebachmeier, line 118: We're inconsistent with 365 vs 365.25 days per year - is this appropriate?

stevebachmeier, line 118: Since you convert seconds to years, don't you need the `from_yearly` function?

src/vivarium/framework/values.py
```diff
@@ -111,10 +111,15 @@ def rescale_post_processor(value: NumberLike, manager: "ValuesManager") -> Numbe
 
     """
     if hasattr(value, "index"):
-        time_step = manager.simulant_step_sizes(value.index).dt
+        return value.mul(
+            manager.simulant_step_sizes(value.index)
+            .astype("timedelta64[ns]")
+            .dt.total_seconds()
```
stevebachmeier, line 117: isn't there a more direct `dt` method you can use instead of seconds and then have to convert?

tests/framework/test_time.py
```diff
@@ -209,7 +234,7 @@ def test_empty_active_pop(SimulationContext, base_config, components):
     listener = [c for c in components if hasattr(c, "args") and "listener" in c.args][0]
     sim.setup()
     sim.initialize_simulants()
-    full_pop_index = sim.get_population().index
+    pop_index = full_pop_index(sim)
```
stevebachmeier, line 237: why the rename from `full_pop_index` to `pop_index`? Regardless, it seems like the helper function should follow suit. 

nit: rename the helper function to `get_...()`

src/vivarium/framework/values.py
patricktnast: I agree that this is problematic, though this is an inconsistency that I have simply preserved from `from_yearly`. I think it probably makes sense to change both that and this to 365.25

src/vivarium/framework/values.py
patricktnast: I copied this from `from_yearly` but yeah i think you could just do `step_size / timedelta(days=365)` and that would work the same.

tests/framework/test_time.py
```diff
@@ -408,6 +433,47 @@ def test_multiple_modifiers(SimulationContext, base_config):
         )
 
 
+def test_untracked_simulants(SimulationContext, base_config):
+    """Ensure that we exclude untracked simulants from emitted event indices and
+    exclude them from affecting the next event time."""
```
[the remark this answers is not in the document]
patricktnast: The current implementation essentially precludes untracked simulants from getting tracked again later. I mention some context for keeping it this way in the ticket description

src/vivarium/framework/time.py
patricktnast: In practice, the effect of this bit of code is just to determine what the duration of the very first timestep should be--_a priori_ either would work OK, but having the first step be a minimum step size is a bit more conservative and particularly better in the situation we see in MNCH where simulants have the minimum step size for the first 28 days of the simulation.

src/vivarium/framework/engine.py
patricktnast: It is fairly old code. What seems to have happened is the original method had some hard-coded lifecycle restrictions, which were superceded by the formalization of lifecycle restrictions and so removed.
https://github.com/ihmeuw/vivarium/pull/107/files
https://github.com/ihmeuw/vivarium/pull/109/commits/62311cc97469b6398d4c60f050a9dca404c6477e
Since `self.get_population()` is now little more than a wrapper for `self._population.get_population(True)` I didn't see a strong reason not to use it.

rmudambi (approved):

stevebachmeier (approved):

PR #379, opened 2023-12-08
Document how to remove components from an interactive simulation

zmbc:
### Description
- *Category*: documentation
- *JIRA issue*: None

### Changes and notes

This is a common use case, and has previously been achieved by researchers performing an editable install and then editing the YAML file directly, which is brittle.

### Testing

Ran doctests.

### DIFF
--- a/docs/source/tutorials/running_a_simulation/interactive.rst
+++ b/docs/source/tutorials/running_a_simulation/interactive.rst
@@ -97,16 +97,6 @@ distributed with ``vivarium``.
    p = get_model_specification_path()
    sim = InteractiveContext(p)
 
-If you are working with a real model, you won't have this convenience function.
-Typically, our models are pip-installable and laid out such that their model specification can
-be found like so:
-
-.. code-block:: python
-
-    import my_vivarium_model
-    p = my_vivarium_model.__file__.replace('__init__.py', 'model_specifications/model_spec.yaml')
-    sim = InteractiveContext(p)
-
 The ``sim`` object produced here is all set up and ready to run if you want
 to jump directly to the :ref:`running the simulation <interactive_run>`
 section.
--- a/docs/source/tutorials/running_a_simulation/interactive.rst
+++ b/docs/source/tutorials/running_a_simulation/interactive.rst
@@ -348,9 +348,7 @@ Removing components
 Currently, there isn't a way to use the above approach to *remove* components from
 an existing model (without using private attributes).
 
-This is a very common use case, because often when we run an interactive simulation,
-we don't need any of the simulation's observers.
-However, removing components should be done with care: if you remove a component without
+Removing components should be done with care: if you remove a component without
 removing other components that depend on it, the simulation will break!
 
 You can remove components by creating a model specification and editing it *before* creating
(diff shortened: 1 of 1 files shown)


docs/source/tutorials/running_a_simulation/interactive.rst
```diff
@@ -265,7 +280,12 @@ population size to be smaller so the simulation takes less time to run.
 
 .. code-block:: python
 
-   sim.configuration.update({'population': {'population_size': 1_000}})
+    # Setting attributes ensures you are updating existing keys, rather than
+    # creating new ones
+    sim.configuration.population.population_size = 1_000
```
aflaxman, line 285: does this still work with the latest config tree?

aflaxman (approved):
We should also brainstorm additional affordances that we might one day add to vivarium to streamline these operations

docs/source/tutorials/running_a_simulation/interactive.rst
zmbc: I know it runs, because of the doctest below. However, I didn't test that it actually worked in the latest version of Vivarium. I see no concerning changes to [the relevant code](https://github.com/ihmeuw/vivarium/blob/710b7344661257a898cbaf80b20cee39569bfffe/src/vivarium/config_tree.py#L514-L521), though.

rmudambi (changes requested):

docs/source/tutorials/running_a_simulation/interactive.rst
zmbc: Ah good point -- I guess I wasn't very clear here. This is intended to show two options, where you would only do one of them. I could remove the `update` option, or comment it out.

docs/source/tutorials/running_a_simulation/interactive.rst
```diff
@@ -97,6 +97,16 @@ distributed with ``vivarium``.
    p = get_model_specification_path()
    sim = InteractiveContext(p)
 
+If you are working with a real model, you won't have this convenience function.
+Typically, our models are pip-installable and laid out such that their model specification can
```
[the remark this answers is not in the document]
zmbc: Yep agreed, I'll move more of this over to vivarium_research!

docs/source/tutorials/running_a_simulation/interactive.rst
```diff
@@ -332,6 +342,39 @@ into our disease model. We could do the following.
...
 to immediately see their effects.
 
+Removing components
+~~~~~~~~~~~~~~~~~~~
+
+Currently, there isn't a way to use the above approach to *remove* components from
+an existing model (without using private attributes).
+
+This is a very common use case, because often when we run an interactive simulation,
+we don't need any of the simulation's observers.
```
rmudambi, line 352: I think you should remove this reference to "observers" since that is not a meaningful concept at the vivarium level. Observation exists in vivarium, but the idea of specific components for observation is only implemented in VPH.

zmbc: Ah interesting, I didn't know that. In that case, this sentence is probably a better fit for `vivarium_research` than `vivarium`.

rmudambi: I just meant this sentence.

rmudambi (approved):

zmbc:
@rmudambi Thanks for your review! If this doesn't need any more finishing touches (no changelog necessary for docs changes, right?) feel free to merge it. I cannot, since I don't have write permissions to this repository.

PR #381, opened 2023-12-18
Variable Clocks Documentation

patricktnast:
### Description

- *Category*: 
documentation
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4771

### Changes and notes
 
To release the individualized clocks we need a bit more information about what it is and how it works (but just a little bit for now, since quite a bit will change)

I noticed that there wasn't really much information about the clock in the first place, so I gave it a first cut. Happy to take suggestions or correct errors.
### Testing

N/A

### DIFF
--- a/docs/source/concepts/time.rst
+++ b/docs/source/concepts/time.rst
@@ -14,7 +14,7 @@ Outline
  - The Simulation Clock
  - Start, stop, and step size.
  - Fundamental assumptions about discrete time simulation.
- - builder.time.clock() vs. event.time
+ - clock time vs. event time
  - Individual Clocks
 
 The Simulation Clock
@@ -25,8 +25,8 @@ a mechanism to advance the simulation time by some duration (the *step size*), a
 the simulation is complete via a configured *end time*. The simplest
 implementation of a Clock is the :class:`SimpleClock <vivarium.framework.time.SimpleClock>` object, which is little more
 than an integer counter that is incremented by a fixed step size until it reaches the
-end time. Public health scenarios rely on data that are often tied to particular years or rates, where the 
-desired step size is not necessarily known in advance. Therefore, more commonly used is the :class:`DateTimeClock <vivarium.framework.time.DateTimeClock>`,
+end time. Modeling real-world events is often dependent on data that are tied to particular years or rates, where the 
+desired step size is not necessarily known in advance. Therefore, it is common to use the :class:`DateTimeClock <vivarium.framework.time.DateTimeClock>`,
 which uses datetime-like objects (specifically :class:`~pandas.Timestamp` and :class:`~pandas.Timedelta`) as the temporal units. The DateTimeClock
 can more easily facilitate the conversion rates to particular increments of time.
 
@@ -36,7 +36,7 @@ Discrete time simulations assume that all changes to a simulant's state vector h
 end of the time step, that is, the current clock time *plus* the step size. :mod:`vivarium` explicates this important distinction 
 and labels this quantity the *event time*. `Events <events_concept>` that correspond to (potential) state changes are mediated through the
 :class:`Event Manager <vivarium.framework.event.EventManager>`, which propagates events to :ref:`components <components_concept>` subscribed to them during particuar phases of the simulation lifecycle.
-The Event Manager uses the event time when calculating time-related outcomes, for example, age or year -dependent rates of morbidity and mortality.
+The Event Manager uses the event time when calculating time-related outcomes, for example, age- or year-dependent rates of morbidity and mortality.
 
 Time Interface
 --------------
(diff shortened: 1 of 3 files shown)


docs/source/concepts/time.rst
```diff
@@ -11,7 +11,52 @@ Thinking about Time in the Simulation
...
+
+The Simulation Clock
+--------------------
+The :class:`SimulationClock <vivarium.framework.time.SimulationClock>` plugin manages the progression of time throughout the simulation. 
+Fundamentally, that means it keeps track of the current time (beginning at the *start time*), provides
+a mechanism to advance the simulation time by some duration (the *step size*), and determines when 
+the simulation is complete via a configured *end time*. The simplest
+implementation of a Clock is the :class:`SimpleClock <vivarium.framework.time.SimpleClock>` object, which is little more
+than an integer counter that is incremented by a fixed step size until it reaches the
+end time. Public health scenarios rely on data that are often tied to particular years or rates, where the
```
rmudambi, line 28: We should probably not specifically call out the public health use case. Maybe better just to say something like "when modeling events in the real world it is often useful to use the DateTimeClock which ties simulation time to real dates".

docs/source/concepts/time.rst
```diff
@@ -11,7 +11,52 @@ Thinking about Time in the Simulation
...
+simulant at a different rate depending on the current state of the disease, for example, updating a simulant's state twice as frequently
+when infected than when in remission. The basic method is to give each simulant its own distinct clock time and step size instead of one global clock.
+
+Currently, the simulation still incorporates a global clock, which determines the start, end, and minimal step size of the simulation. The minimum step
+in term determines the maximum number of iterations that can occur in the simulation and serves to quantize the global step size. In each iteration of the simulation,
+the global clock is advanced to the earliest time in which a simulant must be updated, skipping several "minimum" timesteps if necessary. Simulants that are not scheduled to
+be updated in a particular iteration are simply excluded from the relevant events as propagated by the Event Manager.
+
+The Time Interface provides a method to modify a simulant's step size based on some criteria, :func:`builder.time.register_step_modifier() <vivarium.framework.time.TimeInterface.register_step_modifier>`.
```
rmudambi, line 58: It occurs to me that perhaps this method ought to be called `register_step_size_modifier`.

rmudambi (approved):

docs/source/concepts/time.rst
```diff
@@ -11,7 +11,52 @@ Thinking about Time in the Simulation
 
 Outline
 -------
- - The simulation clock
+ - The Simulation Clock
  - Start, stop, and step size.
  - Fundamental assumptions about discrete time simulation.
- - builder.time.clock().clock() vs. event.time
+ - builder.time.clock() vs. event.time
```
stevebachmeier, line 17: Maybe "clock time vs event time"?

docs/source/concepts/time.rst
```diff
@@ -11,7 +11,52 @@ Thinking about Time in the Simulation
...
+which uses datetime-like objects (specifically :class:`~pandas.Timestamp` and :class:`~pandas.Timedelta`) as the temporal units. The DateTimeClock
+can more easily facilitate the conversion rates to particular increments of time.
+
+Event Times
+-----------
+Discrete time simulations assume that all changes to a simulant's state vector happen at the 
+end of the time step, that is, the current clock time *plus* the step size. :mod:`vivarium` explicates this important distinction 
+and labels this quantity the *event time*. `Events <events_concept>` that correspond to (potential) state changes are mediated through the
+:class:`Event Manager <vivarium.framework.event.EventManager>`, which propagates events to :ref:`components <components_concept>` subscribed to them during particuar phases of the simulation lifecycle.
+The Event Manager uses the event time when calculating time-related outcomes, for example, age or year -dependent rates of morbidity and mortality.
```
stevebachmeier, line 39: "...for example, **age- or year-dependent** rates of ..."

docs/source/concepts/time.rst
```diff
@@ -11,7 +11,52 @@ Thinking about Time in the Simulation
...
+
+Time Interface
+--------------
+The Time plugin provides, via the :ref:`Builder <builder_concept>`, an :class:`interface <vivarium.framework.time.TimeInterface>` to access several clock methods that might be needed
+by other managers or components. In particular, components can access the current time and step size (and, implicitly, the event time).
+
+Individual Clocks
+-----------------
+:mod:`vivarium` also allows one to update simulants asynchronously with different frequencies depending on their state information.
+For example, a component that simulates the progression of a disease might need to update the state of each individual
```
stevebachmeier, line 49: This feels awkward with two "for example"s. 

Maybe "A component that simulates the progression of a disease could thus update the state of each individual simulant at a different rate depending on the current state of the disease, e.g. by updating a simulant's state twice as frequently when infected than when in remission."

docs/source/concepts/time.rst
```diff
@@ -11,7 +11,52 @@ Thinking about Time in the Simulation
...
+simulant at a different rate depending on the current state of the disease, for example, updating a simulant's state twice as frequently
+when infected than when in remission. The basic method is to give each simulant its own distinct clock time and step size instead of one global clock.
+
+Currently, the simulation still incorporates a global clock, which determines the start, end, and minimal step size of the simulation. The minimum step
+in term determines the maximum number of iterations that can occur in the simulation and serves to quantize the global step size. In each iteration of the simulation,
```
stevebachmeier, line 54: typo: "in turn"

docs/source/concepts/time.rst
```diff
@@ -11,7 +11,52 @@ Thinking about Time in the Simulation
...
+the global clock is advanced to the earliest time in which a simulant must be updated, skipping several "minimum" timesteps if necessary. Simulants that are not scheduled to
```
stevebachmeier, line 55: This is confusing to me. What do you mean by "skipping several 'minimum' time steps if necessary"? Isnt' the point of the earliest time step so that you don't skip minimum?

stevebachmeier (approved):

docs/source/concepts/time.rst
patricktnast: tried to make this a bit clearer--the point of the minimum timestep is to preclude arbitrarily small or arbitrarily precise step sizes--so the global step size can be larger than the minumum increment, but not smaller.

PR #189, opened 2021-10-01
Continue boids tutorial

zmbc:
Hello! :wave:

I thought the Boids tutorial was really useful for gaining a basic understanding of vivarium, but I saw that it wasn't finished. This is my best attempt to do so. Please feel free to correct any errors!

PR #383, opened 2024-01-01
Continue boids tutorial

zmbc:
### Description
- *Category*: documentation
- *JIRA issue*: none

### Changes and notes

Successor to #189 and #262, which were closed when the target branches were deleted.

- Add swarming behaviors to boids
- Clean up some issues with the existing tutorial
- Add some rough instructions for how to animate the boids

### Testing

Doctests for most of this, ran animation manually in a Jupyter notebook.


src/vivarium/examples/boids/location.py
```diff
@@ -30,22 +31,21 @@ def columns_created(self) -> List[str]:
...
+        # Start randomly distributed, with random velocities
         new_population = pd.DataFrame(
             {
-                "x": self.width * (0.4 + 0.2 * np.random.random(count)),
-                "y": self.height * (0.4 + 0.2 * np.random.random(count)),
-                "vx": -0.5 + np.random.random(count),
-                "vy": -0.5 + np.random.random(count),
+                "x": self.config.width * np.random.random(count),
+                "y": self.config.height * np.random.random(count),
+                "vx": (1 - np.random.random(count) * 2) * self.config.max_velocity,
```
rmudambi, line 47: Nit: I think this is more intuitive:
```suggestion
                "vx": ((2 * np.random.random(count)) - 1) * self.config.max_velocity,
```

src/vivarium/examples/boids/location_swarming.py
rmudambi, line 1: I didn't really review this file very closely. If you think there's something important for me to look at, please call it out.

src/vivarium/examples/boids/neighbors.py
```diff
@@ -15,17 +15,18 @@ class Neighbors(Component):
     ##############
     @property
     def configuration_defaults(self) -> Dict[str, Any]:
-        return {"neighbors": {"radius": 10}}
+        return {"neighbors": {"radius": 60}}
 
     @property
     def columns_required(self) -> Optional[List[str]]:
-        return ["x", "y"]
+        return ["x", "y", "color"]
```
rmudambi, line 22: Where is `color` used?

zmbc: Nowhere 😅 This was left over from a previous version; I'll remove it.

src/vivarium/examples/boids/location.py
zmbc: Agreed!

src/vivarium/examples/boids/location_swarming.py
zmbc: I'd like to get review on `on_time_step` since that is the new piece, and this is tutorial code that should hopefully be pretty readable. Let me know if you'd like to talk through it in real time.

src/vivarium/examples/boids/location_swarming.py
```diff
@@ -0,0 +1,176 @@
...
+
+import numpy as np
+import pandas as pd
+
+from vivarium import Component
+from vivarium.framework.engine import Builder
+from vivarium.framework.population import SimulantData
+
+
+class Location(Component):
```
rmudambi, line 11: It's confusing to me that there are two different components called `Location`

src/vivarium/examples/boids/location_swarming.py
```diff
@@ -0,0 +1,176 @@
...
+        neighbors = self.neighbors(event.index)
+        pop = self.population_view.get(event.index)
+
+        acceleration = pd.DataFrame(0.0, columns=["x", "y"], index=pop.index)
+
+        # Calculate distances between pairs
+        pairs = (
+            pop.join(neighbors.rename("neighbors"))
+            .reset_index()
+            .explode("neighbors")
```
rmudambi, line 70: I wasn't aware of this method

src/vivarium/examples/boids/location_swarming.py
```diff
@@ -0,0 +1,176 @@
...
+                "x": self.config.width * np.random.random(count),
+                "y": self.config.height * np.random.random(count),
+                "vx": ((2 * np.random.random(count)) - 1) * self.config.max_velocity,
+                "vy": ((2 * np.random.random(count)) - 1) * self.config.max_velocity,
+            },
+            index=pop_data.index,
+        )
+        self.population_view.update(new_population)
+
+    def on_time_step(self, event):
```
rmudambi, line 60: I think we need to break this method down with helper methods to improve readability.

src/vivarium/examples/boids/location_swarming.py
zmbc: Yeah, so this is a bit weird -- these two files are actually supposed to represent two _versions_ of the same file. `location.py` is the first version, then (if you were following the tutorial) you would update it to `location_swarming.py`.

One alternative would be to make Swarming a separate component -- is that the direction I should go? Or is this fine? I considered trying to make each _force_ a separate component, but they would need to share acceleration columns, which are currently not in the state table.

src/vivarium/examples/boids/location_swarming.py
zmbc: It's fun!

src/vivarium/examples/boids/location_swarming.py
zmbc: Yep makes sense. See my question about separate components above.

src/vivarium/examples/boids/forces.py
```diff
@@ -0,0 +1,165 @@
...
+    ##############
+    @property
+    def configuration_defaults(self) -> Dict[str, Any]:
+        return {
+            self.__class__.__name__.lower(): {
+                "max_force": 0.03,
+            },
+        }
+
+    columns_required = ["x", "y", "vx", "vy"]
```
rmudambi, line 23: This should be a property?

src/vivarium/examples/boids/neighbors.py
```diff
@@ -13,13 +13,9 @@ class Neighbors(Component):
...
     ##############
-    @property
-    def configuration_defaults(self) -> Dict[str, Any]:
-        return {"neighbors": {"radius": 60}}
+    configuration_defaults = {"neighbors": {"radius": 60}}
 
-    @property
-    def columns_required(self) -> Optional[List[str]]:
-        return ["x", "y"]
+    columns_required = ["x", "y"]
```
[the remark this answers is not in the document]
zmbc: This is a property, right? It's just not using the `@property` _decorator_ since the function was unnecessary. Or am I missing something?

src/vivarium/examples/boids/movement.py
```diff
@@ -0,0 +1,85 @@
...
+from vivarium.framework.population import SimulantData
+
+
+class Movement(Component):
+    ##############
+    # Properties #
+    ##############
+    configuration_defaults = {
+        "movement": {
+            "width": 1000,  # Width of our field
```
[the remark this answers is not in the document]
zmbc: Yeah, that's what they mean. "Grid" is better but still doesn't seem quite right, maybe "field"?

docs/source/tutorials/boids.rst
```diff
@@ -337,31 +443,69 @@ with that information. We need to teach the boids to swarm!
 There are lots of potential swarming behaviors to play around with, all of which
 change the way that boids clump up and follow each other. But since that isn't
 the focus of this tutorial, we'll implement separation, cohesion, and alignment
-behavior identical to what's in `this D3 example <https://web.archive.org/web/20240103000750/https://d3og.com/git-ashish/2ff94f1f6b985e5fd2d4a15e512c4739/>`_.
+behavior identical to what's in `this D3 example <https://web.archive.org/web/20240103000750/https://d3og.com/git-ashish/2ff94f1f6b985e5fd2d4a15e512c4739/>`_,
```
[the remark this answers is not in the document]
zmbc: There is, but I _prefer_ the Internet Archive version, as it won't disappear on us and become a broken link.

src/vivarium/examples/boids/movement.py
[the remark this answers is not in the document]
rmudambi: Field could work. Movement just seems wrong for this.

src/vivarium/examples/boids/neighbors.py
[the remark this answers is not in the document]
rmudambi: Fair enough

docs/source/tutorials/boids.rst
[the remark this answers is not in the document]
zmbc: Actually I'll include both!

src/vivarium/examples/boids/forces.py
```diff
@@ -0,0 +1,168 @@
...
+        return pairs
+
+    def _normalize_and_limit_force(
+        self,
+        force: pd.DataFrame,
+        velocity: pd.DataFrame,
+        max_force: float,
+        max_velocity: float,
+    ):
+        normalization_factor = np.where(
```
patricktnast, line 94: I don't think I totally understand what's going on with this normalization. It looks like it is doing: 
$$\vec{F} = \vec{F} * \frac{v_{max}}{\left| \vec{F} \right|} - \vec{v} = \hat{F} *v_{max} -\vec{v} $$

Which will ensure that the resulting velocity after application will have a magnitude $v_{max}$ (that is, if the max force scaling were not applied), regardless of the calculated magnitude of the force, right? Is that what you want?

patricktnast (commented):
Can you walk me through the force normalization at some point?

src/vivarium/examples/boids/forces.py
zmbc: Note for posterity: this is not super obvious. The direction of the "steering force" depends on the relative magnitudes of the speed and max_speed, but the max_force scaling below should almost always determine the magnitude of the steering force. See https://www.red3d.com/cwr/papers/1999/gdc99steer.pdf for the background.

Overall, we think this is probably right.

patricktnast (approved):

PR #393, opened 2024-04-09
Mic-4973/component-config-lookup-table

albrja:
### Adds functionality in Component to configure and automatically create lookup tables
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4973

### Changes and notes
-adds methods in component to build and create lookup tables from user configurations.

### Testing
All tests pass. Successfully ran simulation with custom configuration

### DIFF
--- a/src/vivarium/component.py
+++ b/src/vivarium/component.py
@@ -74,16 +74,21 @@ class Component(ABC):
 
     @staticmethod
     def build_lookup_table_config(
-        value: str, cont_col: List[str], cat_col: List[str], skip: bool = False, **kwargs
+        value: str,
+        continuous_columns: List[str] = [],
+        categorical_columns: List[str] = [],
+        skip_build: bool = False,
+        key_name: str = None,
+        **kwargs,
     ) -> dict:
         config = {
             "value": value,
-            "data_columns": {
-                "continuous_columns": cont_col,
-                "categorical_columns": cat_col,
-            },
-            "skip": skip,
+            "continuous_columns": continuous_columns,
+            "categorical_columns": categorical_columns,
+            "skip_build": skip_build,
         }
+        if key_name:
+            config["key_name"] = key_name
         config.update(kwargs)
         return config
 
@@ -569,14 +574,21 @@ class Component(ABC):
         }
 
     def create_lookup_tables(self, builder: "Builder") -> None:
+        if (
+            self.name not in builder.configuration
+            or "lookup_tables" not in builder.configuration[self.name]
+        ):
+            return
         lookup_table_config = builder.configuration[self.name].lookup_tables
 
         for table_name, table_config in lookup_table_config.items():
-            if "special" in table_config and table_config.special:
+            if "skip_build" in table_config and table_config.skip_build:
                 continue
+            # TODO: make path to configuration the data key when we align artifact
+            # keys with configuration path
             if table_config.value == "data":
                 table = builder.lookup.build_table(
-                    table_name,
+                    data=builder.data.load(table_config["key_name"]),
                     key_columns=table_config["categorical_columns"],
                     parameter_columns=table_config["continuous_columns"],
                 )
--- a/src/vivarium/component.py
+++ b/src/vivarium/component.py
@@ -77,7 +77,6 @@ class Component(ABC):
         value: str,
         continuous_columns: List[str] = [],
         categorical_columns: List[str] = [],
-        build_lookup_table: bool = True,
         key_name: str = None,
         **kwargs: Dict[str, Any],
     ) -> dict:
@@ -85,7 +84,6 @@ class Component(ABC):
             "value": value,
             "continuous_columns": continuous_columns,
             "categorical_columns": categorical_columns,
-            "build_lookup_table": build_lookup_table,
         }
         if key_name:
             config["key_name"] = key_name
@@ -226,6 +224,13 @@ class Component(ABC):
             additional columns are necessary.
         """
         return None
+
+    @property
+    def standard_lookup_tables(self) -> List[str]:
+        """
+        Returns a list of keys that are used to create standard lookup tables in a component.
+        """
+        return []
 
     @property
     def initialization_requirements(self) -> Dict[str, List[str]]:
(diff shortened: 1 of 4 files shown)


tests/framework/components/mocks.py
```diff
@@ -16,6 +16,9 @@ def __init__(self, *args, name="mock_component_a"):
         self.args = args
         self.builder_used_for_setup = None
 
+    def create_lookup_tables(self, builder):
```
albrja, line 19: Overwriting this method in these two mock components to handle how we mock builder in some tests.

tests/framework/components/test_component.py
```diff
@@ -340,3 +342,151 @@ def test_listeners_are_registered_at_custom_priorities():
...
     assert component.on_simulation_end in set(simulation_end_methods.get(1, []))
+
+
+def test_component_lookup_table_configuration(hdf_file_path):
+    # Tests that lookup tables are created correctly based on their configuration
+
+    favorite_team = pd.DataFrame(
+        {"value": ["team_1", "team_2", "team_3"], "test_column_1": [1, 2, 3]}
+    )
+    favorite_team = favorite_team.set_index("test_column_1")
```
stevebachmeier, line 353: nit: chain the `.set_index()` here and below

tests/framework/components/test_component.py
```diff
@@ -340,3 +342,151 @@ def test_listeners_are_registered_at_custom_priorities():
...
+    assert not component.lookup_tables["favorite_team"].parameter_columns
+    assert component.lookup_tables["favorite_color"].key_columns == ["test_column_2"]
+    assert component.lookup_tables["favorite_color"].parameter_columns == ["test_column_3"]
+    assert (
+        not component.lookup_tables["favorite_number"].key_columns
+        and not component.lookup_tables["favorite_number"].parameter_columns
+    )
+    assert component.lookup_tables["favorite_number"].data == 0.4
+
+    # Component level asswertions for lookup tables
```
stevebachmeier, line 419: nit: "asswertions"

tests/framework/components/test_component.py
```diff
@@ -340,3 +342,151 @@ def test_listeners_are_registered_at_custom_priorities():
...
+            },
+        }
+    )
+    sim.setup()
+
+    # Assertions for specific lookup tables
+    assert component.lookup_tables["favorite_team"].key_columns == ["test_column_1"]
+    assert not component.lookup_tables["favorite_team"].parameter_columns
+    assert component.lookup_tables["favorite_color"].key_columns == ["test_column_2"]
+    assert component.lookup_tables["favorite_color"].parameter_columns == ["test_column_3"]
```
stevebachmeier, line 412: I think you're missing a case where a lookup table built without key column but with parameter cols?

tests/framework/components/test_component.py
```diff
@@ -340,3 +342,151 @@ def test_listeners_are_registered_at_custom_priorities():
...
+    assert (
+        not component.lookup_tables["favorite_number"].key_columns
+        and not component.lookup_tables["favorite_number"].parameter_columns
+    )
+    assert component.lookup_tables["favorite_number"].data == 0.4
+
+    # Component level asswertions for lookup tables
+    assert "baking_time" not in component.lookup_tables.keys()
```
stevebachmeier, line 420: I'm not totally following the logic here, but it seems like the assertion can be strengthened by asserting that `lookup_tables.keys()` returns exactly what you'd expect?

stevebachmeier (changes requested):
Look good aside from some small things!

src/vivarium/component.py
```diff
@@ -72,6 +72,26 @@ class Component(ABC):
     - `on_simulation_end`
     """
 
+    @staticmethod
+    def build_lookup_table_config(
```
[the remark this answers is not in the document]
stevebachmeier: Per slack, it sounds like this will be used by some subclasses. 👍

stevebachmeier (approved):

src/vivarium/component.py
```diff
@@ -551,6 +573,29 @@ def get_initialization_parameters(self) -> Dict[str, Any]:
...
+    def create_lookup_tables(self, builder: "Builder") -> None:
+        if (
+            self.name not in builder.configuration
+            or "lookup_tables" not in builder.configuration[self.name]
+        ):
+            return
+        lookup_table_config = builder.configuration[self.name].lookup_tables
+
+        for table_name, table_config in lookup_table_config.items():
+            if "skip_build" in table_config and table_config.skip_build:
```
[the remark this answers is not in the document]
albrja: If you don't want to create the lookup tables out of the box. This will mainly be used if you have custom code to create specific lookup tables. You will have to manually trigger the creation of these lookup tables in your custom component. Think unmodeled causes in the mortality component.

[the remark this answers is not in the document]
albrja: Yes it defaults to False so they both need to be checked.

tests/framework/components/test_component.py
albrja: This was verifying the skip_build parameter was skipping the baking time but I will add that assertion as well.

rmudambi (commented):
We shouldn't release this until after you've implemented all the VPH components. I'd suggest pointing this PR at an epic/feature branch.

src/vivarium/component.py
```diff
@@ -72,6 +72,26 @@ class Component(ABC):
     - `on_simulation_end`
     """
 
+    @staticmethod
+    def build_lookup_table_config(
+        value: str,
+        continuous_columns: List[str] = [],
+        categorical_columns: List[str] = [],
```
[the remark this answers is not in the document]
albrja: This does work however....  I think it is probably pretty unclear that if you update say continuous columns, you would have to add the default categorical columns to build the expected table.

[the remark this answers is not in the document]
albrja: I wanted users to only have to specific args they needed to in order to create a table but maybe this is confusing.

src/vivarium/component.py
```diff
@@ -72,6 +72,26 @@ class Component(ABC):
...
+        skip_build: bool = False,
```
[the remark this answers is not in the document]
albrja: Well in that sense it is still configurable right? I do think a property would probably be more intuitive as to what is happening vs this arg doing that. We could have Component call the function to create the custom lookup tables like we do with the lifecycle methods so we would just have to add the code in our subclasses.

rmudambi:
I think there may be a better way to structure the configuration. What do you think of this:
```yaml
mortality: 
    all_cause_mortality_rate:
        build_lookup_table: True
        value: "data"
        continuous_columns: ["age"]
        categorical_columns: ["sex", "year"]
```

Then `component.create_lookup_tables()` would look for a key matching the component's name, and then iterate through all keys that have `build_lookup_table == True`.

src/vivarium/component.py
```diff
@@ -551,6 +573,29 @@ def get_initialization_parameters(self) -> Dict[str, Any]:
...
+        lookup_table_config = builder.configuration[self.name].lookup_tables
+
+        for table_name, table_config in lookup_table_config.items():
+            if "skip_build" in table_config and table_config.skip_build:
+                continue
+            # TODO: make path to configuration the data key when we align artifact
+            # keys with configuration path
+            if table_config.value == "data":
+                table = builder.lookup.build_table(
+                    data=builder.data.load(table_config["key_name"]),
```
patricktnast, line 591: Is it possible to get a `KeyError` here? If so, do we want to handle that earlier (e.g. in `build_lookup_table_config`?)

patricktnast (approved):

src/vivarium/component.py
albrja: It depends what you mean. I've set up build_lookup_table_config to have default args to so we do not have KeyErros here. However, we could have an artifact key error and that is tested.

tests/framework/components/test_component.py
```diff
@@ -374,44 +386,43 @@ def test_component_lookup_table_configuration(hdf_file_path):
...
-                        skip_build=True,
-                    ),
-                    "favorite_number": Component.build_lookup_table_config(
-                        value="data",
-                        categorical_columns=[],
-                        continuous_columns=["test_column_3"],
-                        key_name="simulants.favorite_number",
-                    ),
-                },
+                "favorite_team": Component.build_lookup_table_config(
```
rmudambi, line 395: Since you did end up creating a class for this test (`LookupCreator()`), I think you should define the configs in the class itself.

rmudambi (approved):

tests/framework/components/test_component.py
albrja: Ok yea I had to overwrite the property so I will also do that.

PR #396, opened 2024-04-18
Feature/sbachmei/mic 4992 initialize metrics with zeros dataframe

stevebachmeier:
### Description

- *Category*:  feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4992

### Changes and notes
 
This is the first of a handful of PRs to implement a new approach to
results processing. [Design doc here](https://hub.ihme.washington.edu/display/SSE/Results+Processing).

The scope of this is limited to initializing a new ResultsManager.metrics attribute.
Instead of a Counter, we use a dictionary of pd.DataFrames (one per measure)
where each df is a multiindex of all stratified values and has a single "value"
column of 0s.

### Testing

added tests, all pass


src/vivarium/framework/results/manager.py
```diff
@@ -71,36 +70,36 @@ def setup(self, builder: "Builder"):
...
-    def on_post_setup(self, event: Event):
-        # update self._metrics to have all output keys
-        def create_measure_specific_keys(measure: str, stratifications: List[str]) -> None:
-            measure_str = f"MEASURE_{measure}"
-            individual_stratification_strings = [
-                [
-                    f"{stratification.name.upper()}_{category}"
-                    for category in stratification.categories
-                ]
-                for stratification in sorted(
```
hussain-jafari, line 83: I remember there was a reason for this sorting and calling the upper() method on the stratification name but I can't remember why now. I think if we see what we expect in the index of the initialized dataframes we should be go to go, or will catch the reasons for doing this further along in the process.

hussain-jafari (approved):
LGTM! I like this MultiIndexed DataFrame approach

src/vivarium/framework/results/manager.py
stevebachmeier: Yeah I wasn't sure about the sorting. the upper I think was just to make the columns easier to parse later. I agree that we'll find out down the line!

tests/framework/results/helpers.py
```diff
@@ -25,14 +24,47 @@
 FAMILIARS = ["owl", "cat", "gecko", "banana_slug", "unladen_swallow"]
 POWER_LEVELS = [i * 10 for i in range(5, 9)]
 TRACKED_STATUSES = [True, False]
-RECORDS = [
-    (house, familiar, power_level, ts)
-    for house, familiar, power_level, ts in itertools.product(
-        CATEGORIES, FAMILIARS, POWER_LEVELS, TRACKED_STATUSES
-    )
-]
+RECORDS = list(itertools.product(CATEGORIES, FAMILIARS, POWER_LEVELS, TRACKED_STATUSES))
```
stevebachmeier, line 27: Thanks copilot

src/vivarium/framework/results/manager.py
```diff
@@ -71,36 +70,36 @@ def setup(self, builder: "Builder"):
...
-                    create_measure_specific_keys(measure, stratifications)
+                    stratification_values = []
+                    for stratification in [
+                        s
+                        for s in self._results_context.stratifications
+                        if s.name in stratification_names
+                    ]:
+                        stratification_values.append(
+                            (stratification.name, stratification.categories)
+                        )
```
rmudambi, line 89: Nit: I think this is easier to read:
```suggestion
                    observation_stratifications = [
                        stratification
                        for stratification in self._results_context.stratifications
                        if stratification.name in stratification_names
                    ]
                    stratification_values = [
                        (stratification.name, stratification.categories)
                        for stratification in observer_stratifications
                    ]
```

rmudambi: I also think `stratification_values` makes more sense to be a dictionary, rather than a list of tuples.

tests/framework/components/test_manager.py
```diff
@@ -9,7 +9,7 @@
 )
 from vivarium.framework.configuration import build_simulation_configuration
 
-from .mocks import MockComponentA, MockComponentB, MockGenericComponent
+from ...helper_components import MockComponentA, MockComponentB, MockGenericComponent
```
rmudambi, line 12: Why are you importing with `...` rather than an absolute path?

tests/framework/components/test_parser.py
```diff
@@ -10,7 +10,7 @@
 )
 from vivarium.framework.configuration import build_simulation_configuration
 
-from .mocks import MockComponentA, MockComponentB
+from ...helper_components import MockComponentA, MockComponentB
```
rmudambi, line 13: Same question here.

tests/framework/components/test_manager.py
stevebachmeier: I can never get it to work from the tests/ dir (note that Matt was doing this originally as well fwiw)

rmudambi (approved):

src/vivarium/framework/components/parser.py
```diff
@@ -25,11 +25,9 @@
 
 from layered_config_tree import LayeredConfigTree
 
+from vivarium.framework.components.manager import Component, ComponentConfigError
```
rmudambi, line 28: This isn't importing `Component` from the correct place. It should be
```python
from vivarium import Component
from vivarium.framework.components.manager import ComponentConfigError
```
or if that doesn't work

```python
from vivarium.component import Component
from vivarium.framework.components.manager import ComponentConfigError
```

src/vivarium/framework/engine.py
```diff
@@ -28,23 +28,22 @@
 from layered_config_tree import LayeredConfigTree
 
 from vivarium.exceptions import VivariumError
+from vivarium.framework.artifact import ArtifactInterface
+from vivarium.framework.components import ComponentConfigError, ComponentInterface
+from vivarium.framework.components.manager import Component
```
rmudambi, line 33: Same here for Component

src/vivarium/framework/results/context.py
```diff
@@ -101,7 +101,9 @@ def add_observation(
             (name, aggregator_sources, aggregator, additional_keys)
         )
 
-    def gather_results(self, population: pd.DataFrame, event_name: str) -> Dict[str, float]:
+    def gather_results(
+        self, population: pd.DataFrame, event_name: str
+    ) -> Generator[Dict[str, float], None, None]:
```
rmudambi, line 106: What do the Nones here mean?

hussain-jafari (approved):

rmudambi (approved):
Feel free to merge after fixing those two Component imports

src/vivarium/framework/results/context.py
hussain-jafari: @stevebachmeier I'm assuming this is why we saw the "expected 3 arguments but got only 1" at mid-day yesterday? I remember reading this while looking into it: https://stackoverflow.com/questions/42531143/how-to-type-hint-a-generator-in-python-3

stevebachmeier: So I won't pretend to have a good grasp on how generators work, but the return types of a generator are [yield, send, and return]. We're yielding, so that's the one we want. I don't even know what the other two mean at this point, though.

https://stackoverflow.com/questions/42531143/how-to-type-hint-a-generator-in-python-3

PR #398, opened 2024-04-22
Feature/sbachmei/mic 4993 implement metrics updating

stevebachmeier:
### Description

- *Category*:  feature, refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4993

### Changes and notes
 
This replaces the `gather_results` method (previously called `.update()` on
the metrics Counter object with the new pd.Series approach.

### Testing

Added new test and fixed failing; all pass

### DIFF
--- a/tests/framework/results/test_interface.py
+++ b/tests/framework/results/test_interface.py
@@ -173,8 +173,12 @@ def mock__prepare_population(self, event):
     return population
 
 
-def test_interface_full_observation(mocker):
-    """Test the full interface lifecycle of adding an observation and simulate a `collect_metrics` event."""
+@pytest.mark.parametrize(
+    "when",
+    ["time_step__prepare", "time_step", "time_step__cleanup", "collect_metrics"],
+)
+def test_register_observation_when_options(when, mocker):
+    """Test the full interface lifecycle of adding an observation and simulatean event."""
     # Create interface
     mgr = ResultsManager()
     results_interface = ResultsInterface(mgr)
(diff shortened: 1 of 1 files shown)


src/vivarium/framework/results/context.py
```diff
@@ -103,7 +103,7 @@ def add_observation(
 
     def gather_results(
         self, population: pd.DataFrame, event_name: str
-    ) -> Generator[Dict[str, float], None, None]:
+    ) -> Generator[pd.Series, None, None]:
```
stevebachmeier, line 106: pd.DataFrame would have worked but a series just seems more natural than a single-col df.

tests/helper_components.py
```diff
@@ -126,3 +129,169 @@ def on_collect_metrics(self, event: Event) -> None:
 
     def on_simulation_end(self, event: Event) -> None:
         self.simulation_end_called = True
+
+
+class ColumnCreator(Component):
```
stevebachmeier, line 134: These are all direct copy/paste from component.py

tests/framework/components/test_component.py
```diff
@@ -1,179 +1,23 @@
...
 import pytest
 
 from vivarium import Artifact, Component, InteractiveContext
 from vivarium.framework.artifact import ArtifactException
-from vivarium.framework.engine import Builder
-from vivarium.framework.event import Event
-from vivarium.framework.population import SimulantData
-
-
-class ColumnCreator(Component):
```
stevebachmeier, line 13: All of the helper classes were moved into a new helper_components.py module

tests/framework/results/test_context.py
```diff
@@ -385,22 +385,6 @@ def test_gather_results_with_no_stratifications():
     assert len(list(ctx.gather_results(population, event_name))) == 1
 
 
-def test__format_results():
```
stevebachmeier, line 388: No more `ResultsContext._format_results()` method. The primary purpose of that was to convert the stratified/aggregated dataframe into the long key:value pairs that eventually becamse output.hdf columns.

src/vivarium/framework/results/manager.py
```diff
@@ -89,12 +89,11 @@ def on_post_setup(self, _: Event):
...
-                    self._metrics[measure] = pd.DataFrame(
-                        data=0.0,  # Initialize to 0
-                        columns=["value"],
+                    self._metrics[measure] = pd.Series(
+                        data=0.0,
+                        name=measure,
                         index=pd.MultiIndex.from_tuples(
-                            index_values,
-                            names=stratification_values.keys(),
+                            index_values, names=stratification_values.keys()
```
albrja, line 96: Ok higher level question here to make sure I am understanding this correctly - we changed from having a 1 column dataframe for each measure to a series and we store a dict of these series in self._metrics. I assume the reason self._metrics is not a dataframe (wide?) is because me would have different indices for different measures?

stevebachmeier: It really could be either a df or a series. It just feels more natural to be a series where the multi-index are the groups/stratifications and the values are the, well, values. I initially was intending on having a single-col dataframe with that column named "value" but that was just begging to be converted to a series. But it really is the same.

rmudambi (changes requested):

src/vivarium/framework/results/context.py
```diff
@@ -119,14 +119,14 @@ def gather_results(
             else:
                 filtered_pop = population
             if filtered_pop.empty:
-                yield {}
+                yield pd.Series()
```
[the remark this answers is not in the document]
stevebachmeier: Eek, will fix and add a test

tests/framework/results/helpers.py
```diff
@@ -39,30 +41,103 @@
...
+        update = self.population_view.get(pop_data.index)
+        update["previous_house_points"] = update["house_points"]
+        update["previous_quidditch_wins"] = update["quidditch_wins"]
+        # House points are stratified by 'student_house' and 'power_level'.
+        # Let's have each wizard of gryffindor and of level 50 and 80 gain a point
+        update.loc[
+            (update["student_house"] == "gryffindor")
+            & (update["power_level"].isin(["50", "80"])),
+            "house_points",
+        ] += 1
```
stevebachmeier, line 84: Per discussions w/ Rajan: change this to NOT increase (i.e. be the points PER TIME STEP) and then we can just use `sum()` as the aggregator. This is more in line with how we do things

src/vivarium/framework/results/manager.py
```diff
@@ -89,12 +89,11 @@ def on_post_setup(self, _: Event):
                     }
                     # Get the complete cartesian product of stratifications
                     index_values = list(itertools.product(*stratification_values.values()))
-                    self._metrics[measure] = pd.DataFrame(
-                        data=0.0,  # Initialize to 0
-                        columns=["value"],
+                    self._metrics[measure] = pd.Series(
```
stevebachmeier, line 92: From discussions w/ Rajan: Let's go back to DataFrames b/c we probably want those later when we implement a different type of updater

src/vivarium/framework/results/manager.py
stevebachmeier: After all that, I'm actually switching back to DataFrames

tests/framework/results/helpers.py
```diff
@@ -39,30 +41,103 @@
 ##################
 
 
+class Hogwarts(Component):
```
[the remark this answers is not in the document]
stevebachmeier: I don't like the organization of this test suite at all. I tried making it better but I think I just made it worse. The idea was that these are specific to results only and helper components are for all tests, but I think I'll just plot everything into helper_components as you implied.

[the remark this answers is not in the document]
stevebachmeier: I sort of likened it to how there are multiple conftest files floating around depending on the intended scope of the fixtures.

tests/framework/results/test_interface.py
```diff
@@ -173,7 +173,7 @@ def mock__prepare_population(self, event):
     return population
 
 
-def test_integration_full_observation(mocker):
+def test_interface_full_observation(mocker):
```
rmudambi, line 176: It seems like this test would be better as a proper integration test (i.e. with an `InteractiveContext`).

rmudambi (approved):
Do we have a test that the simulation as a whole outputs the expected csvs, or is that out of scope for this PR?

stevebachmeier:
> Do we have a test that the simulation as a whole outputs the expected csvs, or is that out of scope for this PR?

Out of scope - that's the followup PR which is to fix up the actual InteractiveContext.report() method

tests/framework/results/test_interface.py
stevebachmeier: This test is a weird one. What it seems like it's really testing to me is whether or not aggregators get called on the event that they're registered to ("a_measure" is called because it's registered to  "collect_metrics" but "another_measure" is NOT called b/c it's registered to "time_step__prepare"). 

I think there's some utility to this, though it does seem undercooked. 

I'll try and beef it up as well as rename it I think b/c it's not really a "full_observation" test imo

stevebachmeier: @rmudambi What do you think of this updated test? I changed the name to not be misleading. I'm happy to add a proper integration test but it sort of seems unnecessary since in the following PR I have near-e2e tests using `InteractiveContext` which checks the formatting of the output results.

albrja (approved):

PR #403, opened 2024-04-29
Feature/sbachmei/mic 5020 implement report phase

stevebachmeier:
### Description

- *Category*:  refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4968

### Changes and notes
 
Currently the results (previously output.hdf,  now the actual count data) happens
directly in the engine.py module. The "report" phase is set, but a listener is never
registered and it's never emitted. The engine.py module just just loops through
the results and saves them all to csv.

This refactors things a bit to be more in line w/ the rest of vivarium. 
- Each observation is registered with a `report` callable. 
- A "report" listener is registered
- The report phase is emitted and a new `ResultsManager.on_report()`
    method is called which loops through all observations and extracts
    the relevent `report` Callables and calls them

### Testing

tests pass

### DIFF
--- a/src/vivarium/framework/results/context.py
+++ b/src/vivarium/framework/results/context.py
@@ -1,4 +1,5 @@
 from collections import defaultdict
+from pathlib import Path
 from typing import Callable, Generator, List, Optional, Tuple, Union
 
 import pandas as pd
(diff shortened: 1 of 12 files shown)


src/vivarium/framework/results/context.py
```diff
@@ -90,16 +91,16 @@ def add_observation(
         pop_filter: str,
         aggregator_sources: Optional[List[str]],
         aggregator: Callable[[pd.DataFrame], float],
-        additional_stratifications: List[str] = [],
```
stevebachmeier, line 93: I removed defaults for a bunch of these functions that are ultimately wrapped in the ResultsInterface object so that I could tack `report` on the end without having to keep passing the default method through. I also think it's cleaner this way (though it does mean I had to fix some of the unit tests that were calling these inner methods)

rmudambi: See if removing the defaults is really providing value as you can set a default report function, though I don't think there's anything wrong with doing it as you have.

tests/framework/results/test_interface.py
```diff
@@ -8,6 +8,7 @@
 from tests.framework.results.helpers import CATEGORIES as HOUSES
 from tests.framework.results.helpers import FAMILIARS, mock_get_value
 from vivarium.framework.results import ResultsInterface, ResultsManager
+from vivarium.framework.results.reporters import dataframe_to_csv
```
rmudambi, line 11: Is this used?

rmudambi (approved):

PR #405, opened 2024-04-29
add check for missing Stratifications

stevebachmeier:
### Description

- *Category*:  bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5022

### Changes and notes
 Currently there is nothing stopping an Observation from requesting 
stratifications that never actually get registered; it will just not include
that in the actual groups. This PR raises an error in this case.

### Testing

added a test to check it; all tests pass

### DIFF
--- a/src/vivarium/framework/results/manager.py
+++ b/src/vivarium/framework/results/manager.py
@@ -72,26 +72,40 @@ class ResultsManager(Manager):
 
     def on_post_setup(self, _: Event):
         """Initialize self._metrics with 0s DataFrame' for each measure and all stratifications"""
+        registered_stratifications = self._results_context.stratifications
+        registered_stratification_names = set(
+            stratification.name for stratification in registered_stratifications
+        )
+
         all_missing_stratifications = {}
+        unused_stratifications = registered_stratification_names.copy()
+
         for event_name in self._results_context.observations:
             for (
                 _pop_filter,
-                stratification_names,
+                all_requested_stratification_names,
             ), observations in self._results_context.observations[event_name].items():
                 for measure, *_ in observations:
-                    observation_stratifications = [
-                        stratification
-                        for stratification in self._results_context.stratifications
-                        if stratification.name in stratification_names
-                    ]
-                    missing_stratifications = set(stratification_names).difference(
-                        set(obs.name for obs in observation_stratifications)
-                    )
+                    # Batch missing stratifications
+                    missing_stratifications = set(
+                        all_requested_stratification_names
+                    ).difference(registered_stratification_names)
                     if missing_stratifications:
                         all_missing_stratifications[measure] = missing_stratifications
+
+                    # Remove stratifications from the running list of unused stratifications
+                    unused_stratifications = unused_stratifications.difference(
+                        set(all_requested_stratification_names)
+                    )
+                    # Set up the complete index of all used stratifications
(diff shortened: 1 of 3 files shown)


src/vivarium/framework/results/manager.py
```diff
@@ -83,6 +83,14 @@ def on_post_setup(self, _: Event):
                         for stratification in self._results_context.stratifications
                         if stratification.name in stratification_names
                     ]
+                    missing_stratifications = set(stratification_names).difference(
+                        set(obs.name for obs in observation_stratifications)
+                    )
+                    if missing_stratifications:
+                        raise ValueError(
```
rmudambi, line 90: Two notes:
* This will only record the missing stratifications from the first iteration that has some missing. It would be better to communicate all missing stratifications
* I think it' might be worth warning when stratifications are registered and then never used.

tests/framework/results/test_manager.py
```diff
@@ -301,19 +301,28 @@ def test_stratified_metrics_initialization():
     )
 
 
-def test_no_stratified_metrics_initialization():
-    """Test that if no stratifications are registered then we initialize a
+def test_metrics_initialized_from_no_stratifications_observer():
```
rmudambi, line 304: It would be good to include tests for each of the cases I called out above.

rmudambi (changes requested):

src/vivarium/framework/results/manager.py
```diff
@@ -75,29 +75,43 @@ def setup(self, builder: "Builder"):
...
                 _pop_filter,
-                stratification_names,
+                all_requested_stratification_names,
             ), observations in self._results_context.observations[event_name].items():
                 for observation in observations:
                     measure = observation.name
-                    observation_stratifications = [
+
+                    # Batch missing stratifications
+                    missing_stratifications = set(
```
rmudambi, line 95: Nit: you're performing `set(all_requested_stratification_names)` twice. You could save it as a local variable.

src/vivarium/framework/results/manager.py
```diff
@@ -108,13 +122,29 @@ def on_post_setup(self, _: Event):
...
                     self._metrics[measure] = pd.DataFrame(
                         data=0.0,
                         columns=["value"],
                         index=idx,
                     )
 
+        if unused_stratifications:
+            self.logger.info(
+                "The following Stratifications are registered but not used by any "
+                f"Observers: \n{sorted(list(unused_stratifications))}"
```
rmudambi, line 136: Nit: I don't think stratifications or observers are customarily capitalized.

src/vivarium/framework/results/manager.py
```diff
@@ -108,13 +122,29 @@ def on_post_setup(self, _: Event):
...
+            )
+        if all_missing_stratifications:
```
rmudambi, line 138: Nit: I think I prefer `all_missing_stratifications` to be `missing_stratifications` and what you currently call `missing_stratifications` to be something like `observation_missing_stratifications` because this line `if all_missing_stratifications:` makes me think it means you're checking if all stratifications are missing.

rmudambi (approved):

src/vivarium/framework/results/manager.py
stevebachmeier: I was really struggling w/ naming these things here!

PR #408, opened 2024-05-02
bugfux handle single stratifications

stevebachmeier:
### Description

- *Category*:  bugfix
- *JIRA issue*: na

### Changes and notes
 
There is a bug where if an observer is stratified by a single column,
the dtype of that single-layer MultiIndex is Categorical. But when there
are more than one columns to stratify by, the MultiIndex dtypes were
Object. This was breaking the updater method and returning NaN.

### Testing

Modified the QuidditchWins observer to be stratified by a single
column and then updated all tests to pass.

### DIFF
--- a/src/vivarium/framework/results/manager.py
+++ b/src/vivarium/framework/results/manager.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Callable, List, Optional, Union
 
 import pandas as pd
+from pandas.api.types import CategoricalDtype
 
 from vivarium.framework.event import Event
 from vivarium.framework.results.context import ResultsContext
@@ -119,21 +120,22 @@ class ResultsManager(Manager):
                         for stratification in requested_and_registered_stratifications
                     }
                     if stratification_values:
-                        # Create index of the complete stratifications
-                        idx = pd.MultiIndex.from_tuples(
+                        stratification_names = list(stratification_values.keys())
+                        df = pd.DataFrame(
                             list(itertools.product(*stratification_values.values())),
-                            names=stratification_values.keys(),
-                        )
+                            columns=stratification_names,
+                        ).astype(CategoricalDtype)
                     else:
                         # We are aggregating the entire population so create a single-row index
-                        idx = pd.Index(["all"], name="stratification")
+                        stratification_names = ["stratification"]
+                        df = pd.DataFrame(["all"], columns=stratification_names).astype(
+                            CategoricalDtype
+                        )
 
                     # Initialize a zeros dataframe
-                    self._metrics[measure] = pd.DataFrame(
-                        data=0.0,
-                        columns=["value"],
-                        index=idx,
-                    )
+                    df["value"] = 0.0
+                    df = df.set_index(stratification_names)
+                    self._metrics[measure] = df
 
         if unused_stratifications:
             self.logger.info(
--- a/src/vivarium/framework/results/manager.py
+++ b/src/vivarium/framework/results/manager.py
@@ -134,8 +134,7 @@ class ResultsManager(Manager):
 
                     # Initialize a zeros dataframe
                     df["value"] = 0.0
-                    df = df.set_index(stratification_names)
-                    self._metrics[measure] = df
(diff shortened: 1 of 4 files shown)


tests/framework/results/helpers.py
```diff
@@ -92,51 +92,61 @@ def on_time_step(self, pop_data: SimulantData) -> None:
...
 class QuidditchWinsObserver(Component):
+    """Observer that is stratified by a single column ('familiar')"""
+
     def setup(self, builder: Builder) -> None:
         builder.results.register_observation(
             name="quidditch_wins",
             aggregator_sources=["quidditch_wins"],
             aggregator=sum,
-            excluded_stratifications=["student_house"],
+            excluded_stratifications=["student_house", "power_level"],
```
albrja, line 137: I'm not the biggest Harry Potter fan but I don't remember anyone having a power level attribute.

albrja (approved):

src/vivarium/framework/results/manager.py
```diff
@@ -119,21 +120,22 @@ def on_post_setup(self, _: Event):
...
 
                     # Initialize a zeros dataframe
-                    self._metrics[measure] = pd.DataFrame(
-                        data=0.0,
-                        columns=["value"],
-                        index=idx,
-                    )
+                    df["value"] = 0.0
+                    df = df.set_index(stratification_names)
+                    self._metrics[measure] = df
```
rmudambi, line 138: Nit: can combine these two lines.

tests/framework/results/helpers.py
```diff
@@ -92,51 +92,61 @@ def on_time_step(self, pop_data: SimulantData) -> None:
...
         builder.results.register_observation(
             name="house_points",
             aggregator_sources=["house_points"],
             aggregator=sum,
             requires_columns=[
                 "house_points",
-                "student_house",
-                "power_level",
+                # "student_house",
+                # "power_level",
```
rmudambi, line 107: Should these commented lines be deleted?

tests/framework/results/test_manager.py
```diff
@@ -292,20 +294,16 @@ def test_stratified_metrics_initialization():
     STUDENT_HOUSES_LIST = list(STUDENT_HOUSES)
     POWER_LEVELS_STR = [str(lvl) for lvl in POWER_LEVELS]
     assert metrics["house_points"].index.equals(
-        pd.MultiIndex.from_product(
-            [STUDENT_HOUSES_LIST, POWER_LEVELS_STR],
-            names=["student_house", "power_level"],
+        _get_expected_idx(
```
rmudambi, line 297: This is duplicating the logic of the actual code too much imo. I think you should create the expected MultiIndex directly here and the expected CategoricalIndex below.

tests/framework/results/test_manager.py
```diff
@@ -364,6 +362,10 @@ def _check_house_points(pop: pd.DataFrame, step_number: int) -> None:
         """We know that house points are stratified by 'student_house' and 'power_level'.
         and that each wizard of gryffindor and of level 50 and 80 gains a point
         """
+        # Need the stratification columns to be Categorical
```
rmudambi, line 365: Why do they need to be categorical? I don't see any obvious reason for it looking at the code here.

stevebachmeier: Because the very last assertion checks that the dataframes are equal and they're not if they don't have the same multiindex dtypes

rmudambi: Why can't we just assert that all the values are equal?

tests/framework/results/helpers.py
stevebachmeier: You'll have to ask Matt about this

tests/framework/results/test_manager.py
stevebachmeier: Kinda hard to move too far away from the code since it's the only way I know how to cram Category dtypes into a MultiIndex. I've tried soemthing else that is at least a bit different. If this is no good then I can just assert that all values are the same and all names are the same

rmudambi (approved):

hussain-jafari (approved):

PR #409, opened 2024-05-06
first cut

stevebachmeier:
### Description

- *Category*:  refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5043

### Changes and notes
 
This adds an Observer abstract base class from which concrete
observers can inherit from. The primary motivation for this is to reduce the
amount of duplicate code, specifically by providing a place to define
observation `report()` methods. By passing Observer `report()`
instance methods in when registering specific observations, we
get access to the entire Observer instance and the attributes that
are useful/necessary for generating results (e.g. draw number,
seed, and results directory).

Now the only thing we need to tack onto the report event
as "user_data" are the metrics themselves.

### Testing

Tests pass


src/vivarium/framework/engine.py
```diff
@@ -207,6 +207,10 @@ def __init__(
     def name(self) -> str:
         return self._name
 
+    @property
+    def results(self) -> Dict[str, Any]:
```
rmudambi, line 211: Would it maybe be less confusing to have a method `get_results()` than to have this as a property? Since `sim._results` is very different from `sim.results`. I think that's more consistent with how we get things from a Simulation anyway, (e.g. `sim.get_value()` or `sim.get_population()`).

src/vivarium/framework/engine.py
```diff
@@ -207,6 +207,10 @@ def __init__(
...
+        return self._results.metrics
```
rmudambi, line 212: Since we still have the "metrics" pipeline (though we may eventually get rid of it), would it be better to do this?
```suggestion
        return self._values.get_value("metrics")(self.get_population().index)
```

src/vivarium/framework/results/context.py
```diff
@@ -95,7 +95,7 @@ def add_observation(
         additional_stratifications: List[str],
         excluded_stratifications: List[str],
         when: str,
-        report: Callable[[Path, str, pd.DataFrame, str, str], None],
+        report: Callable[..., None],
```
rmudambi, line 98: Is this really the right typing? It looks like it might be `Callable[[str, Any], None]`, but maybe it's just `Callable[[str, pd.DataFrame], None]`

src/vivarium/framework/results/interface.py
```diff
@@ -146,7 +143,7 @@ def register_observation(
         additional_stratifications: List[str] = [],
         excluded_stratifications: List[str] = [],
         when: str = "collect_metrics",
-        report: Callable[[Path, str, pd.DataFrame, str, str], None] = dataframe_to_csv,
+        report: Callable[..., None] = lambda *_: None,
```
rmudambi, line 146: Ditto

src/vivarium/framework/results/manager.py
```diff
@@ -165,20 +165,12 @@ def on_collect_metrics(self, event: Event):
...
                 for observation in observations:
-                    observation.report(
-                        results_dir=results_dir,
-                        measure=observation.name,
-                        results=metrics[observation.name],
-                        random_seed=random_seed,
-                        input_draw=input_draw,
-                    )
+                    measure = observation.name
+                    observation.report(measure, metrics[measure])
```
rmudambi, line 173: Nit: this could be combined into one line.

src/vivarium/framework/results/manager.py
```diff
@@ -292,7 +284,7 @@ def register_observation(
         additional_stratifications: List[str],
         excluded_stratifications: List[str],
         when: str,
-        report: Callable[[Path, str, pd.DataFrame, str, str], None],
+        report: Callable[..., None],
```
rmudambi, line 287: Check typing

src/vivarium/framework/results/observation.py
```diff
@@ -1,27 +1,26 @@
...
+    """
+
     name: str
     pop_filter: str
     stratifications: Tuple[str, ...]
     aggregator_sources: Optional[List[str]]
     aggregator: Callable[[pd.DataFrame], float]
     when: str
-    report: Callable[[Path, str, pd.DataFrame, str, str], None]
+    report: Callable[..., None]
```
rmudambi, line 26: Typing

src/vivarium/framework/results/observer.py
```diff
@@ -0,0 +1,64 @@
...
+        self.get_report_attributes(builder)
+
+    def get_report_attributes(self, builder: Builder) -> None:
+        """Define commonly-used attributes for reporting."""
+        try:
+            self.results_dir = builder.configuration.output_data.results_directory
+        except KeyError:
+            self.results_dir = None
+        self.input_draw = builder.configuration.input_data.input_draw_number
+        self.random_seed = builder.configuration.randomness.random_seed
```
rmudambi, line 36: What happens if you have a simulation with a different ArtifactManager that doesn't have draws?

I'd suggest this:
```suggestion
    def __init__(self):
        super().__init__()
        self.results_dir = None
        self.input_draw = None
        self.random_seed = None
    
    def get_report_attributes(self, builder: Builder) -> None:
        """Define commonly-used attributes for reporting."""
        if "results_directory" in builder.configuration.output_data.:
            self.results_dir = builder.configuration.output_data.results_directory
        if "input_draw_number" in builder.configuration.input_data:
            self.input_draw = builder.configuration.input_data.input_draw_number
        if "random_seed" in builder.configuration.randomness:
            self.random_seed = builder.configuration.randomness.random_seed
```

src/vivarium/framework/results/observer.py
```diff
@@ -0,0 +1,64 @@
...
+        # Sort the columns such that the stratifications (index) are first
+        # and "value" is last and sort the rows by the stratifications.
+        other_cols = [c for c in results.columns if c != "value"]
+        results = results[other_cols + ["value"]].sort_index().reset_index()
+        results.to_csv(Path(self.results_dir) / f"{measure}.csv", index=False)
+
+
+class StratifiedObserver(Observer):
+    @property
+    def configuration_defaults(self) -> Dict[str, Any]:
```
rmudambi, line 56: For discussion: do we like this way of configuring stratifications? We're starting to move in the direction of having all configurations for a given component live within the namespace of the component or manager that is defining it and this doesn't match that pattern. How do we feel about that?

tests/framework/test_time.py
```diff
@@ -34,7 +34,7 @@ def components():
     return [
         MockComponentA("gretchen", "whimsy"),
         Listener("listener"),
-        MockComponentB("spoon", "antelope", 23),
+        MockComponentB("spoon", "antelope", "23"),
```
rmudambi, line 37: Why was this change needed?

tests/framework/results/test_observer.py
```diff
@@ -0,0 +1,98 @@
...
+    results_dir = Path(tmpdir)
+    input_draw = 111
+    random_seed = 222
+    builder = mocker.Mock()
+    builder.configuration.output_data.results_directory = str(results_dir)
+    builder.configuration.input_data.input_draw_number = input_draw
+    builder.configuration.randomness.random_seed = random_seed
+
+    observer = TestObserver()
+    observer.get_report_attributes(builder)
```
rmudambi, line 57: Why not call `observer.setup_component()`? That seems more robust.

src/vivarium/framework/results/context.py
stevebachmeier: But since we're free to define whatever report method we want, then aren't the args unknown?

tests/framework/test_time.py
stevebachmeier: Because later we `self.name.split("_observer")` and you can't split an int.

src/vivarium/framework/results/observer.py
stevebachmeier: e.g.
```
return {
  self.name: {
    "stratification": {
      "exclude": [],
      "include": [],
    },
  },
}
```

stevebachmeier: After discussion, we will leave for now and ask RT if they have a preference

src/vivarium/framework/engine.py
stevebachmeier: I think this needs some re-work. The problem is that down in the `report` method, it calls the pipeline as well. And so we may be modifying the pipelines more than we intend if we start calling it from multiple locations.

I'm actually getting a breaking test with the way the metrics pipeline is set up since it's a acting as a counter every time it's called. That's probably just a biproduct of a silly test fixture but I think still highlights the potential issue?

src/vivarium/framework/results/observer.py
stevebachmeier: This doesn't play well due to the values being nested two levels deep. How about the changes I've pushed using `.to_dict()` and `.get()`?

src/vivarium/framework/results/context.py
```diff
@@ -95,7 +95,7 @@ def add_observation(
         additional_stratifications: List[str],
         excluded_stratifications: List[str],
         when: str,
-        report: Callable[..., None],
+        report: Callable[[str, pd.DataFrame], None],
```
rmudambi, line 98: Did we decide this should be `Callable[[str, pd.DataFrame], None]` and not `Callable[[str, Any], None]`?

src/vivarium/framework/results/observer.py
```diff
@@ -28,12 +34,19 @@ def setup_component(self, builder: Builder) -> None:
...
     def get_report_attributes(self, builder: Builder) -> None:
         """Define commonly-used attributes for reporting."""
-        try:
-            self.results_dir = builder.configuration.output_data.results_directory
-        except KeyError:
-            self.results_dir = None
-        self.input_draw = builder.configuration.input_data.input_draw_number
-        self.random_seed = builder.configuration.randomness.random_seed
+        self.results_dir = (
+            builder.configuration.to_dict()
```
rmudambi, line 38: Nit: you can save `config_dict = builder.configuration.to_dict()` so you only have to make that call once.

rmudambi (approved):

src/vivarium/framework/engine.py
stevebachmeier: I'm confused what the purpose of the `metrics` property is then. We have a `ResultsManagwer._metrics` attribute (defaultdict[str, pd.DataFrame] of measure-specific results), a `ResultsManager.metrics` property which is just `self._metrics.copy()`, and then the `metrics` pipeline which is an empty dict and has a value modifer that just updates it w/ the metrics proeprty (`metrics.update(self.metrics)`).

src/vivarium/framework/results/context.py
stevebachmeier: for now at least, ResultsContext.gather_results does indeed return specifically a dataframe

PR #415, opened 2024-05-14
Feature/sbachmei/mic 5016 add typing to results modules

stevebachmeier:
### Description

- *Category*:  other
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5016
### Changes and notes
 
this just adds typing to the results modules. Note that it is NOT
passing mypy, but rather just implements the low hanging fruit.

### Testing
 tests continue to pass

### DIFF
--- a/src/vivarium/component.py
+++ b/src/vivarium/component.py
@@ -10,7 +10,7 @@ simulations.
 import re
 from abc import ABC
 from inspect import signature
-from typing import TYPE_CHECKING, Any, Dict, List, Optional
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
 from loguru._logger import Logger
 
@@ -78,9 +78,9 @@ class Component(ABC):
         value: str,
         continuous_columns: List[str] = [],
         categorical_columns: List[str] = [],
-        key_name: str = None,
+        key_name: Optional[str] = None,
         **kwargs: Dict[str, Any],
-    ) -> dict:
+    ) -> Dict[str, Union[str, List[Optional[str]]]]:
         config = {
             "value": value,
             "continuous_columns": continuous_columns,
@@ -348,7 +348,7 @@ class Component(ABC):
     # Lifecycle methods #
     #####################
 
-    def __init__(self):
+    def __init__(self) -> None:
         """
         Initializes a new instance of the Component class.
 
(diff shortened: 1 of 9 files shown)


src/vivarium/component.py
```diff
@@ -348,7 +348,7 @@ def simulation_end_priority(self) -> int:
     # Lifecycle methods #
     #####################
 
-    def __init__(self):
+    def __init__(self) -> None:
```
stevebachmeier, line 351: [PEP-484](https://peps.python.org/pep-0484/) clarifies that __init__ should be typed as returning None:

> (Note that the return type of __init__ ought to be annotated with -> None. The reason for this is subtle. If __init__ assumed a return annotation of -> None, would that mean that an argument-less, un-annotated __init__ method should still be type-checked? Rather than leaving this ambiguous or introducing an exception to the exception, we simply say that __init__ ought to have a return annotation; the default behavior is thus the same as for other methods.)

hussain-jafari (approved):

src/vivarium/framework/results/stratification.py
```diff
@@ -45,9 +47,9 @@ def __post_init__(self):
                 )
             self.mapper = self._default_mapper
             self.is_vectorized = True
-        if not len(self.categories):
+        if len(self.categories) == 0:
```
patricktnast, line 50: nit: I think PEP 8 recommends `if not self.categories:` here, but obviously this is fine as well.

patricktnast (approved):

rmudambi (approved):

src/vivarium/framework/results/stratification.py
stevebachmeier: Interesting, pylint was actually saying _not_ to do that - though I didn't track down the source

stevebachmeier: Ohhh, I see, it doesn't like `if not len(seq)` since you can just do `if not seq` instead. Not sure why it's not yelling at me about the `== 0` though...

```
For sequences, (strings, lists, tuples), use the fact that empty sequences are false:
# Correct:
if not seq:
if seq:
# Wrong:
if len(seq):
if not len(seq):
```

PR #417, opened 2024-05-23
strengthen error message

stevebachmeier:
### Description

- *Category*:  other
- *JIRA issue*: na

### Changes and notes
 
This adds a Stratification __str__ (which admittedly isn't a whole lot different than
the default) and uses that in the error message when trying to register a stratification
with a name that already exists.

### Testing

Moved the test from test_manager to test_context and slightly strengthened
it by adding a pytest.raises(..., match=) pattern

### DIFF
--- a/src/vivarium/framework/results/stratification.py
+++ b/src/vivarium/framework/results/stratification.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Callable, List, Optional, Union, cast
+from typing import Callable, List, Optional, Union
 
 import pandas as pd
 from pandas.api.types import CategoricalDtype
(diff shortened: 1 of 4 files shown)


src/vivarium/framework/results/stratification.py
```diff
@@ -38,6 +38,13 @@ class Stratification:
     mapper: Optional[Callable[[Union[pd.Series[str], pd.DataFrame]], pd.Series[str]]] = None
     is_vectorized: bool = False
 
+    def __str__(self) -> str:
+        return (
+            f"Stratification '{self.name}' with sources {self.sources}, "
+            f"categories {self.categories}, mapper {self.mapper.__name__}, "
+            f"and is_vectorized {self.is_vectorized}"
```
rmudambi, line 45: I think is vectorized can be left out of the str? It's not super important, as it's really just there to help vivarium know how to use the mapper.

rmudambi (approved):

albrja (approved):

PR #418, opened 2024-05-24
Update metrics for better support

stevebachmeier:
### Description

- *Category*:  feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5062

### Changes and notes
 
This PR implements required changes to update vph's DisabilityObserver
to register a single ylds observation instead of one per cause. Two major
things are done here to that end:

1. We allow for the observation.aggregator to return a pd.Series or a float
instead of just a float. This is required in order to allow for a metrics
dataframe of any number of columns (instead of just a pd.Series or
1-col dataframe)
2. Special-case handling in the event that the metrics indeed has
more than a single column. This can happen when we want to stratify
result by more than just registered stratifications, e.g. the 
DisabilityObserver wants to stratify yld results by disease, but disease
can't be a proper stratification because a simulant can have more than
one at a time.

### Testing

Tests pass

### DIFF
--- a/src/vivarium/framework/results/context.py
+++ b/src/vivarium/framework/results/context.py
@@ -100,7 +100,7 @@ class ResultsContext:
         name: str,
         pop_filter: str,
         aggregator_sources: Optional[List[str]],
-        aggregator: Callable[[pd.DataFrame], float],
+        aggregator: Callable[[pd.DataFrame], Union[float, pd.Series[float]]],
         additional_stratifications: List[str],
         excluded_stratifications: List[str],
         when: str,
@@ -144,7 +144,10 @@ class ResultsContext:
                     aggregator = observation.aggregator
                     aggregates = self._aggregate(pop_groups, aggregator_sources, aggregator)
                     aggregates = self._coerce_to_dataframe(aggregates)
-                    aggregates.rename(columns={aggregates.columns[0]: "value"}, inplace=True)
+                    if aggregates.shape[1] == 1:
+                        aggregates.rename(
+                            columns={aggregates.columns[0]: "value"}, inplace=True
+                        )
                     aggregates = self._expand_index(aggregates)
                     if not list(stratifications):
                         aggregates.index.name = "stratification"
@@ -184,7 +187,7 @@ class ResultsContext:
     def _aggregate(
         pop_groups: Union[DataFrameGroupBy, pd.DataFrame],
         aggregator_sources: Optional[List[str]],
-        aggregator: Callable[[pd.DataFrame], float],
+        aggregator: Callable[[pd.DataFrame], Union[float, pd.Series[float]]],
     ) -> Union[pd.Series[float], pd.DataFrame]:
         return (
             pop_groups[aggregator_sources].apply(aggregator).fillna(0.0)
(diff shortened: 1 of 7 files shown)


src/vivarium/framework/results/context.py
```diff
@@ -144,7 +144,10 @@ def gather_results(
                     aggregator = observation.aggregator
                     aggregates = self._aggregate(pop_groups, aggregator_sources, aggregator)
                     aggregates = self._coerce_to_dataframe(aggregates)
-                    aggregates.rename(columns={aggregates.columns[0]: "value"}, inplace=True)
+                    if aggregates.shape[1] == 1:
```
rmudambi, line 147: Should we move this inside `_coerce_to_dataframe`? We also talked about renaming `_coerce_to_dataframe`.

src/vivarium/framework/results/manager.py
```diff
@@ -179,6 +179,12 @@ def on_report(self, event: Event) -> None:
                     )
 
     def gather_results(self, event_name: str, event: Event) -> None:
+        """Update the existing metrics with new results. Typically, a results
```
rmudambi, line 182: I don't think we need to say what is typical and atypical behavior. Both 1 column or multiple columns are accepted inputs, so we can just say that we support each and how.

src/vivarium/framework/results/manager.py
```diff
@@ -188,7 +194,12 @@ def gather_results(self, event_name: str, event: Event) -> None:
                     raise ValueError(
                         f"There is a results group {results_group} but no corresponding measure"
                     )
-                self._metrics[measure] += results_group
+                # Look for extra columns in the results_group and initialize with 0.
+                extra_cols = list(set(results_group.columns) - {METRICS_COLUMN})
```
rmudambi, line 198: Isn't this going to reset the observation to 0 every time, so we won't get any accumulation? We really want to compare the columns in `results_group` to those in `self._metrics[measure]`. If this didn't cause any tests to fail, our coverage likely isn't good enough.

tests/framework/results/test_context.py
```diff
@@ -420,34 +419,6 @@ def test_gather_results_with_no_stratifications():
     )
 
 
-def test__bad_aggregator_return():
```
rmudambi, line 423: We could still have this test with an aggregator returning something that isn't a DataFrame or a Series.

rmudambi (changes requested):

tests/framework/results/test_context.py
stevebachmeier: But doesn't .apply on a df always return one of those two?

stevebachmeier: In fact, I think we can double down and not only remove the test but actually remove the isinstance check

tests/framework/results/test_context.py
```diff
@@ -571,26 +571,15 @@ def test__aggregate(stratifications, aggregator_sources, aggregator):
         ),
     ],
 )
-def test__coerce_to_dataframe(aggregates):
-    new_aggregates = ResultsContext()._coerce_to_dataframe(aggregates=aggregates)
+def test__format(aggregates):
```
rmudambi, line 574: These two tests are named identically? Can they have more descriptive names?

rmudambi (approved):

PR #420, opened 2024-05-29
move run_simulation to instance method

stevebachmeier:
### Description

- *Category*:  refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5085

### Changes and notes
 This just moves the steps of a simulation (setup, initialize_simulants,
run, finalize, and report) into a new SimulationContext.run() method. This
will allow for cleaner handling in the upcoming ParallelSimulationContext
(because we will have a method to actually overwrite).

NOTE: I renamed the old SimulationContext.run() (which just iterated through
time and called SimulationContext.step()) to SimulationContext.take_steps().
If that's too large of a change I can put it back (and come up w/ a different 
name for the new wrapper method)

### Testing

added a test; all pass


src/vivarium/framework/engine.py
```diff
@@ -399,9 +407,5 @@ def run_simulation(
     simulation = SimulationContext(
         model_specification, components, configuration, plugin_configuration
     )
-    simulation.setup()
-    simulation.initialize_simulants()
-    simulation.run()
-    simulation.finalize()
-    simulation.report()
+    simulation.run_simulation()
```
stevebachmeier, line 410: It kind of annoys moe that this method is the same name as the function that calls it but I can't think of anything better (I tried using a simulation.run() and then renaming the old simulation.run() to simulation.take_steps(), but that's too user-facing b/c of IteractiveContext)

tests/framework/test_engine.py
```diff
@@ -172,6 +172,39 @@ def test_SimulationContext_name_management(SimulationContext):
...
+    sim.report = mocker.Mock()
+
+    sim.run_simulation()
+
+    # Assert each mocked method was called as expected
+    sim.setup.assert_has_calls([mocker.call()], any_order=True)
+    sim.initialize_simulants.assert_has_calls([mocker.call()], any_order=True)
+    sim.run.assert_has_calls([mocker.call()], any_order=True)
+    sim.finalize.assert_has_calls([mocker.call()], any_order=True)
+    sim.report.assert_has_calls([mocker.call()], any_order=True)
```
rmudambi, line 192: Don't we care about the order? We do have a test somewhere in pseudopeople that tests the order of the application of noise functions if you want to test it that closely.

src/vivarium/framework/engine.py
rmudambi: Would it be cleaner to get rid of this `run_simulation` method and just create the SimulationContext in the cli method and call sim.run_simulation there?

rmudambi (approved):

src/vivarium/framework/engine.py
stevebachmeier: I thought about it but kinda liked the idea that this entry point instantiated the object and then ran it. But looking at it now I don't like it so much

tests/framework/test_engine.py
stevebachmeier: I figured order was maintained by vivarium itself but yeah, I'll go remind myself how I did order checks in psp

src/vivarium/framework/engine.py
stevebachmeier: Though whoever wrote the docs went out of the way to discuss how "the combination of initializing and running the simulation is encpsulated in the `run_simulation` command..." ([reference](https://github.com/ihmeuw/vivarium/blob/main/docs/source/concepts/entry_points.rst))

We should chat about whether this makes sense or not

tests/framework/test_engine.py
stevebachmeier: We do, but I figured that's enforced by vivarium and I recalled it being somewhat difficult to implement in psp. I'll have a look though

PR #421, opened 2024-05-30
remove metrics pipeline

stevebachmeier:
### Description

- *Category*:  refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5088

### Changes and notes
 
The metrics pipeline was temporarily used during development of the
new results management system but is no longer necessary. Instead
of keeping the indirection, this removes it.

### Testing

Tests are updated and all pass


src/vivarium/framework/engine.py
```diff
@@ -208,7 +206,7 @@ def name(self) -> str:
         return self._name
 
     def get_results(self) -> Dict[str, Any]:
-        return self._values.get_value("metrics")(self.get_population().index)
+        return self._results.metrics
```
rmudambi, line 209: Remind me - the plan is for the next PR to do the formatting here, right?

stevebachmeier: Yes

PR #428, opened 2024-06-03
add cli 'simulate run' test

stevebachmeier:
### Description

- *Category*:  test
- *JIRA issue*: na

### Changes and notes
 We didn't have any tests for the actual cli entry point. This adds
at least one and also checks that a new metadata.yaml is created
as expected. (we may very well get rid of this metadata.yaml file altogether
but it seemed like a good place to put some simulation info, namely
the simulation run time)

### Testing

tests pass

### DIFF
--- a/tests/interface/test_cli.py
+++ b/tests/interface/test_cli.py
@@ -1,0 +1,50 @@
+from pathlib import Path
+
+import pytest
+import yaml
+from click.testing import CliRunner
+
+from tests import HARRY_POTTER_CONFIG
+from vivarium.interface.cli import simulate
+
+
+@pytest.fixture
+def runner():
+    return CliRunner()
+
+
+@pytest.fixture
+def model_spec(base_config, tmp_path) -> str:
+    base_config.update(HARRY_POTTER_CONFIG)
+    model_spec = {}
+    model_spec["configuration"] = base_config.to_dict()
+    model_spec["components"] = {
+        "tests": [
+            "Hogwarts()",
+            "HousePointsObserver()",
+            "NoStratificationsQuidditchWinsObserver()",
+            "QuidditchWinsObserver()",
+            "HogwartsResultsStratifier()",
+        ],
+    }
+    filepath = f"{tmp_path}/model_spec.yaml"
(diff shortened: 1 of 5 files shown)


tests/interface/test_cli.py
```diff
@@ -0,0 +1,50 @@
...
+    assert {file.name for file in results_dir.rglob("*.parquet")} == {
+        "house_points.parquet",
+        "quidditch_wins.parquet",
+        "no_stratifications_quidditch_wins.parquet",
+    }
+    with open(results_dir / "metadata.yaml", "r") as f:
+        metadata = yaml.safe_load(f)
+
+    assert set(metadata) == set(["input_draw", "random_seed", "simulation_run_time"])
+    assert metadata["simulation_run_time"] > 0.0
```
rmudambi, line 50: Since you're here, why not also test the other flags of `simulate run`, like `-i`.

rmudambi (approved):

tests/interface/test_cli.py
stevebachmeier: @rmudambi Can you expand what you mean? Are you proposing doing the same test but w/ an actual test artifact (`-i`) and/or a location defined (`-l`) and the corresponding artifact? These tests don't actually use an artifact so those options weren't readily testable.

rmudambi: You could mock one though?

stevebachmeier: @rmudambi How is this? It kinda feels like cheating to just add to metadata, but I also feel like this kind of stuff is usesful for each sim to save out anyway?

rmudambi: Does the model spec get dumped somewhere? Checking the artifact key in the model spec would be a better test if you can do that.

albrja (approved):

tests/interface/test_cli.py
stevebachmeier: I'm going to go ahead and merge this and then update to use the artifact key in the model spec in another PR that saves out the model spec

PR #430, opened 2024-06-04
remove -l option; save out finalized model spec

stevebachmeier:
### Description

- *Category*:  other
- *JIRA issue*: 
    - https://jira.ihme.washington.edu/browse/MIC-5101
    - https://jira.ihme.washington.edu/browse/MIC-5103

### Changes and notes
 This feeds two birds with one scone by removing the unused
`simulate run --location` option as well as saving out the finalized
model specification. I'm open to suggestions on a better name than
"complete_model_specification".

### Testing

tests pass

### DIFF
--- a/docs/source/concepts/entry_points.rst
+++ b/docs/source/concepts/entry_points.rst
@@ -88,10 +88,10 @@ modification of a simulation defined in a model specification file.
    intended use case, so tread cautiously.
 
 By intention, the context exposes a very simple interface for managing the
-:ref:`simulation lifecycle <lifecycle_concept>`.  The combination of
-initializing and running the simulation is encapsulated in the
-:func:`run_simulation <vivarium.framework.engine.run_simulation>` command
-also available in the :mod:`engine <vivarium.framework.engine>`.
+:ref:`simulation lifecycle <lifecycle_concept>`.  Once a context is instantiated,
+all steps of running a simulation are encapsulated in the
+:func:`run_simulation <vivarium.framework.engine.SimulationContext.run_simulation>`
+method.
 
 The simulation :class:`Builder <vivarium.framework.engine.Builder>` is also
 part of the engine. It is the main interface that components use to interact
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -25,6 +25,7 @@ from typing import Any, Dict, List, Set, Union
 
 import numpy as np
 import pandas as pd
+import yaml
 from layered_config_tree import ConfigurationKeyError, LayeredConfigTree
 
 from vivarium import Component
@@ -112,6 +113,15 @@ class SimulationContext:
         model_specification = build_model_specification(
             model_specification, component_configuration, configuration, plugin_configuration
         )
+        results_dir = (
+            model_specification.to_dict()
+            .get("configuration", {})
+            .get("output_data", {})
+            .get("results_directory")
+        )
+        if results_dir:
+            with open(f"{results_dir}/complete_model_specification.yaml", "w") as f:
+                yaml.dump(model_specification.to_dict(), f)
 
         self._plugin_configuration = model_specification.plugins
         self._component_configuration = model_specification.components
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -21,7 +21,7 @@ tools to easily setup and run a simulation.
 
 from pathlib import Path
 from pprint import pformat
-from typing import Any, Dict, List, Set, Union
+from typing import Any, Dict, List, Optional, Set, Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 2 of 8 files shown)


src/vivarium/framework/randomness/index_map.py
```diff
@@ -215,7 +215,7 @@ def _convert_to_ten_digit_int(self, column: pd.Series) -> pd.Series:
 
         """
         if isinstance(column.iloc[0], datetime.datetime):
-            column = self._clip_to_seconds(column.view(np.int64))
+            column = self._clip_to_seconds(column.astype(np.int64))
```
stevebachmeier, line 218: I just did this to get rid of a very noise FutureWarning

tests/framework/test_engine.py
```diff
@@ -331,7 +331,7 @@ def test_SimulationContext_report_write(base_config, components, tmpdir):
     ]
     finished_sim = run_simulation(base_config, components, configuration)
     # Check for expected results written
-    results_list = [file for file in results_root.rglob("*")]
+    results_list = [file for file in results_root.rglob("*.parquet")]
```
stevebachmeier, line 334: I had to specify `.parquet` b/c now we also have the model spec in the results dir. (I of course could've added that to the expected list instead)

src/vivarium/framework/engine.py
```diff
@@ -112,6 +113,15 @@ def __init__(
...
             model_specification, component_configuration, configuration, plugin_configuration
         )
+        results_dir = (
+            model_specification.to_dict()
+            .get("configuration", {})
+            .get("output_data", {})
+            .get("results_directory")
+        )
+        if results_dir:
+            with open(f"{results_dir}/complete_model_specification.yaml", "w") as f:
```
stevebachmeier, line 123: Happy to take suggestions on what to call this. We already have a "model_specification.yaml" copied here as well (which is the model spec input we provide to `simulate run`).

tests/interface/test_cli.py
```diff
@@ -33,15 +33,24 @@ def model_spec(base_config, tmp_path) -> str:
...
+    expected_parameters = {
+        "model_specification",
+        "artifact_path",
+        "results_directory",
+        "verbose",
+        "quiet",
+        "with_debugger",
+    }
+    different_params = run_parameters.symmetric_difference(expected_parameters)
+    if run_parameters.symmetric_difference(expected_parameters):
```
stevebachmeier, line 48: The idea here is to raise in the event we ever change `simulate run` options or args again and forget to update the `expected_parameters`. It should help force us to consider adding an assertion like we do below for `-i` (which is really the only one we need to test since `-o` is inherently tested since we load from there and `-v`, `-q`, and `--pdb` don't really need testing)

src/vivarium/framework/engine.py
rmudambi: Why are we doing this in `__init__` rather than in `run_simulation`? This doesn't seem like the right place, since this means that every simulation (including the `InteractiveContext`) will write to disk, which doesn't seem right.

Regarding the name, I'd suggest comparing this to what we output in psimulate, which I believe prints out the entire model spec.

tests/framework/test_engine.py
rmudambi: I thought you said we are already writing out a model spec file. How was that being ignored?

stevebachmeier: @rmudambi Because that version of the model spec is [copied in the cli.py](https://github.com/ihmeuw/vivarium/blob/main/src/vivarium/interface/cli.py#L134) which this test side-steps.

src/vivarium/framework/engine.py
stevebachmeier: hmm, good call

stevebachmeier: Actually, I don't think it would - it only writes out to disk if a results dir is provided (ie via `simulate run` OR if a user went out of their way to add one to the config...in which case I think it should be saved out)

docs/source/concepts/entry_points.rst
```diff
@@ -88,10 +88,10 @@ modification of a simulation defined in a model specification file.
    intended use case, so tread cautiously.
 
 By intention, the context exposes a very simple interface for managing the
-:ref:`simulation lifecycle <lifecycle_concept>`.  The combination of
```
stevebachmeier, line 91: The fact that this was initially documented this way is the main reason I chose initially not to move the SimulationContext instantiation to cli.py. But here we are anyway

src/vivarium/interface/cli.py
```diff
@@ -129,21 +128,25 @@ def run(
...
 
-    main = handle_exceptions(run_simulation, logger, with_debugger)
-    finished_sim = main(model_specification, configuration=override_configuration)
+    sim = SimulationContext(
+        model_specification=model_specification, configuration=override_configuration
+    )
+    with open(f"{results_directory}/model_specification.yaml", "w") as f:
+        yaml.dump(sim.model_specification.to_dict(), f)
+
+    handle_exceptions(sim.run_simulation, logger, with_debugger)()
```
rmudambi, line 144: I think I like this better as two lines like it was before. It's easy to miss the trailing `()` when reading.

src/vivarium/interface/cli.py
```diff
@@ -129,21 +128,25 @@ def run(
...
     if artifact_path:
         input_data["artifact_path"] = artifact_path
     override_configuration = {"output_data": output_data, "input_data": input_data}
 
-    main = handle_exceptions(run_simulation, logger, with_debugger)
-    finished_sim = main(model_specification, configuration=override_configuration)
+    sim = SimulationContext(
+        model_specification=model_specification, configuration=override_configuration
+    )
+    with open(f"{results_directory}/model_specification.yaml", "w") as f:
```
rmudambi, line 141: This is fine, but if we want to systematize how we do running of simulations across our 3 types of simulations (i.e. `SimulationContext`, `InteractiveContext`, and `ParallelSimulationContext` or whatever we call it), it might make sense to figure out what the operations are across the three and how to encapsulate them to reduce duplication of code and improve flexibility. 

It may be easier to think through this once you start working on the `ParallelSimulationContext`.

rmudambi (approved):

src/vivarium/interface/cli.py
stevebachmeier: Yeah I agree. I'll merge this for now and it'll likely change later

rmudambi (approved):

patricktnast (approved):

PR #432, opened 2024-06-10
Mic 5086/remove hardcoding columns

albrja:
### Determines and creates table based on inputs and not hard coded values for build_table testing utility function.
- *Category*: Test
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5086

### Changes and notes
-Determines the key and values for parameter and key columns instead of hard coded values

### Testing
All tests pass

### DIFF
--- a/src/vivarium/testing_utilities.py
+++ b/src/vivarium/testing_utilities.py
@@ -9,7 +9,7 @@ Utility functions and classes to make testing ``vivarium`` components easier.
 
 from itertools import product
 from pathlib import Path
-from typing import List
+from typing import Any, Dict, List
 
 import numpy as np
 import pandas as pd
@@ -150,14 +150,32 @@ def _non_crn_build_population(
 
 
 def build_table(
-    value,
-    parameter_columns={
+    value: Any,
+    parameter_columns: Dict = {
         "age": (0, 125),
         "year": (1990, 2020),
     },
-    key_columns={"sex": ("Female", "Male")},
-    value_columns=["value"],
-):
+    key_columns: Dict = {"sex": ("Female", "Male")},
+    value_columns: List = ["value"],
+) -> pd.DataFrame:
+    """
+    Parameters
+    ----------
+    value
+        Value(s) to put in the value columns of a lookup table.
+    parameter_columns
+        A dictionary where the keys are parameter (continuous) columns of a lookup table
+        and the values are tuple of the range (inclusive) for that column.
+    key_columns
+        A dictionary where the keys are key (categorical) columns of a lookup table
+        and the values are a tuple of the categories for that column
+    value_columns
+        A list of value columns that will appear in the returned lookup table
+
+    Returns
+        A pandas dataframe that has the cartesian product of the range of all parameter columns
+        and the values of the key columns.
+    """
     if not isinstance(value, list):
         value = [value] * len(value_columns)
 
@@ -174,23 +192,22 @@ def build_table(
     products = product(*product_dict.values())
 
     rows = []
-    for p in products:
-        # Note: p is going to be a tuple of the cartesian product of the key column values and parameter column
+    for item in products:
+        # Note: item is going to be a tuple of the cartesian product of the key column values and parameter column
         # values and will be ordered in the order of the parameter then key dict keys
         r_values = []
-        for v in value:
-            if v is None:
+        for val in value:
+            if val is None:
                 r_values.append(np.random.random())
-            elif callable(v):
-                r_values.append(v(p))
+            elif callable(val):
+                r_values.append(val(item))
             else:
-                r_values.append(v)
+                r_values.append(val)
 
         # Get list of values for rows (index values)
-        key_columns_index_values = p[len(parameter_columns) :]
-        key_columns_index_values = [val for val in key_columns_index_values]
-        # Trasnform parameter column values
-        parameter_columns_index_values = p[: len(parameter_columns)]
+        key_columns_index_values = list(item[len(parameter_columns) :])
+        # Transform parameter column values
+        parameter_columns_index_values = item[: len(parameter_columns)]
         # Create intervals for parameter columns. Example year, year+1 for year_start and year_end
         parameter_columns_index_values = [
             v for val in parameter_columns_index_values for v in (val, val + 1)
--- a/src/vivarium/testing_utilities.py
+++ b/src/vivarium/testing_utilities.py
@@ -10,6 +10,7 @@ Utility functions and classes to make testing ``vivarium`` components easier.
 from pathlib import Path
 from typing import List
 
+from itertools import product
 import numpy as np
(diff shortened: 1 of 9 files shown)


src/vivarium/testing_utilities.py
```diff
@@ -148,31 +149,62 @@ def _non_crn_build_population(
     return population
 
 
-def build_table(value, year_start, year_end, columns=("age", "year", "sex", "value")):
-    value_columns = columns[3:]
+def build_table(
+    value,
+    parameter_columns={
+        "age": (0, 125),
```
stevebachmeier, line 155: Are these hard-coded defaults defined somewhere? If not I think they should be (_especially_ b/c 2020 will change basically every year)

src/vivarium/testing_utilities.py
```diff
@@ -148,31 +149,62 @@ def _non_crn_build_population(
     return population
 
 
-def build_table(value, year_start, year_end, columns=("age", "year", "sex", "value")):
-    value_columns = columns[3:]
+def build_table(
```
stevebachmeier, line 152: Add type hints

src/vivarium/testing_utilities.py
```diff
@@ -148,31 +149,62 @@ def _non_crn_build_population(
...
+):
     if not isinstance(value, list):
         value = [value] * len(value_columns)
 
     if len(value) != len(value_columns):
         raise ValueError("Number of values must match number of value columns")
 
+    # Get product of parameter columns
+    range_parameter_product = {
+        key: list(range(value[0], value[1])) for key, value in parameter_columns.items()
```
stevebachmeier, line 169: This does NOT include the end points (2020 and 125); is that intended

src/vivarium/testing_utilities.py
```diff
@@ -148,31 +149,62 @@ def _non_crn_build_population(
...
+    }
+    # Build out dict of items we will need cartesian product of to make dataframe
+    product_dict = dict(range_parameter_product)
```
stevebachmeier, line 172: range_parameter_product is already a dict

src/vivarium/testing_utilities.py
```diff
@@ -7,6 +7,7 @@
 
 """
 
+from itertools import product
```
stevebachmeier, line 10: Why is this file packaged instead of living under the tests module?

src/vivarium/testing_utilities.py
```diff
@@ -148,31 +149,62 @@ def _non_crn_build_population(
...
-                r_values = []
-                for v in value:
-                    if v is None:
-                        r_values.append(np.random.random())
-                    elif callable(v):
-                        r_values.append(v(age, sex, year))
-                    else:
-                        r_values.append(v)
-                rows.append([age, age + 1, year, year + 1, sex] + r_values)
+    for p in products:
```
stevebachmeier, line 177: nit: Maybe it's a me thing, but I really dislike single-letter variable names. At least they're not `s`, `n`, or `c` (which mucks up pdb)

src/vivarium/testing_utilities.py
```diff
@@ -148,31 +149,62 @@ def _non_crn_build_population(
...
+                r_values.append(np.random.random())
+            elif callable(v):
+                r_values.append(v(p))
+            else:
+                r_values.append(v)
+
+        # Get list of values for rows (index values)
+        key_columns_index_values = p[len(parameter_columns) :]
+        key_columns_index_values = [val for val in key_columns_index_values]
+        # Trasnform parameter column values
```
stevebachmeier, line 192: typo

src/vivarium/testing_utilities.py
```diff
@@ -148,31 +149,62 @@ def _non_crn_build_population(
...
+        parameter_columns_index_values = p[: len(parameter_columns)]
+        # Create intervals for parameter columns. Example year, year+1 for year_start and year_end
+        parameter_columns_index_values = [
+            v for val in parameter_columns_index_values for v in (val, val + 1)
```
stevebachmeier, line 196: Oh, this is why it's ok to be messing the end points

src/vivarium/testing_utilities.py
stevebachmeier: nm, it's handled below

tests/framework/lookup/test_lookup.py
```diff
@@ -195,7 +202,13 @@ def test_invalid_data_type_build_table(base_config):
 def test_lookup_table_interpolated_return_types(base_config):
     year_start = base_config.time.start.year
     year_end = base_config.time.end.year
-    data = build_table(lambda age, sex, year: year, year_start, year_end)
+    data = build_table(
+        lambda x: x[0],
+        parameter_columns={
```
stevebachmeier, line 207: I think there should be a test that includes 1 parameter col and more than 2 parameter cols since it's been generalized

stevebachmeier (approved):

tests/framework/artifact/test_manager.py
```diff
@@ -57,7 +59,12 @@ def test_subset_rows():
 def test_subset_columns():
     values = [0, "red", 100]
     data = build_table(
-        values, 1990, 2010, columns=("age", "year", "sex", "draw", "color", "value")
+        values,
+        parameter_columns={
```
rmudambi, line 63: Why did you need to specify the parameter columns here?

rmudambi (approved):

src/vivarium/testing_utilities.py
albrja: They are just arbitrary values based relatively on what our normal data looks like. No real data is compared to this these are just values used in this brute force approach to make this dataframe.

src/vivarium/testing_utilities.py
albrja: This is making a copy so we don't overwrite the original.

src/vivarium/testing_utilities.py
albrja: I don't understand your question.

tests/framework/artifact/test_manager.py
albrja: To change the range of the year column to keep it the same as it was since it is different from the default - 2010 vs 2020

src/vivarium/testing_utilities.py
rmudambi: Because it is used by VPH. When `vivarium_testing_utils` is more developed, it should probably move there.

tests/framework/artifact/test_manager.py
rmudambi: I understand that. I'm saying that the specific endpoints probably don't matter for this test, so you should be able to use the defaults.

rmudambi (approved):

patricktnast (approved):

PR #433, opened 2024-06-13
Feature/sbachmei/mic 5117 refactor observation to different types

stevebachmeier:
### Description

- *Category*:  refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5117

### Changes and notes
 
This is not a feature change, just a refactor. To pave the way for different
types of observations (different than just a summation type), this
changes the registration entry point to be more specific for our most-used
type, i.e. `register_summing_observation`.

Changes include:

- Adding newly-named entry points
- Changed `Observation` to be a more generic class by removing the
  stratification-specific attributes.
- Creating a new `StratifiedObservation` class which inherits from
  `Observation` and adds the stratification-specific attributes. It does
  not define a creator attribute or the new updater or formatter attributes.
- Creating a new `SummingObservation` which inherits from`StratifiedObservation`
  and specifically defines the creator and updater attributes.
    - The relevent code for these two new attributes are moved from `ResultsContext`
       and into the `SummingObservation` as methods.
- Reordering the args for observation instantiation (annoying, sorry)

### Testing

Updated tests and they pass.

### DIFF
--- a/src/vivarium/framework/randomness/index_map.py
+++ b/src/vivarium/framework/randomness/index_map.py
@@ -215,7 +215,7 @@ class IndexMap:
 
         """
         if isinstance(column.iloc[0], datetime.datetime):
-            column = self._clip_to_seconds(column.view(np.int64))
+            column = self._clip_to_seconds(column.astype(np.int64))
         elif np.issubdtype(column.iloc[0], np.integer):
             if not len(column >= 0) == len(column):
                 raise RandomnessError(
--- a/src/vivarium/framework/results/interface.py
+++ b/src/vivarium/framework/results/interface.py
@@ -144,7 +144,7 @@ class ResultsInterface:
     # def register_observation(self, ...):
     #     pass
 
-    def register_summing_observation(
+    def register_adding_observation(
         self,
         name: str,
         pop_filter: str = "tracked==True",
@@ -192,7 +192,7 @@ class ResultsInterface:
         ------
         None
         """
-        self._manager.register_summing_observation(
+        self._manager.register_adding_observation(
             name,
             pop_filter,
             when,
--- a/src/vivarium/framework/results/interface.py
+++ b/src/vivarium/framework/results/interface.py
@@ -50,6 +50,10 @@ class ResultsInterface:
         """The name of this ResultsInterface."""
         return self._name
 
+    ##################################
+    # Stratification-related methods #
+    ##################################
+
     # TODO: It is not reflected in the sample code here, but the “when” parameter should be added
     #  to the stratification registration calls, probably as a List. Consider this after observer implementation
     def register_stratification(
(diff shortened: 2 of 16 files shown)


albrja (approved):

src/vivarium/framework/randomness/index_map.py
```diff
@@ -215,7 +215,7 @@ def _convert_to_ten_digit_int(self, column: pd.Series) -> pd.Series:
 
         """
         if isinstance(column.iloc[0], datetime.datetime):
-            column = self._clip_to_seconds(column.view(np.int64))
+            column = self._clip_to_seconds(column.astype(np.int64))
```
stevebachmeier, line 218: unrelated change to remove [pd.Series.astype DeprecationWarnings](https://pandas.pydata.org/docs/reference/api/pandas.Series.view.html)

src/vivarium/framework/results/manager.py
```diff
@@ -180,22 +180,21 @@ def gather_results(self, event_name: str, event: Event) -> None:
...
         ):
             if results_group is not None:
                 if measure is None:
                     raise ValueError(
                         f"There is a results group {results_group} but no corresponding measure"
                     )
-                # Look for extra columns in the results_group and initialize with 0.
-                extra_cols = list(
-                    set(results_group.columns) - set(self._raw_results[measure].columns)
+                if updater is None:
```
rmudambi, line 191: When can `measure` or `updater` be `None`? This should error on registration rather than during `gather_results`.

rmudambi (approved):

tests/framework/results/test_context.py
```diff
@@ -389,7 +389,7 @@ def test_gather_results_with_empty_pop_filter():
         formatter=lambda: None,
     )
 
-    for result, _measure in ctx.gather_results(population, event_name):
+    for result, _measure, _formatter in ctx.gather_results(population, event_name):
```
patricktnast, line 392: `ctx.gather_results` returns `updater`, not `formatter`, right?

patricktnast (approved):

src/vivarium/framework/results/interface.py
```diff
@@ -132,20 +136,28 @@ def register_binned_stratification(
...
         when: str = "collect_metrics",
         formatter: Callable[
             [str, pd.DataFrame], pd.DataFrame
         ] = lambda measure, results: results,
+        additional_stratifications: List[str] = [],
+        excluded_stratifications: List[str] = [],
+        aggregator_sources: Optional[List[str]] = None,
+        aggregator: Callable[[pd.DataFrame], Union[float, pd.Series[float]]] = len,
+        requires_columns: List[str] = [],
+        requires_values: List[str] = [],
```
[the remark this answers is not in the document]
stevebachmeier: Yes, will do in next PR

src/vivarium/framework/results/manager.py
stevebachmeier: I don't think they can be none. Will remove these in next PR

stevebachmeier: Oh actually @rmudambi I remember now - I had to add these to proprly handle the type hinting since `self._results_context.gather_results` can indeed yield None, None, None (when there is no population after filtering). Kind of a sad fact of type hints I think.

stevebachmeier: Oh, duh, I can check that all three are not none. I'll fix in next PR

PR #440, opened 2024-06-21
move results initialization to Observation attributes

stevebachmeier:
### Description

- *Category*:  refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4297

### Changes and notes
 This is mostly a copy/paste job that moves the initialization of results
(0s dataframe for StratifiedObservations and an empty dataframe for
UnstratifedObservations) from the results manager to a new
`results_initializer` attribute.

The tricky part that remains very ugly but I can't think of a better way offhand
is how to handle batch-logging for missing and/or unused stratifications. As always,
I'm open to better ideas!

### Testing

tests pass

### DIFF
--- a/src/vivarium/framework/results/manager.py
+++ b/src/vivarium/framework/results/manager.py
@@ -293,31 +293,6 @@ class ResultsManager(Manager):
     # Helper methods #
     ##################
 
-    @staticmethod
-    def _track_stratifications(
-        measure: str,
-        event_requested_stratification_names: set[str],
-        registered_stratification_names: set[str],
-        missing_stratifications: Dict[str, set[str]],
-        used_stratifications: set[str],
-    ) -> set[str]:
-        """Track used and missing stratifications for batch-logging"""
-
-        # Update the `missing_stratifications`` dict for this particular measure/observer
-        # if there are any missing stratifications
-        observer_missing_stratifications = event_requested_stratification_names.difference(
-            registered_stratification_names
-        )
-        if observer_missing_stratifications:
-            missing_stratifications[measure] = observer_missing_stratifications
-
-        # Add newly used stratifications to the running `used_stratifications` list
-        used_stratifications = used_stratifications.union(
-            event_requested_stratification_names
-        )
-
-        return used_stratifications
-
     def _get_stratifications(
         self,
         stratifications: List[str] = [],
--- a/src/vivarium/framework/results/manager.py
+++ b/src/vivarium/framework/results/manager.py
@@ -79,6 +79,30 @@ class ResultsManager(Manager):
 
         self.set_default_stratifications(builder)
 
+    @staticmethod
+    def _track_stratifications(
+        measure: str,
+        event_requested_stratification_names: set[str],
+        registered_stratification_names: set[str],
+        missing_stratifications: Dict[str, set[str]],
+        unused_stratifications: set[str],
+    ) -> set[str]:
+        """Track stratifications for batch-logging"""
+
+        # Batch missing stratifications
+        observer_missing_stratifications = event_requested_stratification_names.difference(
+            registered_stratification_names
+        )
+        if observer_missing_stratifications:
+            missing_stratifications[measure] = observer_missing_stratifications
+
+        # Remove stratifications from the running list of unused stratifications
+        unused_stratifications = unused_stratifications.difference(
+            event_requested_stratification_names
+        )
+
+        return unused_stratifications
+
     def on_post_setup(self, _: Event) -> None:
         """Initialize results with 0s DataFrame' for each measure and all stratifications"""
         registered_stratifications = self._results_context.stratifications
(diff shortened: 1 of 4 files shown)


src/vivarium/framework/results/manager.py
```diff
@@ -93,36 +86,38 @@ def on_post_setup(self, _: Event) -> None:
...
+                    kwargs["requested_stratification_names"] = requested_stratification_names
+
+                    df, unused_stratifications = observation.results_initializer(
+                        measure, **kwargs
+                    )
+                    kwargs["unused_stratifications"] = unused_stratifications
                     self._raw_results[measure] = df
 
-        if unused_stratifications:
+        if len(unused_stratifications) > 0:
```
rmudambi, line 115: Why did you change this and the line below?

src/vivarium/framework/results/manager.py
```diff
@@ -93,36 +86,38 @@ def on_post_setup(self, _: Event) -> None:
...
-                            missing_stratifications,
-                            unused_stratifications,
-                        )
-                    else:
-                        # Initialize a completely empty dataframe
-                        df = pd.DataFrame()
+                    kwargs["requested_stratification_names"] = requested_stratification_names
+
+                    df, unused_stratifications = observation.results_initializer(
+                        measure, **kwargs
```
rmudambi, line 110: There's no reason to create a `kwargs` dictionary here. It just makes the code more confusing to read.

src/vivarium/framework/results/observation.py
```diff
@@ -96,6 +113,59 @@ def __init__(
...
+        requested_stratification_names: List[str],
+        registered_stratifications: List[Stratification],
+        registered_stratification_names: Set[str],
+        missing_stratifications: Dict[str, Set[str]],
+        unused_stratifications: Set[str],
+    ) -> Tuple[pd.DataFrame, Set[str]]:
+        """Initialize a dataframe of 0s with complete set of stratifications as the
+        index and return it along with any registered but unused stratifications
+        """
+        requested_stratification_names = set(requested_stratification_names)
```
rmudambi, line 128: I know this is what the code was already doing, but it doesn't make sense to be checking if we have missing or extra stratifications here. This method should just be initializing the dataframe. We should manage this validation natively in the manager.

src/vivarium/framework/results/manager.py
```diff
@@ -90,34 +114,37 @@ def on_post_setup(self, _: Event) -> None:
...
                     measure = observation.name
-                    kwargs["requested_stratification_names"] = requested_stratification_names
-
-                    df, unused_stratifications = observation.results_initializer(
-                        measure, **kwargs
+                    df = observation.results_initializer(
+                        event_requested_stratification_names, registered_stratifications
                     )
-                    kwargs["unused_stratifications"] = unused_stratifications
                     self._raw_results[measure] = df
```
rmudambi, line 132: These lines can be combined

src/vivarium/framework/results/manager.py
```diff
@@ -90,34 +114,37 @@ def on_post_setup(self, _: Event) -> None:
...
-
-        if len(unused_stratifications) > 0:
+                    if observation.stratifications is not None:
+                        unused_stratifications = self._track_stratifications(
```
rmudambi, line 134: Isn't it simpler to build up the set of stratifications that have actually been used and then compare that to the set of all stratifications that have been registered than to maintain a list of stratifications that haven't been used?

src/vivarium/framework/results/observation.py
```diff
@@ -164,7 +146,7 @@ def initialize_results(
         df[VALUE_COLUMN] = 0.0
         df = df.set_index(stratification_names)
 
-        return df, unused_stratifications
+        return df
```
rmudambi, line 149: So much cleaner!

src/vivarium/framework/results/manager.py
stevebachmeier: I'm indifferent

src/vivarium/framework/results/manager.py
```diff
@@ -313,6 +313,30 @@ def register_observation(
...
 
+    @staticmethod
+    def _track_stratifications(
+        measure: str,
+        event_requested_stratification_names: set[str],
+        registered_stratification_names: set[str],
+        missing_stratifications: Dict[str, set[str]],
+        used_stratifications: set[str],
+    ) -> set[str]:
+        """Track unused and missing stratifications for batch-logging"""
```
albrja, line 324: This doc string to me implies this function is going to do something with unused/missing stratifications but you return used stratifications which is a bit confusing

stevebachmeier: @albrja I've updated the comments to hopefully make it clearer?

src/vivarium/framework/results/manager.py
```diff
@@ -126,19 +102,21 @@ def on_post_setup(self, _: Event) -> None:
...
+                        used_stratifications = self._track_stratifications(
                             measure,
                             event_requested_stratification_names,
                             registered_stratification_names,
                             missing_stratifications,
-                            unused_stratifications,
+                            used_stratifications,
                         )
 
+        unused_stratifications = registered_stratification_names.difference(
```
rmudambi, line 117: This works but isn't something like this simpler?
```python
used_stratifications = set()
for event_name in self._results_context.observations:
    for _, stratifications in self._results_context.observations[event_name]:
        used_stratifications |= stratifications

unused_stratifications = registered_stratifications - used stratifications
missing_stratifications = used_stratifications - registered_stratifications
```

We lose information about the specific observers that are requesting unregistered stratifications, but I'm not sure how important that information is. It should be pretty easy for a user to track down in any case.

rmudambi (approved):

src/vivarium/framework/results/manager.py
stevebachmeier: Isn't that exactly what I'm doing (only it's happening in the `_track_stratifications` helper function)?

stevebachmeier: Wait, I think I misread what you were saying

stevebachmeier: Ok, yeah I see what you're saying. Yes, if you're ok w/ losing the measure-specific info then I agree this is much better. Done

PR #443, opened 2024-06-28
Bugfix/sbachmei/mic 5138 enforce pythonhashseed

stevebachmeier:
### Description

- *Category*:  bugfix
- *JIRA issue*: na

### Changes and notes
 
This was a bear. The short of it is that results may not be reproducible
due to randomized hashing of unordered objects (e.g. sets). This PR raises 
if the user hadn't set PYTHONHASHSEED to 42.

### Testing

tests pass

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,6 +18,8 @@ jobs:
         python-version: ["3.9", "3.10", "3.11"]
     steps:
       - uses: actions/checkout@v3
+      - name: Set PYTHONHASHSEED
+        run: echo "PYTHONHASHSEED=42" >> $GITHUB_ENV
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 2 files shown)


patricktnast (approved):
Some really fine debugging work here!

hussain-jafari (approved):

.github/workflows/build.yml
```diff
@@ -18,6 +18,8 @@ jobs:
         python-version: ["3.9", "3.10", "3.11"]
     steps:
       - uses: actions/checkout@v3
+      - name: Set PYTHONHASHSEED
+        run: echo "PYTHONHASHSEED=42" >> $GITHUB_ENV
```
albrja, line 22: I don't understand why this needs to be done for builds... is it just so te warning below doesnt get raised?

stevebachmeier: Right. We're enforcing that PYTHONHASHSEED == 42 when `SimulationContext` is instantiated. Plenty of pytests in this repo do indeed instantiate so we need to set it here as well (remember that the github jobs are NOT being run in your local env/shell)

stevebachmeier:
NOTE: We decided not to merge this change because we feel the onus should be on the designer/implementer to ensure they are working with different data structures correctly. We also intend to add e2e tests to test against reproducibility.

PR #453, opened 2024-07-23
implement to_observe method to Observations

stevebachmeier:
### Description

- *Category*:  feature
- *JIRA issue*: na

### Changes and notes
 This adds a `to_observe` method to all observations. The default
is True which results in results being gathered at every time event/time
step. You can pass in an Event to the method to help determine whether
it should return False which results in no observation being gathered.

### Testing

Added a test; all pass

### DIFF
--- a/src/vivarium/framework/lookup/interpolation.py
+++ b/src/vivarium/framework/lookup/interpolation.py
@@ -109,7 +109,9 @@ class Interpolation:
             )
 
         if self.categorical_parameters:
-            sub_tables = interpolants.groupby(list(self.categorical_parameters))
+            sub_tables = interpolants.groupby(
+                list(self.categorical_parameters), observed=False
+            )
         else:
             sub_tables = [(None, interpolants)]
         # specify some numeric type for columns, so they won't be objects but
--- a/src/vivarium/framework/results/context.py
+++ b/src/vivarium/framework/results/context.py
@@ -181,22 +181,13 @@ class ResultsContext:
                 yield None, None, None
             else:
                 if stratifications is None:
-                    for observation in observations:
-                        if not observation.to_observe(event):
-                            yield None, None, None
-                        else:
-                            df = observation.results_gatherer(filtered_pop)
-                            yield df, observation.name, observation.results_updater
+                    pop = filtered_pop
                 else:
-                    pop_groups = self._get_groups(stratifications, filtered_pop)
-                    for observation in observations:
-                        if not observation.to_observe(event):
-                            yield None, None, None
-                        else:
-                            aggregates = observation.results_gatherer(
-                                pop_groups, stratifications
-                            )
-                            yield aggregates, observation.name, observation.results_updater
+                    pop = self._get_groups(stratifications, filtered_pop)
+                for observation in observations:
+                    yield observation.gather_results(
+                        event, pop, stratifications
+                    ), observation.name, observation.results_updater
 
     @staticmethod
     def _filter_population(population: pd.DataFrame, pop_filter: str) -> pd.DataFrame:
--- a/src/vivarium/framework/results/context.py
+++ b/src/vivarium/framework/results/context.py
@@ -138,9 +138,9 @@ class ResultsContext:
         already_used = None
         if self.observations:
             # NOTE: self.observations is a list where each item is a dictionary
-            # of the form {event_name: {(pop_filter, stratifications): List[Observation]}}.
+            # of the form {lifecycle_phase: {(pop_filter, stratifications): List[Observation]}}.
             # We use a triple-nested for loop to iterate over only the list of Observations
-            # (i.e. we do not need the event_name, pop_filter, or stratifications).
+            # (i.e. we do not need the lifecycle_phase, pop_filter, or stratifications).
             for observation_details in self.observations.values():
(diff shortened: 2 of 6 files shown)


src/vivarium/framework/results/context.py
```diff
@@ -181,13 +182,21 @@ def gather_results(
             else:
                 if stratifications is None:
                     for observation in observations:
-                        df = observation.results_gatherer(filtered_pop)
-                        yield df, observation.name, observation.results_updater
+                        if not observation.to_observe(event):
+                            yield None, None, None
```
albrja, line 186: This always confuses me. Why do we yield instead of return?

albrja (approved):

albrja:
"The default is True which results in results being gathered at every time event/time
step" - for clarification, does this mean every lifecycle event or no?

src/vivarium/framework/results/context.py
stevebachmeier: computational efficiency. We _could_ combine these datasets into a mega-list and return it all at once and then loop through as needed later. But generators allow for lazy evaluation so we don't need to store all that at once in memory

src/vivarium/framework/lookup/interpolation.py
```diff
@@ -109,7 +109,9 @@ def __call__(self, interpolants: pd.DataFrame) -> pd.DataFrame:
             )
 
         if self.categorical_parameters:
-            sub_tables = interpolants.groupby(list(self.categorical_parameters))
+            sub_tables = interpolants.groupby(
+                list(self.categorical_parameters), observed=False
```
stevebachmeier, line 113: This just got rid of a bunch of FutureWarnings (`observed` is changing from default False to default True)

src/vivarium/framework/results/context.py
```diff
@@ -181,13 +182,21 @@ def gather_results(
             else:
                 if stratifications is None:
                     for observation in observations:
-                        df = observation.results_gatherer(filtered_pop)
-                        yield df, observation.name, observation.results_updater
+                        if not observation.to_observe(event):
```
patricktnast, line 185: I think my slight preference would be to put more of this in `observation` if possible, since we are only calling methods from `observation`. Then one could do e.g. 
```
    yield observation.get_this_tuple_of_results_if_to_observe(event, filtered_pop)
```
(name not serious ofc)
My reasoning here would be that we can avoid deeply nesting conditions and encapsulate things from observation more (e.g. maybe you don't need to_observe in the interface if you do this but need a different method instead. Doesn't matter much either way.

patricktnast (approved):

tests/framework/results/test_context.py
```diff
@@ -30,6 +31,12 @@ def _aggregate_state_person_time(x: pd.DataFrame) -> float:
     return len(x) * (28 / 365.25)
 
 
+@pytest.fixture
+def mocked_event(mocker) -> Event:
+    event: Event = mocker.Mock(spec=Event)
```
albrja, line 36: I wonder what made you do this

albrja (approved):

patricktnast (approved):
I like it!

src/vivarium/framework/results/observation.py
```diff
@@ -35,6 +37,21 @@ class BaseObservation(ABC):
     results_updater: Callable[[pd.DataFrame, pd.DataFrame], pd.DataFrame]
     results_formatter: Callable[[str, pd.DataFrame], pd.DataFrame]
     stratifications: Optional[Tuple[str]]
+    to_observe: Callable[[Event], bool]
+
+    def gather_results(
```
rmudambi, line 42: Having methods called `gather_results` and `results_gatherer` is very confusing. We need better names for these methods.

rmudambi (approved):

hussain-jafari (approved):

src/vivarium/framework/lookup/interpolation.py
stevebachmeier:
```
observed: bool, default False
    This only applies if any of the groupers are Categoricals. If True: only show observed values for
    categorical groupers. If False: show all values for categorical groupers.

    Deprecated since version 2.1.0: The default value will change to True in a future version of pandas.
```

src/vivarium/framework/results/context.py
```diff
@@ -155,7 +156,7 @@ def register_observation(
         ].append(observation)
 
     def gather_results(
-        self, population: pd.DataFrame, event_name: str
+        self, population: pd.DataFrame, event_name: str, event: Event
```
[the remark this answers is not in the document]
stevebachmeier: Hmm, I think I just chose a poor name for "event" name. Really, it's lifecycle phase (e.g. "collect_metrics")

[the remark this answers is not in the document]
stevebachmeier: I've changed the name of this variable to "lifecycle_phase"

src/vivarium/framework/results/observation.py
stevebachmeier: Agreed - renamed to this to simply "observe", i.e. 

```
    yield observation.observe(event, pop, stratificatons) ...
    ...
```

PR #455, opened 2024-08-01
Add backup writing to vivarium simulation Context

patricktnast:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5203

### Changes and notes
 
This is a prerequisite for the simulation backups in psimulate. I elected to make the SimulationContext pickle itself, and also make a new public "past stop time" method that we can use from VCT.

For mysterious reasons, we can't pickle the sim object within a pytest context unless I globally disable cap logging with "pytest -s". Pickling works just fine in normal use, and trying to disable caplog or capsys etc. per-test hasn't worked so far. this appears to be something wrong on `dill`'s end, cf
https://github.com/uqfoundation/dill/issues/663
https://github.com/pytest-dev/pytest/issues/10845
https://github.com/pytest-dev/pytest/issues/12448
For now I've put in a skipped test.

### DIFF
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -253,7 +253,7 @@ class SimulationContext:
         self._clock.step_forward(self.get_population().index)
 
     def step(self) -> None:
-        self._logger.debug(self._clock.time)
+        self._logger.debug(self.current_time)
         for event in self.time_step_events:
             self._logger.debug(f"Event: {event}")
             self._lifecycle.set_state(event)
@@ -272,14 +272,14 @@ class SimulationContext:
     ) -> None:
         if backup_freq:
             time_to_save = time() + backup_freq
-            while self._clock.time < self._clock.stop_time:
+            while self.current_time < self._clock.stop_time:
                 self.step()
                 if time() >= time_to_save:
                     self._logger.debug(f"Writing Simulation Backup to {backup_path}")
                     self.write_backup(backup_path)
                     time_to_save = time() + backup_freq
         else:
-            while self._clock.time < self._clock.stop_time:
+            while self.current_time < self._clock.stop_time:
                 self.step()
 
     def finalize(self) -> None:
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -21,8 +21,10 @@ tools to easily setup and run a simulation.
 
 from pathlib import Path
 from pprint import pformat
+from time import time
 from typing import Any, Dict, List, Optional, Set, Union
 
+import dill
 import numpy as np
 import pandas as pd
 import yaml
@@ -259,7 +261,7 @@ class SimulationContext:
         self._clock.step_forward(self.get_population().index)
 
     def run(self) -> None:
-        while self._clock.time < self._clock.stop_time:
+        while not self.past_stop_time():
             self.step()
 
     def finalize(self) -> None:
@@ -297,6 +299,10 @@ class SimulationContext:
         except ConfigurationKeyError:
             self._logger.info("No results directory set; results are not written to disk.")
 
+    def write_backup(self, backup_path: Path) -> None:
+        with open(backup_path, "wb") as f:
+            dill.dump(self, f)
+
     def get_performance_metrics(self) -> pd.DataFrame:
         timing_dict = self._lifecycle.timings
         total_time = np.sum([np.sum(v) for v in timing_dict.values()])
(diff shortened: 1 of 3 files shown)


src/vivarium/framework/engine.py
```diff
@@ -322,6 +327,9 @@ def add_components(self, component_list: List[Component]) -> None:
     def get_population(self, untracked: bool = True) -> pd.DataFrame:
         return self._population.get_population(untracked)
 
+    def past_stop_time(self) -> bool:
```
stevebachmeier, line 330: nit: why the indirection?

Also, maybe flipping the logic back and naming it `before_stop_time` is clearer? (1) you don't need the not and (2) technically you're not "past" the stop time if the clock time == stop time.

src/vivarium/framework/engine.py
```diff
@@ -297,6 +298,10 @@ def _write_results(self) -> None:
         except ConfigurationKeyError:
             self._logger.info("No results directory set; results are not written to disk.")
 
+    def write_backup(self, backup_path: Path) -> None:
+        with open(backup_path, "wb") as f:
```
stevebachmeier, line 302: Should the backup path have a default that's handled here in vivarium?

stevebachmeier (approved):

src/vivarium/framework/engine.py
patricktnast: We use this again in `vivarium_cluster_tools`, because I wanted to add the new `run` method incorporating the backups there instead of from simulationcontext. So this is just to avoid using private attributes from outside SimulationContext.

src/vivarium/framework/engine.py
patricktnast: I suppose we could, I didnt have a great idea for what it should be

src/vivarium/framework/engine.py
patricktnast: originally, i thought putting the e.g. `run_with_backups` method in vivarium wouldn't be feasible, but it definitely seems possible. maybe check the PR there and see if you'd rather we incorporate that function as a method here and simply call it from VCT

tests/framework/test_engine.py
```diff
@@ -347,6 +348,17 @@ def test_SimulationContext_report_write(SimulationContext, base_config, componen
...
+@pytest.mark.skip(reason="TODO: Figure out how to make Dill serialize in pytest")
+def test_SimulationContext_write_backup(SimulationContext, tmpdir):
+    sim = SimulationContext()
+    backup_path = tmpdir / "backup.pkl"
+    sim.write_backup(backup_path)
+    assert backup_path.exists()
+    with open(backup_path, "rb") as f:
+        sim_backup = dill.load(f)
+    assert isinstance(sim_backup, SimulationContext)
+
```
albrja, line 360: Would there be a way you could test that a backup is saved up until failure? This might be more a VCT but that is a test we will definitely want.

albrja (approved):

src/vivarium/framework/engine.py
```diff
@@ -259,7 +260,7 @@ def step(self) -> None:
         self._clock.step_forward(self.get_population().index)
 
     def run(self) -> None:
-        while self._clock.time < self._clock.stop_time:
+        while not self.past_stop_time():
```
rmudambi, line 263: I think it's better to put all of the backup logic here rather than in VCT. Then VCT can just call `sim.run(backup_path, backup_freq)`, where these values are None if we aren't backing up.

We can also get rid of `past_stop_time` if we do that.

tests/framework/test_engine.py
```diff
@@ -347,6 +348,17 @@ def test_SimulationContext_report_write(SimulationContext, base_config, componen
         assert results.equals(written_results)
 
 
+@pytest.mark.skip(reason="TODO: Figure out how to make Dill serialize in pytest")
```
rmudambi, line 351: Could you mock dill to get the tests at least to run?

patricktnast: done

src/vivarium/framework/engine.py
patricktnast: removed

src/vivarium/framework/engine.py
patricktnast: how's this?

tests/framework/test_engine.py
patricktnast: Since I refactored sim.run(), I added a test that we save a certain number of times while the simulation is running. Does that capture the behavior you're looking for or do you think we need to add an explicit execution error somewhere and check against that?

src/vivarium/framework/engine.py
```diff
@@ -207,6 +208,11 @@ def __init__(
     def name(self) -> str:
         return self._name
 
+    @property
+    def current_time(self) -> Time:
```
rmudambi, line 212: Is this actually used anywhere?

patricktnast: It's used by VCT so that we can know how many steps are left in the restarted run.

hussain-jafari (approved):

PR #458, opened 2024-08-06
Mic-5182/Jenkinsfiles

albrja:
### Set up Jenkins pipeline for push and pull requests
- *Category*: CI
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5182


.flake8
```diff
@@ -0,0 +1,46 @@
...
+    tests/**:D105,D106,D107,D205,D207,D208,D212,D214,D215,D301,D4,S101,E266
+    src/vivarium/lib/**:D105,D106,D107,D205,D207,D208,D212,D214,D215,D301,D4
+    **/__init__.py:F401
+
+# Function complexity and line length
+max-complexity = 10
+max-line-length = 95
+
+# Docstrings
+docstring-convention = google
```
albrja, line 46: Bandit and flake8 files were added but are currently being ignored in the Jenkins pipeline. I figured this would make there less work in the future to incorporate them but I am fine with removing them.

rmudambi (commented):
We don't want all these formatting changes. They appear to be driven by flake8, which suggests we shouldn't be running it, since we are using black separately to do our formatting right now.

PR #460, opened 2024-08-07
Feature/sbachmei/mic 5163 exclude unwanted results

stevebachmeier:
### Description

- *Category*:  feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5163

### Changes and notes
 This implements the ability to exclude results via the model spec
with a new `stratification: excluded_categories` key.

The work got a bit bigger than expected b/c we had to handle (1) typical
results that are actual Stratification objects (2) non-typical results that
are not Stratification objects but get handled on a case-by-case bases
(e.g. ylds which you cannot stratify by cause b/c you can have multiple per
person), and (3) results that have an extra unique category that are
not explicitly requested but come "for free" (e.g. all_causes for deaths
and other_causes for disability).

At a high level, what we're doing is modifying what is happening during
ResultsContext.gather_results():

-  We extract any excluded categories for a given result measure and
drop them from the column mapping during each Stratification.__call__ 
- Then, for each (pop_filter, stratfiications) tuple that gets registered
from observations, we _filter_population to both query the defined
pop_filter and also remove any rows that have NaNs in
columns we are stratifying by.

### Testing

tests pass and I ran some small simulations against the maternal
and child models.

### DIFF
--- a/docs/source/tutorials/exploration.rst
+++ b/docs/source/tutorials/exploration.rst
@@ -95,6 +95,7 @@ configuration by simply printing it.
    sim = get_disease_model_simulation()
 
    del sim.configuration['input_data']
+   del sim.configuration['stratification']['excluded_categories']
 
 .. testcode:: configuration
 
--- a/src/vivarium/framework/components/manager.py
+++ b/src/vivarium/framework/components/manager.py
@@ -17,8 +17,7 @@ setup everything it holds when the context itself is setup.
 """
 
 import inspect
-import typing
-from typing import Any, Dict, Iterator, List, Tuple, Union
+from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Sequence, Tuple, Union
 
 from layered_config_tree import (
     ConfigurationError,
@@ -31,7 +30,7 @@ from vivarium.exceptions import VivariumError
 from vivarium.framework.lifecycle import LifeCycleManager
 from vivarium.manager import Manager
 
-if typing.TYPE_CHECKING:
+if TYPE_CHECKING:
     from vivarium.framework.engine import Builder
 
 
@@ -181,7 +180,7 @@ class ComponentManager(Manager):
             self._components.add(c)
 
     def get_components_by_type(
-        self, component_type: Union[type, Tuple[type, ...]]
+        self, component_type: Union[type, Sequence[type]]
     ) -> List[Component]:
         """Get all components that are an instance of ``component_type``.
 
@@ -196,7 +195,8 @@ class ComponentManager(Manager):
             A list of components of type ``component_type``.
 
         """
-        return [c for c in self._components if isinstance(c, component_type)]
+        # Convert component_type to a tuple for isinstance
+        return [c for c in self._components if isinstance(c, tuple(component_type))]
 
     def get_component(self, name: str) -> Component:
         """Get the component with name ``name``.
@@ -348,7 +348,7 @@ class ComponentInterface:
         return self._manager.get_component(name)
 
     def get_components_by_type(
-        self, component_type: Union[type, Tuple[type, ...], list[type]]
+        self, component_type: Union[type, Sequence[type]]
     ) -> List[Component]:
         """Get all components that are an instance of ``component_type``.
 
@@ -364,7 +364,7 @@ class ComponentInterface:
             A list of components of type ``component_type``.
 
         """
-        return self._manager.get_components_by_type(tuple(component_type))
+        return self._manager.get_components_by_type(component_type)
 
     def list_components(self) -> Dict[str, Component]:
         """Get a mapping of component names to components held by the manager.
--- a/src/vivarium/framework/components/manager.py
+++ b/src/vivarium/framework/components/manager.py
@@ -348,7 +348,7 @@ class ComponentInterface:
         return self._manager.get_component(name)
 
     def get_components_by_type(
-        self, component_type: Union[type, Tuple[type, ...]]
+        self, component_type: Union[type, Tuple[type, ...], list[type]]
     ) -> List[Component]:
         """Get all components that are an instance of ``component_type``.
 
@@ -364,7 +364,7 @@ class ComponentInterface:
             A list of components of type ``component_type``.
 
         """
-        return self._manager.get_components_by_type(component_type)
+        return self._manager.get_components_by_type(tuple(component_type))
 
     def list_components(self) -> Dict[str, Component]:
         """Get a mapping of component names to components held by the manager.
--- a/src/vivarium/framework/results/context.py
+++ b/src/vivarium/framework/results/context.py
@@ -61,7 +61,7 @@ class ResultsContext:
         name: str,
         sources: List[str],
         categories: List[str],
-        excluded_categories: Optional[List[str]],
+        excluded_categories: List[str],
         mapper: Optional[Callable[[Union[pd.Series[str], pd.DataFrame]], pd.Series[str]]],
         is_vectorized: bool,
     ) -> None:
@@ -78,7 +78,7 @@ class ResultsContext:
             List of string values that the mapper is allowed to output.
         excluded_categories
             List of mapped string values to be excluded from results processing.
-            If `None` (the default), will use exclusions as defined in the configuration.
+            If empty (the default), will use exclusions as defined in the configuration.
         mapper
             A callable that emits values in `categories` given inputs from columns
             and values in the `requires_columns` and `requires_values`, respectively.
@@ -110,10 +110,10 @@ class ResultsContext:
             )
 
         # Handle excluded categories. If excluded_categories are explicitly
-        # passed in, we use that instead of what is in the odel spec.
+        # passed in, we use that instead of what is in the model spec.
         to_exclude = (
             excluded_categories
-            if excluded_categories is not None
+            if excluded_categories
             else self.excluded_categories.get(name, [])
         )
         unknown_exclusions = set(to_exclude) - set(categories)
--- a/src/vivarium/framework/results/context.py
+++ b/src/vivarium/framework/results/context.py
@@ -43,6 +43,9 @@ class ResultsContext:
 
     def setup(self, builder: Builder) -> None:
         self.logger = builder.logging.get_logger(self.name)
+        self.excluded_categories = (
+            builder.configuration.stratification.excluded_categories.to_dict()
+        )
 
     # noinspection PyAttributeOutsideInit
     def set_default_stratifications(self, default_grouping_columns: List[str]) -> None:
@@ -53,21 +56,12 @@ class ResultsContext:
             )
         self.default_stratifications = default_grouping_columns
 
-    def set_stratification_excluded_categories(
-        self, excluded_categories: dict[str, list[str]]
-    ) -> None:
-        if self.excluded_categories:
-            raise ResultsConfigurationError(
-                "Multiple calls are being made to set excluded stratification categories "
-                "for results production."
-            )
-        self.excluded_categories = excluded_categories
-
     def add_stratification(
         self,
         name: str,
         sources: List[str],
         categories: List[str],
+        excluded_categories: Optional[List[str]],
         mapper: Optional[Callable[[Union[pd.Series[str], pd.DataFrame]], pd.Series[str]]],
         is_vectorized: bool,
     ) -> None:
@@ -82,6 +76,9 @@ class ResultsContext:
             categorization.
         categories
             List of string values that the mapper is allowed to output.
+        excluded_categories
+            List of mapped string values to be excluded from results processing.
+            If `None` (the default), will use exclusions as defined in th model spec.
         mapper
             A callable that emits values in `categories` given inputs from columns
             and values in the `requires_columns` and `requires_values`, respectively.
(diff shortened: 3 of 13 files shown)


src/vivarium/framework/results/context.py
```diff
@@ -57,6 +61,7 @@ def add_stratification(
         name: str,
         sources: List[str],
         categories: List[str],
+        excluded_categories: Optional[List[str]],
```
albrja, line 64: Why is this a list vs a dict now?

src/vivarium/framework/results/interface.py
```diff
@@ -73,6 +73,7 @@ def register_stratification(
         self,
         name: str,
         categories: List[str],
+        excluded_categories: Optional[List[str]] = None,
```
albrja, line 76: Per comment above, do we just convert the dict from the config into a list so now these will always be a list?

albrja (approved):

src/vivarium/framework/results/context.py
stevebachmeier: Not sure what you mean. This is a new thing and was never a dict

stevebachmeier: Oh you're referring to the zoom discussion about the broken doctest?

At the highest / model spec level, `excluded_categories` is a dict because you need to specify which exclusions for which set of results.
```
# model spec
...
configuration:
  stratification:
    excluded_categories:
      cause_of_death:
        - 'stillborn'
      disability:
        - 'all_causes'
        - 'mild_child_wating'
```

But this attribute is attached to a very specific stratification (e.g. 'cause_of_death') and so is just the list to exclude.

docs/source/tutorials/exploration.rst
```diff
@@ -95,6 +95,7 @@ configuration by simply printing it.
    sim = get_disease_model_simulation()
 
    del sim.configuration['input_data']
+   del sim.configuration['stratification']['excluded_categories']
```
stevebachmeier, line 98: This key was throwing errors in the expected `print(sim.configuration)` test below. The actual type of `sim.configuration.stratification.excluded_categories` is an emtpy `LayeredConfigTree` which I just couldn't get to work (I tried None, an empty string, literally nothing, and LayeredConfigTree() and none of them worked).

stevebachmeier: I actually was able to get this test to pass if I set the default to None (instead of an empty dict which then gets converted to a LayeredConfigTree) but that broke a bunch of tests in vph when trying to `.update` the value b/c you cannot update a ConfigNode.

src/vivarium/framework/results/interface.py
```diff
@@ -86,6 +87,9 @@ def register_stratification(
             Name of the of the column created by the stratification.
         categories
             List of string values that the mapper is allowed to output.
+        excluded_categories
+            List of mapped string values to be excluded from results processing.
+            If `None` (the default), will use exclusions as defined in th model spec.
```
rmudambi, line 92: Typo: 
```suggestion
            If `None` (the default), will use exclusions as defined in the configuration.
```

src/vivarium/framework/results/interface.py
```diff
@@ -145,6 +150,7 @@ def register_binned_stratification(
         ------
         None
         """
+        # TODO: implement excluded_categories like in `register_stratification`
```
rmudambi, line 153: Will this be dealt with in a subsequent PR?

src/vivarium/framework/results/manager.py
```diff
@@ -171,6 +177,9 @@ def register_stratification(
             Name of the of the column created by the stratification.
         categories
             List of string values that the mapper is allowed to output.
+        excluded_categories
+            List of mapped string values to be excluded from results processing.
+            If `None` (the default), will use exclusions as defined in th model spec.
```
rmudambi, line 182:
```suggestion
            If `None` (the default), will use exclusions as defined in the configuration
```

src/vivarium/framework/results/context.py
```diff
@@ -71,6 +76,9 @@ def add_stratification(
             categorization.
         categories
             List of string values that the mapper is allowed to output.
+        excluded_categories
+            List of mapped string values to be excluded from results processing.
+            If `None` (the default), will use exclusions as defined in th model spec.
```
rmudambi, line 81:
```suggestion
            If `None` (the default), will use exclusions as defined in the configuration.
```

src/vivarium/framework/results/manager.py
```diff
@@ -71,6 +72,8 @@ def get_results(self) -> Dict[str, pd.DataFrame]:
 
     # noinspection PyAttributeOutsideInit
     def setup(self, builder: "Builder") -> None:
+        self._results_context.setup(builder)
```
rmudambi, line 75: Was this previously not being called at all?

src/vivarium/framework/results/manager.py
```diff
@@ -138,6 +141,8 @@ def gather_results(self, lifecycle_phase: str, event: Event) -> None:
         with 0.0.
         """
         population = self._prepare_population(event)
+        if population.empty:
+            return
```
rmudambi, line 145: Why is this necessary?

src/vivarium/framework/results/stratification.py
```diff
@@ -59,17 +62,29 @@ def __post_init__(self) -> None:
...
         if self.is_vectorized:
-            raw_mapped_column = self.mapper(population[self.sources])
+            mapped_column = self.mapper(population[self.sources])
         else:
-            raw_mapped_column = population[self.sources].apply(self.mapper, axis=1)
-        mapped_column = raw_mapped_column.astype(
-            CategoricalDtype(categories=self.categories, ordered=True)
+            mapped_column = population[self.sources].apply(self.mapper, axis=1)
+
+        if mapped_column.isnull().any():
```
rmudambi, line 75: Why is this check needed? Shouldn't it be sufficient to do the check below?

rmudambi (approved):

src/vivarium/framework/results/interface.py
stevebachmeier: This is at the stratfiication-level and so is always a list. The model spec also provides a list at the stratification level, it's just that each stratification is stored as the keys to a dictionary.

src/vivarium/framework/results/interface.py
stevebachmeier: @rmudambi  I wasn't sure how much of a priority this even is. It wasn't immediately clear to me how to handle excluded categories but then I'm also aware that we don't actually have an example of using `register_binned_stratification`.

src/vivarium/framework/results/manager.py
stevebachmeier: It wasn't, though there previously wasn't anything interesting there (only the logger). But now there's the excluded categories

src/vivarium/framework/results/manager.py
stevebachmeier: I guess it's not strictly necessary but seemed silly to pass an empty population through `gather_results`

src/vivarium/framework/results/stratification.py
stevebachmeier: I was trying to handle the fact that NaNs are not equal and so the set of NaNs would blow up the error message. I'll try and think of a way to handle that though b/c it would be nice to check against other unknown categories as well before raising.

src/vivarium/framework/results/context.py
```diff
@@ -61,7 +61,7 @@ def add_stratification(
         name: str,
         sources: List[str],
         categories: List[str],
-        excluded_categories: Optional[List[str]],
+        excluded_categories: List[str],
```
rmudambi, line 64: Why did you change this from None to an empty list?

src/vivarium/framework/results/stratification.py
```diff
@@ -71,13 +71,19 @@ def __call__(self, population: pd.DataFrame) -> pd.DataFrame:
...
         if unknown_categories:
+            # Reduce all nans to a single one
+            single_nan_list = (
+                [mapped_column[mapped_column.isna()].iat[0]]
+                if mapped_column.isna().any()
+                else []
+            )
+            unknown_categories = single_nan_list + [
+                cat for cat in unknown_categories if not pd.isna(cat)
+            ]
```
rmudambi, line 86: Why can't this be:
```python
unknown_categories = [
    cat for cat in unknown_categories if not pd.isna(cat)
]
if mapped_column.isna().any():
    unknown_categories.append(np.nan)
```

src/vivarium/framework/components/manager.py
```diff
@@ -364,7 +364,7 @@ def get_components_by_type(
             A list of components of type ``component_type``.
 
         """
-        return self._manager.get_components_by_type(component_type)
+        return self._manager.get_components_by_type(tuple(component_type))
```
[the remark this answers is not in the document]
stevebachmeier: It does actually have to be a tuple though for `isinstance`. I do think it's cleaner to convert it to a tuple at the very end so I've done that

src/vivarium/framework/results/context.py
stevebachmeier: It just seemed more consistent w/ the other list arguments defauling to [] instead of None.

src/vivarium/framework/results/stratification.py
stevebachmeier: I feel like pd.isna wasn't working as expected but now I'm not so sure. I'll try again

stevebachmeier: Ok, yeah that works and is way better. I did stick w/ appending the type of NaN since that could potentially shed light on the issue but same basic logic

src/vivarium/framework/results/interface.py
stevebachmeier: I added this to the tests

src/vivarium/framework/results/context.py
stevebachmeier: Put it back to None b/c it conveys different information than an empty list.

src/vivarium/framework/results/context.py
```diff
@@ -166,17 +206,27 @@ def gather_results(
...
+                raise ValueError(
+                    f"Stratification column '{new_column}' "
+                    "already exists in the state table or as a pipeline which is a required "
+                    "name for stratifying results - choose a different name."
+                )
+            population[new_column] = stratification(population)
 
+        # Optimization: We store all the producers by pop_filter and stratifications
+        # so that we only have to apply them once each time we compute results.
         for (pop_filter, stratifications), observations in self.observations[
```
stevebachmeier, line 224: rename to `stratififaton_names` or `*_columns` or something

src/vivarium/framework/results/context.py
```diff
@@ -202,52 +206,88 @@ def gather_results(
...
+
+                    yield (results, observation.name, observation.results_updater)
 
-    @staticmethod
     def _filter_population(
-        population: pd.DataFrame, pop_filter: str, stratifications: tuple[str, ...]
+        self,
+        population: pd.DataFrame,
+        pop_filter: str,
+        stratifications: Optional[tuple[str, ...]],
```
rmudambi, line 250: Nit: rename this to `stratification_names` as well.

src/vivarium/framework/results/context.py
```diff
@@ -202,52 +206,88 @@ def gather_results(
...
     ) -> pd.DataFrame:
         """Filter the population based on the filter string as well as any
         excluded stratification categories
         """
         pop = population.query(pop_filter) if pop_filter else population.copy()
-        # Drop all rows in the 'stratifications' columns that have NaN values
         if stratifications:
-            pop = pop.dropna(subset=list(stratifications))
+            # Drop all rows in the mapped_stratification columns that have NaN values
```
rmudambi, line 257: Nit: maybe explain that we are dropping these rows because being NaN means that the value was previously one of the excluded categories.

src/vivarium/framework/results/context.py
```diff
@@ -202,52 +206,88 @@ def gather_results(
...
+                subset=[
+                    get_mapped_col_name(stratification) for stratification in stratifications
+                ]
+            )
         return pop
 
-    @staticmethod
     def _get_groups(
-        stratifications: Tuple[str, ...], filtered_pop: pd.DataFrame
+        self, stratifications: Tuple[str, ...], filtered_pop: pd.DataFrame
```
rmudambi, line 266: Nit: this can still be static? I don't see `self` being used.

tests/framework/results/test_interface.py
```diff
@@ -16,6 +17,106 @@ def _silly_aggregator(_: pd.DataFrame) -> float:
...
 
 
+####################################
+# Test stratification registration #
+####################################
+
+
+def test_register_stratification(mocker):
+    def _silly_mapper():
+        return "foo"
```
rmudambi, line 27: Shouldn't the mapper return values in `{"some-category", "some-other-category", "some-unwanted-category"}`? I know we're not testing that in this test, but it is a bit confusing.

rmudambi (approved):

src/vivarium/framework/results/context.py
```diff
@@ -166,44 +206,88 @@ def gather_results(
...
-        return (
-            filtered_pop.groupby(list(stratifications), observed=False)
-            if list(stratifications)
-            else filtered_pop.groupby(lambda _: "all")
-        )
+        """Group the population by stratifications.
+        NOTE: Stratifications at this point can be an empty tuple.
+        HACK: It's a bit hacky how we are handling the groupby object if there
+        are no stratifications. The alternative is to use the entire population
+        #nstead of a groupby object, but then we would need to handle
```
patricktnast, line 275:
```suggestion
        instead of a groupby object, but then we would need to handle
```

src/vivarium/framework/results/stratification.py
```diff
@@ -14,14 +16,15 @@ class Stratification:
...
 
-    The `Stratification` class has five fields: `name`, `sources`, `mapper`,
-    `categories`, and `is_vectorized`. The `name` is the name of the column
-    created by the mapper. The `sources` is a list of columns in the extended
-    state table that are the inputs to the mapper function.  Simulants will
-    later be grouped by this column (or these columns) during stratification.
-    `categories` is a set of values that the mapper is allowed to output. The
-    `mapper` is the method that transforms the source to the name column.
-    The method produces an output column by calling the mapper on the source
+    The `Stratification` class has six fields: `name`, `sources`, `mapper`,
```
patricktnast, line 19: I guess this PR only adds to what existed already, but if we're going to go through each attribute one by one, we should probably use a systematic docstring format i.e. what we use for say, Builder 
https://github.com/ihmeuw/vivarium/blob/main/src/vivarium/framework/engine.py
I'd say "X class has # fields" is also unnecessary.

patricktnast (approved):

src/vivarium/framework/results/stratification.py
stevebachmeier: Agree, and I'll do all that doc stuff in the final PRs for this epic. For now I only modified docs if they were categorically incorrect

PR #467, opened 2024-08-19
Fixes for doc building; minor docstring updates

stevebachmeier:
### Description

- *Category*:  documentation
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5220

### Changes and notes
 We previously worked on the content of the results system docstrings
without paying too much attention to whether or not we could actually get
doc builds to pass (let alone how they look). This fixes the failing
docbuilds and makes other minor tweaks to docstrings.

### Testing
 tests pass and docs build.


docs/source/concepts/results.rst
```diff
@@ -0,0 +1,9 @@
+.. _results_concept:
+
+==================
+Simulation Results
+==================
+
+.. todo::
+
+   Everything here.
```
stevebachmeier, line 9: Next PR

src/vivarium/framework/results/context.py
```diff
@@ -132,9 +130,11 @@ def add_stratification(
         Raises
         ------
         ValueError
-            - If the stratification `name` is already used.
```
stevebachmeier, line 135: This wasn't rendering well.

src/vivarium/framework/results/interface.py
```diff
@@ -1,16 +1,14 @@
...
+=================
+Results Interface
+=================
 
 This module provides a :class:`ResultsInterface <ResultsInterface>` class with
 methods to register stratifications and results producers (referred to as "observations")
 to a simulation.
 """
 
-from __future__ import annotations
```
stevebachmeier, line 11: We already had determined that we couldn't use, e.g. `pd.Series[str]` (which requires this import to work currently), but this import itself was throwing warnings about not being able to resolve pd.Series or pd.DataFrame - it was super difficult to track this down as the culprit!

src/vivarium/framework/results/observation.py
```diff
@@ -2,9 +2,20 @@
...
-"""
 
-from __future__ import annotations
+An observation is a class object that records simulation results; they are responsible
+for initializing, gathering, updating, and formatting results.
+
+The provided :class:`BaseObservation` class is an abstract base class that should
+be subclassed by concrete observations. While there are no required abstract methods
+to define when subclassing, the class does provide common attributes as well
+as an `observe` method that determines whether to observe results for a given event.
```
rmudambi, line 12: Shouldn't this say a `to_observe` method?

rmudambi (approved):

src/vivarium/framework/results/observation.py
stevebachmeier: Not quite. There _is_ a `to_observe` attribute (which is a Callable that returns a boolean), but I'm highlighting the `observe` method that is really the thing that calls a given observation's `results_gatherer` method (if `to_observe()` returns true for a given event)

PR #475, opened 2024-08-26
update formatting throughout codebase

stevebachmeier:
### Description

- *Category*:  documentation
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5243

### Changes and notes
 We previously settled on sticking with numpy style documentation.
This PR makes small tweaks for consistency. Most notably, it implements
a new type hints extension so that we do not need to duplicate
type hints in the signature as well as the docstring.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -50,6 +50,7 @@ needs_sphinx = "4.0"
 # ones.
 extensions = [
     "sphinx.ext.autodoc",
+    "sphinx_autodoc_typehints",
     "sphinx.ext.intersphinx",
     "sphinx.ext.doctest",
     "sphinx.ext.todo",
--- a/src/vivarium/component.py
+++ b/src/vivarium/component.py
@@ -5,6 +5,7 @@ Component
 
 A base Component class to be used to create components for use in ``vivarium``
 simulations.
+
 """
 
 import re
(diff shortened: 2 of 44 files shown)


src/vivarium/framework/engine.py
```diff
@@ -358,33 +357,39 @@ class Builder:
 
     Attributes
     ----------
-    logging: LoggingInterface
+    configuration : ``LayeredConfigTree``
```
stevebachmeier, line 360: I couldn't get this to work so I just kept it nit-picked. Quotes didn't work but for some reason double back-ticks did 🤷‍♂️

docs/source/conf.py
```diff
@@ -50,6 +50,7 @@
 # ones.
 extensions = [
     "sphinx.ext.autodoc",
+    "sphinx_autodoc_typehints",
```
stevebachmeier, line 53: This is the thing that allows us to not type hint parameters and returns in the docstring even though we already have them typed in the signature.

albrja (approved):

src/vivarium/component.py
```diff
@@ -159,54 +158,47 @@ def name(self) -> str:
 
     @property
     def sub_components(self) -> List["Component"]:
-        """
-        Provide components managed by this component.
+        """Provide components managed by this component.
 
         Returns
         -------
         List[Component]
```
rmudambi, line 165: Wasn't this supposed to be removed?

rmudambi (approved):

src/vivarium/component.py
stevebachmeier: This (and the `str` return type above it) were the two I reffered to earlier that for some reason sphinx has a problem with removing.

PR #478, opened 2024-09-11
raise an error if a component attempts to access a non-existent population_view

rmudambi:
### Description

- *Category*: refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5322

### Changes and notes
- Raise an error if a component attempts to access a non-existent population_view 

### Testing
All tests pass

### DIFF
--- a/src/vivarium/component.py
+++ b/src/vivarium/component.py
@@ -162,6 +162,7 @@ class Component(ABC):
 
         Returns
         -------
+        PopulationView
             The PopulationView for this component
 
         Raises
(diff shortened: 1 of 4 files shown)


src/vivarium/component.py
```diff
@@ -156,6 +156,28 @@ def name(self) -> str:
 
         return self._name
 
+    @property
+    def population_view(self) -> PopulationView:
+        """Provides the PopulationView for this component.
+
+        Returns
+        -------
+        PopulationView
```
rmudambi, line 165: I thought I didn't need to specify this return type here since the method is typed, but it fails if I don't.

hussain-jafari (approved):

patricktnast (approved):

albrja (approved):

src/vivarium/component.py
stevebachmeier: You shouldn't have to declare it here as long as you're using the sphinx_autodoc_typehints extension (which we are), but I have seen this before as well.

PR #487, opened 2024-09-24
Mypy Stratifications.py

patricktnast:
### Description

- *Category*: 
typing
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5344

### Changes and notes
 
Made the following changes:
Made `unknown_categories` a set as it is originally defined.
dropped the dataclass, since we're just doing a __post_init__ method anyway.
changed the string representation of the mapper, since a callable might not necessarily have a __name__ attr.
made type aliases for the two supported stratification types


src/vivarium/framework/results/stratification.py
```diff
@@ -1,23 +1,23 @@
...
 from pandas.api.types import CategoricalDtype
 
-from vivarium.types import ScalarValue
-
 STRATIFICATION_COLUMN_SUFFIX: str = "mapped_values"
 
 
-@dataclass
+VectorMapper = Callable[[pd.DataFrame], pd.Series]  # type: ignore [type-arg]
+ScalarMapper = Callable[[pd.Series], str]  # type: ignore [type-arg]
```
rmudambi, line 18: Why are you ignoring type-arg here? Why not `pd.Series[Any]`?

src/vivarium/framework/results/stratification.py
```diff
@@ -72,21 +72,27 @@ def __post_init__(self) -> None:
...
-                    f"provided for stratification {self.name}. The list of sources "
-                    "must be of length 1 if no mapper is provided."
-                )
-            self.mapper = self._default_mapper
-            self.is_vectorized = True
-        if not self.categories:
-            raise ValueError("The categories argument must be non-empty.")
+        self.name = name
+        self.sources = sources
+        self.user_provided_mapper = mapper
```
rmudambi, line 77: Should this be "private" (i.e. `self._user_provided_mapper`)?

rmudambi (approved):

rmudambi:
This PR should be targeting `release/v3.0.7`

src/vivarium/framework/results/stratification.py
patricktnast: from the types.py:
`    # TODO: Parameterizing pandas objects fails below python 3.12`
(when using TypeAliases, specifically, I assume)

rmudambi: Oh right, I forgot about that. We should have a comment about this?

src/vivarium/framework/results/stratification.py
```diff
@@ -18,80 +19,47 @@
...
+    name: str
+    """Name of the stratification."""
+    sources: list[str]
+    """A list of the columns and values needed as input for the `mapper`."""
+    categories: list[str]
+    """Exhaustive list of all possible stratification values."""
+    excluded_categories: list[str]
+    """List of possible stratification values to exclude from results processing.
+    If None (the default), will use exclusions as defined in the configuration."""
+    mapping_function: VectorMapper | ScalarMapper | None
```
rmudambi, line 40: Changing this name is not backwards compatible? Have you checked that this won't break downstream repos and models?

patricktnast: changed this back and changed the "new" mapper attribute to `vectorized_mapper`

rmudambi (approved):

hussain-jafari (approved):

PR #490, opened 2024-10-03
Sbachmei/mic 5347/mypy resources

stevebachmeier:
### Description

- *Category*:  other
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5347

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,6 +44,7 @@ if __name__ == "__main__":
         "dill",
         # Type stubs
         "pandas-stubs",
+        "networkx-stubs",
     ]
 
     setup_requires = ["setuptools_scm"]
--- a/src/vivarium/framework/resource.py
+++ b/src/vivarium/framework/resource.py
@@ -1,4 +1,3 @@
-# mypy: ignore-errors
 """
 ===================
 Resource Management
@@ -18,13 +17,18 @@ dependencies or raise exceptions if this is not possible.
 
 """
 
-from types import MethodType
-from typing import Any, Callable, Iterable, List
+from __future__ import annotations
+
+from collections.abc import Iterator
+from typing import TYPE_CHECKING, Any
 
 import networkx as nx
 
 from vivarium.exceptions import VivariumError
 from vivarium.manager import Interface, Manager
+
+if TYPE_CHECKING:
+    from vivarium.framework.engine import Builder
 
 
 class ResourceError(VivariumError):
@@ -59,9 +63,9 @@ class ResourceGroup:
     def __init__(
         self,
         resource_type: str,
-        resource_names: List[str],
-        producer: Callable,
-        dependencies: List[str],
+        resource_names: list[str],
+        producer: Any,
+        dependencies: list[str],
     ):
         self._resource_type = resource_type
         self._resource_names = resource_names
@@ -77,7 +81,7 @@ class ResourceGroup:
         return self._resource_type
 
     @property
-    def names(self) -> List[str]:
+    def names(self) -> list[str]:
         """The long names (including type) of all resources in this group."""
         return [f"{self._resource_type}.{name}" for name in self._resource_names]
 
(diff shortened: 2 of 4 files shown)


src/vivarium/framework/resource.py
```diff
@@ -59,9 +64,9 @@ class ResourceGroup:
     def __init__(
         self,
         resource_type: str,
-        resource_names: List[str],
-        producer: Callable,
-        dependencies: List[str],
+        resource_names: list[str],
+        producer: Callable[..., Any],
```
stevebachmeier, line 68: I'm not 100% clear on this, but I _think_ a producer really can take anything as an arg and return anything? Would like to talk about it w/ someone who is more familiar w this resource stuff.

src/vivarium/framework/resource.py
```diff
@@ -144,24 +149,24 @@ def sorted_nodes(self):
         """
         if self._sorted_nodes is None:
             try:
-                self._sorted_nodes = list(nx.algorithms.topological_sort(self.graph))
+                self._sorted_nodes = list(nx.algorithms.topological_sort(self.graph))  # type: ignore[func-returns-value]
```
stevebachmeier, line 152: This was giving an error that `nx.algorithms.topological_sort` alwasy returns None which simply isn't true. I didn't dig too deep but assume that the stub package I downloaded (which is not by the same folks who maintain networkx) simply typed it wrong.

albrja (approved):

src/vivarium/framework/resource.py
```diff
@@ -250,7 +255,7 @@ def _to_graph(self) -> nx.DiGraph:
                 if dependency not in self._resource_group_map:
                     # Warn here because this sometimes happens naturally
                     # if observer components are missing from a simulation.
-                    self.logger.warning(
+                    self.logger.warning(  # type: ignore[no-untyped-call]
```
stevebachmeier, line 258: untyped from loguru. I tried using the `loguru-mypy` package but it didn't get rid of this particular error wo I just ignore it

setup.py
```diff
@@ -44,6 +44,7 @@
         "dill",
         # Type stubs
         "pandas-stubs",
+        "networkx-stubs",
```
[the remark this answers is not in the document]
stevebachmeier: a stub package for networkx

src/vivarium/framework/resource.py
```diff
@@ -335,7 +340,7 @@ def add_resources(
         """
         self._manager.add_resources(resource_type, resource_names, producer, dependencies)
 
-    def __iter__(self):
+    def __iter__(self) -> Iterator[Callable[..., Any]]:
```
stevebachmeier, line 343: I think this is correct but I'd like to talk about this w/ someone familiar with python's implementation of iterables

src/vivarium/framework/resource.py
```diff
@@ -106,20 +110,20 @@ def __str__(self) -> str:
 class ResourceManager(Manager):
     """Manages all the resources needed for population initialization."""
 
-    def __init__(self):
+    def __init__(self) -> None:
         # This will be a dict with string keys representing the the resource
         # and the resource group they belong to. This is a one to many mapping
         # as some resource groups contain many resources.
-        self._resource_group_map = {}
+        self._resource_group_map: dict[str, ResourceGroup] = {}
```
rmudambi, line 117: Can you convert these inline comments to docstrings?

src/vivarium/framework/resource.py
rmudambi: Is it worth contributing a PR to the stubs package to fix this?

rmudambi (approved):

src/vivarium/framework/resource.py
stevebachmeier: Interesting, I'll take a look. I've never contributed to a 3rd party package before so that would be fun

src/vivarium/framework/resource.py
stevebachmeier: I thought about doing that, but they won't show up b/c they're private anyway. Are you suggesting just so it's easier to know which comment goes w/ wihich attr?

PR #494, opened 2024-10-08
fix mypy typing Values 

patricktnast:
### Description

- *Category*: 
typing
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5343

### Changes and notes
 
mostly adding types.
Refactored Pipeline to use private attributes with property getters and setters wrapped with dynamicvalueerrors if the attribute is getted without being set or set when already set.
### Testing

tests pass


src/vivarium/framework/values.py
```diff
@@ -193,15 +229,66 @@ class Pipeline:
...
+    def manager(self) -> "ValuesManager":
+        if self._manager is None:
+            raise DynamicValueError(self._get_attr_error("manager"))
+        return self._manager
+
+    @manager.setter
+    def manager(self, manager: "ValuesManager") -> None:
+        if self._manager is not None:
+            raise DynamicValueError(self._set_attr_error("manager", manager))
+        self._manager = manager
```
albrja, line 287: Woah this is working because of the different decorators? Didn't know you could do that

albrja (approved):

src/vivarium/framework/values.py
```diff
@@ -52,11 +80,13 @@ def replace_combiner(value: Any, mutator: Callable, *args: Any, **kwargs: Any) -
     -------
         A modified version of the input value.
     """
-    args = list(args) + [value]
-    return mutator(*args, **kwargs)
+    new_args = list(args) + [value]
```
stevebachmeier, line 83: nit: maybe `expanded_args`?

src/vivarium/framework/values.py
```diff
@@ -24,14 +25,41 @@
 from vivarium.manager import Interface, Manager
 from vivarium.types import NumberLike
 
+if TYPE_CHECKING:
+    from vivarium.framework.engine import Builder
+
+
+class ValueSource(Protocol):
```
stevebachmeier, line 32: ooc, what was the mypy error that forced adding these interfaces?

src/vivarium/framework/values.py
```diff
@@ -24,14 +25,41 @@
...
+    def __call__(self, *args: Any, **kwargs: Any) -> NumberLike:
+        ...
+
+
+class ValueMutator(Protocol):
+    def __call__(self, *args: Any, value: NumberLike, **kwargs: Any) -> NumberLike:
```
stevebachmeier, line 38: Can we be more specific w/ typing *args and **kwargs throughout these? (I don't think we can but I haven't thought hard about it yet)

src/vivarium/framework/values.py
```diff
@@ -112,10 +142,16 @@ def rescale_post_processor(value: NumberLike, manager: "ValuesManager") -> Numbe
             axis=0,
         )
     else:
-        return from_yearly(value, manager.step_size())
+        time_step = manager.step_size()
+        if not isinstance(time_step, (pd.Timedelta, timedelta)):
+            raise DynamicValueError(
+                "The rescale post processor requires a time step size that is a "
+                "pandas Timedelta object."
```
stevebachmeier, line 149: or a timedetla obj, no?

src/vivarium/framework/values.py
```diff
@@ -173,7 +209,7 @@ class Pipeline:
     generic components that create a set of pipeline modifiers for
     values that won't be used in the particular simulation.
 
-    Attributes
+    Parameters
```
stevebachmeier, line 212: nit: I think it should be attributes when you're defining the _class_ docstring (but then they'd be Parameters if you were doing this in the init). Like, you're literally saying this class has the attributes ... (public only though, so really it shouldin't include source, combiner, or manager since those are all private). This all becomes tricky in vivarium w/ how many of our attrs are defined in non-init methods and so it's difficult to document.

patricktnast: I normally wouldn't mess with it, but "Attributes" instead of "Parameters" after the subsequent changes caused a doc build error

src/vivarium/framework/values.py
patricktnast: It might be possible to do this more compactly...We have to give the signatures of the Callables; For the value Source I you can make an alias like `Callable[... , NumberLike]` but I didn't know how to do a type alias for something like

Callable[[..., NumberLike, ...], NumberLike] for args and kwargs with a new positional argument in between. Maybe there is an answer in https://peps.python.org/pep-0612/#abstract

src/vivarium/framework/values.py
```diff
@@ -193,15 +229,66 @@ class Pipeline:
...
+        return (
+            f"A second component is attempting to set the {attribute} for pipeline {self.name} "
+            f"with {new_value}, but it already has a {attribute}: {current_value}."
+        )
+
+    @property
+    def source(self) -> ValueSource:
+        if self._source is None:
+            raise DynamicValueError(self._get_attr_error("source"))
+        return self._source
```
rmudambi, line 257: We could reduce the code duplication like this?

```python
T = TypeVar("T")

def _get_property(self, property: T | None, property_name: str) -> T:
    if property is None:
        raise DynamicValueError(self._get_attr_error(property_name))
    return property

def _set_property(self, property_name: str, new_value: Any) -> None:
    private_name = f"_{property_name}"
    old_value = getattr(self, private_name)
    if old_value is not None:
        raise DynamicValueError(self._set_attr_error(property_name, old_value, _new_value))
    setattr(self, private_name, new_value)

@property
def source(self) -> ValueSource:
    return self._get_property(self._source, "source")

@source.setter
def source(self, source: ValueSource) -> None:
    _set_property("source", source)
```

src/vivarium/framework/values.py
```diff
@@ -193,15 +229,66 @@ class Pipeline:
...
-        self.name = None
-        self.source = None
-        self.mutators = []
-        self.combiner = None
-        self.post_processor = None
-        self.manager = None
+    def __init__(self) -> None:
+        self.name: str | None = None
+        self._source: ValueSource | None = None
+        self.mutators: list[ValueMutator] = []
```
rmudambi, line 235: This is a mutable default...

src/vivarium/framework/values.py
```diff
@@ -243,22 +326,22 @@ def _call(self, *args, skip_post_processor=False, **kwargs):
...
-        self._pipelines = defaultdict(Pipeline)
+        self._pipelines: dict[str, Pipeline] = defaultdict(Pipeline)
 
     @property
-    def name(self):
+    def name(self) -> str:
         return "values_manager"
 
-    def setup(self, builder):
+    def setup(self, builder: "Builder") -> None:
```
rmudambi, line 344: You don't need the quotes here. It would be good to remove quotes in typing everywhere in this file.

src/vivarium/framework/values.py
```diff
@@ -270,11 +353,11 @@ def setup(self, builder):
         builder.lifecycle.add_constraint(self.register_value_producer, allow_during=["setup"])
         builder.lifecycle.add_constraint(self.register_value_modifier, allow_during=["setup"])
 
-    def on_post_setup(self, _):
+    def on_post_setup(self, _: Any) -> None:
```
rmudambi, line 356:
```suggestion
    def on_post_setup(self, _event: Event) -> None:
```

src/vivarium/framework/values.py
```diff
@@ -298,12 +381,12 @@ def on_post_setup(self, _):
...
         self,
         value_name: str,
-        source: Callable,
-        requires_columns: List[str] = (),
-        requires_values: List[str] = (),
-        requires_streams: List[str] = (),
-        preferred_combiner: Callable = replace_combiner,
-        preferred_post_processor: Callable = None,
+        source: ValueSource,
+        requires_columns: Iterable[str] = (),
```
rmudambi, line 385: Is it the case that for all of these all we do is iterate over them and that the order doesn't matter?

src/vivarium/framework/values.py
```diff
@@ -354,10 +432,10 @@ def _register_value_producer(
     def register_value_modifier(
         self,
         value_name: str,
-        modifier: Callable,
-        requires_columns: List[str] = (),
-        requires_values: List[str] = (),
-        requires_streams: List[str] = (),
+        modifier: ValueSource | ValueMutator,
```
rmudambi, line 435: This can really also be a `ValueSource`?

src/vivarium/framework/values.py
```diff
@@ -414,7 +492,12 @@ def get_value(self, name) -> Pipeline:
         return self._pipelines[name]  # May create a pipeline.
 
     @staticmethod
-    def _convert_dependencies(func, requires_columns, requires_values, requires_streams):
+    def _convert_dependencies(
+        func: ValueSource | ValueMutator,
```
rmudambi, line 496: Shouldn't Pipeline also be in the type union?

src/vivarium/framework/values.py
```diff
@@ -431,20 +514,20 @@ def _convert_dependencies(func, requires_columns, requires_values, requires_stre
         return dependencies
 
     @staticmethod
-    def _get_modifier_name(modifier):
+    def _get_modifier_name(modifier: ValueSource | ValueMutator) -> str:
```
rmudambi, line 517: Same question regarding `ValueSource`

src/vivarium/framework/values.py
```diff
@@ -588,10 +671,10 @@ def register_rate_producer(
     def register_value_modifier(
         self,
         value_name: str,
-        modifier: Callable,
-        requires_columns: List[str] = (),
-        requires_values: List[str] = (),
-        requires_streams: List[str] = (),
+        modifier: ValueSource | ValueMutator,
```
rmudambi, line 674: I guess this is the fundamental place to consider if `ValueSource` is appropriate

.gitignore
```diff
@@ -118,3 +118,4 @@ src/vivarium/_version.py
 
 # macOS
 *.DS_Store
+vivarium.code-workspace
```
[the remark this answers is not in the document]
patricktnast: it's a configuration setting for vscode, i can move that (and remove this)

src/vivarium/framework/values.py
patricktnast: Ah, i tried to go all the way with this but I was missing the TypeVar part, so it was throwing errors. thanks!

src/vivarium/framework/values.py
patricktnast: I think so-- we can use paramSpec, but that would necessitate some other changes.
https://peps.python.org/pep-0612/#abstract

src/vivarium/framework/values.py
```diff
@@ -24,14 +25,41 @@
 from vivarium.manager import Interface, Manager
 from vivarium.types import NumberLike
 
+if TYPE_CHECKING:
+    from vivarium.framework.engine import Builder
+
+
+class ValueSource(Protocol):
+    def __call__(self, *args: Any, **kwargs: Any) -> NumberLike:
```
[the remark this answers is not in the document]
patricktnast: I understand now. I was generalizing from the post processors, but really those are special cases.

src/vivarium/framework/values.py
patricktnast: From the perspective of a protocol, I would say that that is the case:

> If the pipeline has a ``replace_combiner``, the
>             modifier should accept the same arguments as the pipeline source
>             with an additional last positional argument for the results of the
>             previous stage in the pipeline. For the ``list_combiner`` strategy,
>             the pipeline modifiers should have the same signature as the pipeline
>             source.

Though arguably the protocols aren't quite aptly named in this case.

src/vivarium/framework/values.py
patricktnast: I don't think it does--_convert_dependencies only loops over them, and I don't think the order should matter. I didn't look over Resource super closely though. If you think its safer I could type them as Sequence instead?

src/vivarium/framework/values.py
patricktnast: I don't think it is? These aren't arguments to the constructor.

rmudambi (approved):

PR #495, opened 2024-10-08
interpolation typing

hussain-jafari:
### Description
- *Category*: typing
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5341

### Changes and notes
Fix typing in lookup/interpolation.py.

### Testing
Ran interpolation tests.


src/vivarium/framework/lookup/interpolation.py
```diff
@@ -109,9 +109,10 @@ def __call__(self, interpolants: pd.DataFrame) -> pd.DataFrame:
                 interpolants, self.categorical_parameters, self.continuous_parameters
             )
 
+        sub_tables: list[tuple[tuple[Any, ...] | None, pd.DataFrame]]
```
stevebachmeier, line 112: Move this to the top of the file so you only need to type hint this once

src/vivarium/framework/lookup/interpolation.py
```diff
@@ -40,9 +40,9 @@ class Interpolation:
     def __init__(
         self,
         data: pd.DataFrame,
-        categorical_parameters: Union[List[str], Tuple[str, ...]],
+        categorical_parameters: list[str] | tuple[str, ...],
         continuous_parameters: _ParameterType,
-        value_columns: Union[List[str], Tuple[str, ...]],
+        value_columns: list[str] | tuple[str, ...],
```
stevebachmeier, line 45: might also consider using Iterable

src/vivarium/framework/lookup/interpolation.py
```diff
@@ -218,21 +228,21 @@ def check_data_complete(data, continuous_parameters) -> None:
         If a parameter contains non-continuous bins.
     """
 
+    sub_tables: list[tuple[tuple[Any, ...] | None, pd.DataFrame]]
     param_edges = [
-        p[1:] for p in continuous_parameters if isinstance(p, (Tuple, List))
+        p[1:] for p in continuous_parameters if isinstance(p, (tuple, list))
```
stevebachmeier, line 233: Can't each item in `continuous_parameters` _only_ be a tuple or list anyway? (as per the definition of _ParameterType). So I don't think we even need the `if isinstance()` part?

stevebachmeier (changes requested):
This is kind of a tough one to review, but I _think_ there are some inconsistencies here (though I'm confused how mypy even passed so perhaps I'm just wrong)

src/vivarium/framework/lookup/interpolation.py
```diff
@@ -8,13 +7,14 @@
...
+from __future__ import annotations
 
-from typing import List, Tuple, Union
+from typing import Any, Union
 
 import numpy as np
 import pandas as pd
 
-_ParameterType = Union[List[List[str]], List[Tuple[str, str, str]]]
+_ParameterType = Union[list[list[str]], list[tuple[str, str, str]]]
```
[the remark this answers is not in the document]
rmudambi: You definitely can use `|` even when defining types like here.

src/vivarium/framework/lookup/interpolation.py
```diff
@@ -8,13 +7,14 @@
 simulations.
 
 """
+from __future__ import annotations
 
-from typing import List, Tuple, Union
+from typing import Any, Union
```
rmudambi, line 12: When you're done with this file, we shouldn't be importing `Union`

src/vivarium/framework/lookup/interpolation.py
```diff
@@ -66,17 +66,17 @@ def __init__(
         self.extrapolate = extrapolate
         self.validate = validate
 
+        sub_tables: list[tuple[tuple[Any, ...] | None, pd.DataFrame]]
```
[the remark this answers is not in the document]
rmudambi: I think it may be need to be `Hashable` rather than `Any` since it needs to be assigned as a key to a dictionary later?

src/vivarium/framework/lookup/interpolation.py
```diff
@@ -120,22 +121,27 @@ def __call__(self, interpolants: pd.DataFrame) -> pd.DataFrame:
...
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return "Interpolation()"
 
 
-def validate_parameters(data, categorical_parameters, continuous_parameters, value_columns):
+def validate_parameters(
+    data: pd.DataFrame,
+    categorical_parameters: list[str] | tuple[str, ...],
```
rmudambi, line 141: This could be type-hinted as `Iterable`.

src/vivarium/framework/lookup/interpolation.py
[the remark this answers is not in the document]
rmudambi: I'm not sure we need this type. This is only used for `continuous_parameters` and that seems like it could be type-hinted as `list[Sequence[str]]` (note: not `list[Iterable[str]]`, since the internal sequence must be ordered). The input is validated to ensure it matches the appropriate structure (i.e. has exactly 3 elements). 

If you don't remove it, you should definitely make it not be private, since it will appear in the api docs.

src/vivarium/framework/lookup/interpolation.py
```diff
@@ -283,9 +293,9 @@ class Order0Interp:
 
     def __init__(
         self,
-        data,
+        data: pd.DataFrame,
         continuous_parameters: _ParameterType,
-        value_columns: List[str],
+        value_columns: list[str] | tuple[str, ...],
```
rmudambi, line 298: I think you should leave this typed as `list[str]` and convert inputs to a list in the Interpolation `__init__`

docs/nitpick-exceptions
```diff
@@ -22,6 +22,7 @@ py:class loguru._logger.Logger
 # Type aliases don't play nicely with sphinx when you import them
 # elsewhere.  Works fine for static type checker though.  I think this
 # is because sphinx does runtime checks.
+py:class _ParameterType
```
rmudambi, line 25: If you remove the `_ParameterType` as I suggested, make sure to remove this from here.

rmudambi (approved):

stevebachmeier (approved):

albrja (approved):

patricktnast (approved):

PR #498, opened 2024-10-12
fix bugs preventing Machine from functioning

rmudambi:
### Description

- *Category*: bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5372

### Changes and notes
Fix non-functional Machine by:
- creating and initializing the state column
- calling the transition function on time-step
- 

### Testing
Tests pass
Fixed broken state-machine tests that were incorrectly passing
Added new tests for initialization

### DIFF
--- a/src/vivarium/framework/state_machine.py
+++ b/src/vivarium/framework/state_machine.py
@@ -10,17 +10,22 @@ A state machine implementation for use in ``vivarium`` simulations.
 from __future__ import annotations
 
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional, Tuple
+from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Tuple
 
 import numpy as np
 import pandas as pd
 
 from vivarium import Component
+from vivarium.framework.event import Event
 
 if TYPE_CHECKING:
     from vivarium.framework.engine import Builder
-    from vivarium.framework.population import PopulationView
-    from vivarium.types import ClockTime
+    from vivarium.framework.population import PopulationView, SimulantData
+    from vivarium.types import ClockTime, LookupTableData
+
+
+def default_initializer(_builder: Builder) -> LookupTableData:
+    return 0.0
 
 
 def _next_state(
@@ -192,6 +197,16 @@ class State(Component):
     ##############
 
     @property
+    def configuration_defaults(self) -> Dict[str, Any]:
+        return {
+            f"{self.name}": {
+                "data_sources": {
+                    "initialization_weights": self.get_initialization_weights,
+                },
+            },
+        }
+
+    @property
     def model(self) -> str:
         return self._model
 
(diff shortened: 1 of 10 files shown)


src/vivarium/framework/state_machine.py
```diff
@@ -191,6 +196,16 @@ class State(Component):
     # Properties #
     ##############
 
+    @property
+    def configuration_defaults(self) -> Dict[str, Any]:
```
stevebachmeier, line 200: nit: use `dict`

src/vivarium/framework/state_machine.py
```diff
@@ -199,12 +214,18 @@ def model(self) -> str:
     # Lifecycle methods #
     #####################
 
-    def __init__(self, state_id: str, allow_self_transition: bool = False):
+    def __init__(
+        self,
+        state_id: str,
+        allow_self_transition: bool = False,
+        initialization_weights: Callable[[Builder], LookupTableData] = default_initializer,
```
stevebachmeier, line 221: nit: I think `callable` is a built-in now?

tests/framework/components/test_component.py
```diff
@@ -169,9 +169,7 @@ def test_component_initializer_is_registered_and_called_if_defined():
     component = ColumnCreator()
     simulation = InteractiveContext(components=[component])
     population = simulation.get_population()
-    expected_pop_view = pd.DataFrame(
-        {column: 9 for column in component.columns_created}, index=population.index
-    )
+    expected_pop_view = component.get_initial_state(population.index)
```
stevebachmeier, line 172: This feels like it's getting a little too close to using the thing you're trying to test.

tests/framework/test_state_machine.py
```diff
@@ -35,38 +23,120 @@ def test_initialize_allowing_self_transition():
...
+        {"population": {"population_size": 10000}, "randomness": {"key_columns": []}}
+    )
+    state_a = State("a", initialization_weights=lambda _: 0.2)
+    state_b = State("b", initialization_weights=lambda _: 0.8)
+    machine = Machine("state", states=[state_a, state_b])
+    simulation = InteractiveContext(components=[machine], configuration=base_config)
+
+    state = simulation.get_population()["state"]
+    assert np.all(simulation.get_population().state != "start")
+    assert round((state == "a").mean(), 1) == 0.2
```
stevebachmeier, line 47: Any reason not to use `np.isclose`? 0.24 would rown down to 0.2 but is an unacceptable 16% different.

stevebachmeier (approved):

src/vivarium/framework/state_machine.py
rmudambi: `callable(foo)` is a function that returns True if `foo` is callable. It's not a something that can be used in typing.

tests/framework/test_state_machine.py
rmudambi: Mostly laziness (and it allows for smaller populations. The other tests in this module use `round`. We should really use the FuzzyChecker.

albrja (approved):

PR #504, opened 2024-10-15
create get_population_initializers method

rmudambi:
### Description

- *Category*: refactor

### Changes and notes
- removed iterators on `ResourceInterface` and `ResourceManager`
- moved that functionality to `get_population_initializers()`

This refactor makes it easier to track where this method is called, 
since you can't effectively find usages of `__iter__()` (at least not in
PyCharm). It also gives the method a meaningful name.

### Testing
All tests pass

### DIFF
--- a/src/vivarium/framework/resource/manager.py
+++ b/src/vivarium/framework/resource/manager.py
@@ -1,19 +1,7 @@
 """
-===================
-Resource Management
-===================
-
-This module provides a tool to manage dependencies on resources within a
-:mod:`vivarium` simulation. These resources take the form of things that can
-be created and utilized by components, for example columns in the
-:mod:`state table <vivarium.framework.population>`
-or :mod:`named value pipelines <vivarium.framework.values>`.
-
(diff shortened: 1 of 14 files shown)


hussain-jafari (approved):

src/vivarium/framework/resource/manager.py
```diff
@@ -189,20 +188,16 @@ def _to_graph(self) -> nx.DiGraph:
...
         creation time.
         """
-        return iter(
-            [
-                r.producer
-                for r in self.sorted_nodes
-                if r.type in {"column", NULL_RESOURCE_TYPE}
-            ]
-        )
+        return [
```
albrja, line 198: Should we be worried about order?

albrja (approved):

src/vivarium/framework/resource/manager.py
rmudambi: This change doesn't affect the order in which the initializers will be returned.

patricktnast (approved):

PR #505, opened 2024-10-15
add name to Pipeline __init__

rmudambi:
### Description

- *Category*: refactor

### Changes and notes
- Update Pipeline so that it will always have a name
- Fixed some incorrect docs
- Removed property setters that are no longer needed

### Testing
Tests pass
Tests also pass in VPH if I cherry-pick this commit onto main

### DIFF
--- a/src/vivarium/framework/values.py
+++ b/src/vivarium/framework/values.py
@@ -119,7 +119,7 @@ def rescale_post_processor(value: NumberLike, manager: ValuesManager) -> NumberL
         multiplication over like a :mod:`numpy` array or :mod:`pandas`
         data frame.
     manager
-        The ValuesManager for this simulation
+        The ValuesManager for this simulation.
 
     Returns
     -------
--- a/src/vivarium/framework/values.py
+++ b/src/vivarium/framework/values.py
@@ -14,9 +14,9 @@ simulations, see the value system :ref:`concept note <values_concept>`.
 """
 from __future__ import annotations
 
-from collections import defaultdict
+from collections.abc import Callable, Iterable
 from datetime import timedelta
-from typing import TYPE_CHECKING, Any, Callable, Iterable, Protocol, TypeVar
+from typing import TYPE_CHECKING, Any, Protocol, TypeVar
 
 import pandas as pd
 
@@ -118,9 +118,8 @@ def rescale_post_processor(value: NumberLike, manager: ValuesManager) -> NumberL
         Annual rates, either as a number or something we can broadcast
         multiplication over like a :mod:`numpy` array or :mod:`pandas`
         data frame.
-    time_step
-        A pandas time delta representing the size of the upcoming time
-        step.
+    manager
+        The ValuesManager for this simulation
 
     Returns
     -------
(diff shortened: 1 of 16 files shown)


src/vivarium/framework/values.py
```diff
@@ -471,7 +479,9 @@ def get_value(self, name: str) -> Pipeline:
             (frequently just a :class:`pandas.Index` representing the
             simulants).
         """
-        return self._pipelines[name]  # May create a pipeline.
+        pipeline = self._pipelines.get(name) or Pipeline(name)
```
hussain-jafari, line 482: I understand the expected behavior but why isn't this returning a bool?

albrja (approved):

src/vivarium/framework/values.py
rmudambi: https://realpython.com/python-or-operator/#non-boolean-contexts

hussain-jafari: Oh cool didn't know you could do that with or

hussain-jafari (approved):

src/vivarium/framework/values.py
```diff
@@ -203,24 +202,18 @@ class Pipeline:
...
-            A list of callables that directly modify the pipeline source or
-            contribute portions of the value.
-        post_processor
-            An optional final transformation to perform on the combined output of
-            the source and mutators.
-        """
-        self.name: str | None = None
+    def __init__(self, name: str) -> None:
+        self.name: str = name
+        """The name of the value represented by this pipeline."""
```
stevebachmeier, line 207: Maybe I'm forgetting, but isn't the original docstring correct since __init__ is a proper function and we only do this type of docstring for special-case things like properties?

src/vivarium/framework/values.py
```diff
@@ -118,9 +118,8 @@ def rescale_post_processor(value: NumberLike, manager: ValuesManager) -> NumberL
         Annual rates, either as a number or something we can broadcast
         multiplication over like a :mod:`numpy` array or :mod:`pandas`
         data frame.
-    time_step
-        A pandas time delta representing the size of the upcoming time
-        step.
+    manager
+        The ValuesManager for this simulation
```
stevebachmeier, line 122: nit: missing a period

src/vivarium/framework/values.py
```diff
@@ -311,13 +289,45 @@ def _call(self, *args: Any, skip_post_processor: bool = False, **kwargs: Any) ->
     def __repr__(self) -> str:
         return f"_Pipeline({self.name})"
 
+    @classmethod
+    def setup_pipeline(
+        cls,
```
stevebachmeier, line 294: Does mypy not yell at this for `cls` not being type hinted? 

And related, do we need a docstring for the `cls` parameter below?

src/vivarium/framework/values.py
```diff
@@ -471,7 +479,9 @@ def get_value(self, name: str) -> Pipeline:
             (frequently just a :class:`pandas.Index` representing the
             simulants).
         """
-        return self._pipelines[name]  # May create a pipeline.
+        pipeline = self._pipelines.get(name) or Pipeline(name)
+        self._pipelines[name] = pipeline
```
stevebachmeier, line 483: Feels weird to (potentially) get a dict value and then turn right around and re-assign it to the same key (but I also like not adding if/else logic so maybe it's fine?)

stevebachmeier (approved):

src/vivarium/framework/values.py
rmudambi: TBH I don't remember what we decided on officially. I think this is more maintainable, since it puts the docstring next to the entity it is documenting, but happy to revert if we prefer the other way.

src/vivarium/framework/values.py
rmudambi: `cls` is special like `self`. Neither one requires type-hinting or a docstring.

src/vivarium/framework/values.py
rmudambi: Yeah I thought about that for a bit as well. I ultimately decided that I preferred the succinctness of the reassignment, which I think is more readable as well. Since the behavior is identical and the reassignment isn't exactly expensive, I did it this way.

patricktnast (approved):

PR #506, opened 2024-10-16
enable registration of resources using list of entities

rmudambi:
### Description

- *Category*: feature

### Changes and notes
- enable registration of initialization resources using a list of entities
- enable registration of pipeline dependencies using a list of entities
- move imports in component.py that are only needed for type-checking
- update `initialization_requirements` default argument to `[]`
- introduce deprecation warnings when specifying requirements using dict
- Updated state_machine's initialization requirements to use a list

### Testing
All tests pass
Tests also pass in VPH if I cherry-pick this commit onto main

### DIFF
--- a/src/vivarium/component.py
+++ b/src/vivarium/component.py
@@ -23,14 +23,16 @@ from layered_config_tree import ConfigurationError, LayeredConfigTree
 from loguru._logger import Logger
 
 from vivarium.framework.artifact import ArtifactException
-from vivarium.framework.event import Event
-from vivarium.framework.lookup import LookupTable
-from vivarium.framework.population import PopulationError, PopulationView
-from vivarium.types import LookupTableData
+from vivarium.framework.population import PopulationError
 
 if TYPE_CHECKING:
     from vivarium.framework.engine import Builder
-    from vivarium.framework.population import SimulantData
+    from vivarium.framework.event import Event
+    from vivarium.framework.lookup import LookupTable
+    from vivarium.framework.population import PopulationView, SimulantData
+    from vivarium.framework.randomness import RandomnessStream
+    from vivarium.framework.values import Pipeline
+    from vivarium.types import LookupTableData
 
 DEFAULT_EVENT_PRIORITY = 5
 """The default priority at which events will be triggered."""
@@ -233,7 +235,9 @@ class Component(ABC):
         return None
 
     @property
-    def initialization_requirements(self) -> Dict[str, List[str]]:
+    def initialization_requirements(
+        self,
+    ) -> dict[str, list[str]] | list[str | Pipeline | RandomnessStream]:
         """Provides the names of all values required by this component during
         simulant initialization.
 
@@ -244,11 +248,7 @@ class Component(ABC):
             list for a key implies no requirements for that key during
             initialization.
         """
-        return {
-            "requires_columns": [],
-            "requires_values": [],
-            "requires_streams": [],
-        }
+        return []
 
     @property
     def population_view_query(self) -> Optional[str]:
@@ -768,7 +768,7 @@ class Component(ABC):
                 self.post_setup_priority,
             )
 
-    def _register_simulant_initializer(self, builder: "Builder") -> None:
+    def _register_simulant_initializer(self, builder: Builder) -> None:
         """Registers a simulant initializer if this component has defined one.
 
         This method allows the component to initialize simulants if it has its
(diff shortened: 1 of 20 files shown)


src/vivarium/component.py
```diff
@@ -244,11 +248,7 @@ def initialization_requirements(self) -> Dict[str, List[str]]:
             list for a key implies no requirements for that key during
             initialization.
         """
-        return {
-            "requires_columns": [],
-            "requires_values": [],
-            "requires_streams": [],
-        }
+        return []
```
stevebachmeier, line 251: It just occurrs to me that type-hinting dynamic things like this is confusing and I'm kinda surprised that mypy allows this to pass. Like, you're returning an empty list here but the type hints say it can also be a dict which simply isn't handled here.

src/vivarium/component.py
```diff
@@ -233,7 +235,9 @@ def columns_required(self) -> Optional[List[str]]:
         return None
 
     @property
-    def initialization_requirements(self) -> Dict[str, List[str]]:
+    def initialization_requirements(
+        self,
+    ) -> dict[str, list[str]] | list[str | Pipeline | RandomnessStream]:
```
stevebachmeier, line 240: nit: fix up the Returns docstring below

src/vivarium/component.py
```diff
@@ -781,11 +781,22 @@ def _register_simulant_initializer(self, builder: "Builder") -> None:
         builder
             The builder with which to register the initializer.
         """
+        if isinstance(self.initialization_requirements, list):
+            initialization_requirements = {
+                "required_resources": self.initialization_requirements
+            }
+        else:
+            initialization_requirements = self.initialization_requirements
+            self.logger.warning(
```
stevebachmeier, line 790: nit: use proper DeprecationWarning?

```
import warnings

warnings.warn(
    "The dict format for initialization_requirements is deprecated. "
    "You should use provide a list of the required resources.",
    DeprecationWarning,
    stacklevel=2,
)
```

src/vivarium/framework/population/manager.py
```diff
@@ -284,22 +287,54 @@ def register_simulant_initializer(
         requires_streams
             The randomness streams necessary to initialize the simulant
             attributes.
+        required_resources
+            The resources that the initializer requires to run. Strings are
+            interpreted as column names, and Pipelines and RandomnessStreams
```
stevebachmeier, line 292: This confused me for a second. Either

"Strings are interpreted as column names and Pipelines and RandomnessStreams are interpreted as value pipelines and randomness streams, respectively."

or

"Strings are interpreted as column names, Pipelines are interpreted as value pipelines, and RandomnessStreams are interpreted as randomness streams."

src/vivarium/framework/population/manager.py
```diff
@@ -284,22 +287,54 @@ def register_simulant_initializer(
...
+            are interpreted as value pipelines and randomness streams,
         """
-        if isinstance(creates_columns, str):
-            creates_columns = [creates_columns]
+
+        has_individual_requires = requires_columns or requires_values or requires_streams
+
+        if has_individual_requires and required_resources:
```
stevebachmeier, line 298: Just so I understand, this is _only_ a check b/c we are trying to be backward compatible w/ individual resources, right?

If so, then maybe mention that in the ValueError below that providing individual resources is deprecated

src/vivarium/framework/values.py
```diff
@@ -483,26 +490,52 @@ def get_value(self, name: str) -> Pipeline:
...
-                [f"column.{name}" for name in requires_columns]
-                + [f"value.{name}" for name in requires_values]
-                + [f"stream.{name}" for name in requires_streams]
+            # The dependencies of the pipeline itself will have been declared
+            # when the pipeline was registered.
+            return [f"value.{func.name}"]
+
+        if requires_columns or requires_values or requires_streams:
+            self.logger.warning(
+                "Specifying requirements individually is deprecated. You should "
```
stevebachmeier, line 508: DeprecationWarning?

stevebachmeier (approved):

src/vivarium/component.py
rmudambi: You're always allowed to say a function could return more than it actually will. A somewhat silly example, but imagine you have this function:
```python
def foo() -> bool:
    return True
```
This is obviously fine, even though it is not possible for `False` to be returned. A more precise type-hinting would be
```python
def foo() -> Literal[True]:
    return True
```

src/vivarium/component.py
rmudambi: I was going to do that, and then I was reading that you need to wire this up properly to work with loguru, and I didn't want to figure out how to do that. Technically, we probably shouldn't be logging vivarium in this manner since vivarium is a library: https://loguru.readthedocs.io/en/stable/overview.html#suitable-for-scripts-and-libraries

albrja (approved):

src/vivarium/framework/population/manager.py
rmudambi: I'm not doing that here, because this method should only get called via `component._register_simulant_initializer()`, which already produces a `DeprecationWarning`.

PR #512, opened 2024-10-20
refactor resource arguments of add_resources

rmudambi:
### Description

- *Category*: refactor

### Changes and notes
- Consolidated `resource_type` and `resource_names` arguments into an
  iterable of strings and resources
- Ensure that ResourceGroups aren't empty
- Ensure that all Resources in a ResourceGroup are of the same type

### Testing
All tests pass
Added some new tests

### DIFF
--- a/src/vivarium/framework/resource/manager.py
+++ b/src/vivarium/framework/resource/manager.py
@@ -14,13 +14,11 @@ import networkx as nx
 
 from vivarium.framework.resource.exceptions import ResourceError
 from vivarium.framework.resource.group import ResourceGroup
-from vivarium.framework.resource.resource import RESOURCE_TYPES, Resource
+from vivarium.framework.resource.resource import RESOURCE_TYPES, Column, Resource
 from vivarium.manager import Interface, Manager
 
 if TYPE_CHECKING:
     from vivarium.framework.engine import Builder
-    from vivarium.framework.randomness import RandomnessStream
-    from vivarium.framework.values import Pipeline
 
 NULL_RESOURCE_TYPE = "null"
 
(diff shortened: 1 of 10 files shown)


src/vivarium/framework/resource/manager.py
```diff
@@ -137,16 +121,17 @@ def _get_resource_group(
         --------
         :class:`ResourceGroup`
         """
+        resources_ = [Column(r) if isinstance(r, str) else r for r in resources]
         dependencies_ = [Column(d) if isinstance(d, str) else d for d in dependencies]
-        if not resource_names:
+
+        if not resources_:
             # We have a "producer" that doesn't produce anything, but
```
stevebachmeier, line 128: What's an example of a null producer?

tests/framework/resource/test_manager.py
```diff
@@ -41,74 +41,62 @@ def producer(self, _simulant_data: SimulantData) -> None:
...
-    assert group_2.type == NULL_RESOURCE_TYPE
-    assert group_2.names == [f"{NULL_RESOURCE_TYPE}.1"]
-    assert group_2.producer == component.producer
+    assert group_2.type == "null"
+    assert group_2.names == ["null.1"]
+    assert group_2.producer == producer
     assert not group_2.dependencies
 
 
-def test_resource_manager_add_resources_bad_type(manager: ResourceManager) -> None:
```
stevebachmeier, line 79: Is this just not a relevant test anymore?

stevebachmeier (approved):

src/vivarium/framework/resource/manager.py
rmudambi: A NullProducer happens when a component has an `on_initialize_simulants` method, but that method doesn't create any columns. This could happen if the component wanted to initialize some attribute of itself using columns or pipelines created by other components.

tests/framework/resource/test_manager.py
rmudambi: Since there is no string "type" parameter to the method, they can't really provide a bad type.

albrja (approved):

PR #513, opened 2024-10-20
improve distinction between initializers and non-initializers

rmudambi:
### Description

- *Category*: refactor

### Changes and notes
- Replace `producer` argument with optional `initializer` argument
-  Add `is_initializer` property to ResourceGroup
- Add `is_initialized` property to Resource which is True for Columns and
  NullResources and False otherwise.

### Testing
Tests pass
Added a few new tests

### DIFF
--- a/src/vivarium/framework/resource/group.py
+++ b/src/vivarium/framework/resource/group.py
@@ -69,12 +69,11 @@ class ResourceGroup:
     @property
     def names(self) -> list[str]:
         """The long names (including type) of all resources in this group."""
-        return list(self._resources.keys())
+        return list(self._resources)
 
     @property
     def initializer(self) -> Callable[[SimulantData], None]:
         """The method that initializes this group of resources."""
-        # todo see if we can call is_initializer instead
         if self._initializer is None:
             raise ResourceError("This resource group does not have an initializer.")
         return self._initializer
--- a/src/vivarium/framework/resource/group.py
+++ b/src/vivarium/framework/resource/group.py
@@ -56,15 +56,11 @@ class ResourceGroup:
                 "Resource groups with an initializer must have initialized resources."
             )
 
-        self._type = list(produced_resources)[0].resource_type
+        self.type = list(produced_resources)[0].resource_type
+        """The type of resource produced by this resource group's producer."""
         self._resources = {resource.resource_id: resource for resource in produced_resources}
         self._initializer = initializer
         self._dependencies = dependencies
-
-    @property
-    def type(self) -> str:
-        """The type of resource produced by this resource group's producer."""
-        return self._type
 
     @property
     def names(self) -> list[str]:
--- a/src/vivarium/framework/resource/group.py
+++ b/src/vivarium/framework/resource/group.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from collections.abc import Iterable, Iterator
 from typing import Any
 
+from vivarium.framework.resource.exceptions import ResourceError
 from vivarium.framework.resource.resource import Resource
 
 
@@ -20,13 +21,18 @@ class ResourceGroup:
 
     def __init__(
         self,
-        resource_type: str,
-        resource_names: list[str],
+        produced_resources: Iterable[Resource],
         producer: Any,
         dependencies: Iterable[Resource] = (),
     ):
-        self._resource_type = resource_type
-        self._resource_names = resource_names
+        if not produced_resources:
+            raise ResourceError("Resource groups must have at least one resource.")
+
+        if len(set(r.resource_type for r in produced_resources)) != 1:
+            raise ResourceError("All produced resources must be of the same type.")
+
+        self._resources = list(produced_resources)
+        """The resources produced by this resource group's producer."""
         self._producer = producer
         """The method or object that produces this group of resources."""
         self._dependencies = dependencies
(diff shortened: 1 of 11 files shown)


src/vivarium/framework/resource/group.py
```diff
@@ -22,42 +25,70 @@ class ResourceGroup:
...
     def type(self) -> str:
         """The type of resource produced by this resource group's producer."""
-        return self._resources[0].resource_type
+        return self._type
 
     @property
     def names(self) -> list[str]:
         """The long names (including type) of all resources in this group."""
-        return [resource.resource_id for resource in self._resources]
+        return list(self._resources.keys())
```
stevebachmeier, line 72: nit: don't need the `.keys()`

src/vivarium/framework/resource/group.py
```diff
@@ -22,42 +25,70 @@ class ResourceGroup:
...
 
     @property
-    def producer(self) -> Any:
-        """The method or object that produces this group of resources."""
-        return self._producer
+    def initializer(self) -> Callable[[SimulantData], None]:
+        """The method that initializes this group of resources."""
+        # todo see if we can call is_initializer instead
+        if self._initializer is None:
+            raise ResourceError("This resource group does not have an initializer.")
```
stevebachmeier, line 79: Do you foresee this ever being a problem? Like if iterating through all initializers or something like that?

src/vivarium/framework/resource/group.py
```diff
@@ -22,42 +25,70 @@ class ResourceGroup:
...
     def names(self) -> list[str]:
         """The long names (including type) of all resources in this group."""
-        return [resource.resource_id for resource in self._resources]
+        return list(self._resources.keys())
 
     @property
-    def producer(self) -> Any:
-        """The method or object that produces this group of resources."""
-        return self._producer
+    def initializer(self) -> Callable[[SimulantData], None]:
```
stevebachmeier, line 75: Is the name change just to get away from overloading the term?

src/vivarium/framework/resource/manager.py
```diff
@@ -73,47 +74,52 @@ def sorted_nodes(self) -> list[ResourceGroup]:
     def setup(self, builder: Builder) -> None:
         self.logger = builder.logging.get_logger(self.name)
 
-    # TODO [MIC-5380]: Refactor add_resources for better type hinting
```
stevebachmeier, line 76: 👍

src/vivarium/framework/resource/manager.py
```diff
@@ -73,47 +74,52 @@ def sorted_nodes(self) -> list[ResourceGroup]:
...
-
-        for resource in resource_group:
-            if resource in self._resource_group_map:
-                other_producer = self._resource_group_map[resource].producer
+        resource_group = self._get_resource_group(resources, dependencies, initializer)
+
+        for resource_id in resource_group:
+            if resource_id in self._resource_group_map:
+                other_resource = self._resource_group_map[resource_id]
+                if resource_group.is_initializer:
```
stevebachmeier, line 107: This added logic seems like it is a bugfix? Were we previously able to have a non-initializing resource that would result in an unwanted raise?

src/vivarium/framework/resource/resource.py
```diff
@@ -17,16 +17,33 @@ def resource_id(self) -> str:
         """The long name of the resource, including the type."""
         return f"{self.resource_type}.{self.name}"
 
+    # TODO [MIC_TBD]: make this an abstract method when support for old
```
stevebachmeier, line 20: nit: don't forget to create this ticket and add MIC #

src/vivarium/framework/resource/resource.py
```diff
@@ -17,16 +17,33 @@ def resource_id(self) -> str:
...
+    #  requirements specification is dropped
+    @property
+    def is_initialized(self) -> bool:
+        """Return True if the resource needs to be initialized."""
+        return False
```
stevebachmeier, line 25: I guess I don't understand the `is_initialized` logic. How can this always be false for any Resource (besides Null an Columns)?

src/vivarium/framework/resource/group.py
rmudambi: I thought it was more consistent terminology. We already talk about simulant initializers elsewhere, and that's what these are.

src/vivarium/framework/resource/resource.py
rmudambi: Initializers are always a component's `on_initialize_simulants` method. These are iterated over during simulation initialization (and are the entire reason this resource management system is required). Pipelines and RandomnessStreams don't need to be initialized - they just require certain resources to be present in order for them to be called.

src/vivarium/framework/resource/group.py
rmudambi: This is just so we can type hint this as not being optional

src/vivarium/framework/resource/manager.py
rmudambi: This is to do with the fact that now not all resources will have an initializer (whereas preciously they all have producers). I just had an idea about how to improve this by passing the component that I will try out and maybe put up in a new PR.

stevebachmeier (approved):

albrja (approved):

PR #514, opened 2024-10-21
refactor to create values package

rmudambi:
### Description

- *Category*: refactor

### Changes and notes
- Refactor to create values package

This is all just copy/paste 

### Testing
All tests pass


src/vivarium/framework/values/pipeline.py
```diff
@@ -0,0 +1,171 @@
...
+
+from vivarium.framework.resource import Resource
+from vivarium.framework.values.exceptions import DynamicValueError
+
+if TYPE_CHECKING:
+    from vivarium.framework.values.combiners import ValueCombiner
+    from vivarium.framework.values.manager import ValuesManager
+    from vivarium.framework.values.post_processors import PostProcessor
+
+T = TypeVar("T")
```
albrja, line 16: What is this for?

src/vivarium/framework/values/post_processors.py
```diff
@@ -0,0 +1,101 @@
...
+    Returns
+    -------
+        The annual rates rescaled to the size of the current time step size.
+    """
+    if isinstance(value, (pd.Series, pd.DataFrame)):
+        return value.mul(
+            manager.simulant_step_sizes(value.index)
+            .astype("timedelta64[ns]")
+            .dt.total_seconds()
+            / (60 * 60 * 24 * 365.0),
```
albrja, line 46: I thought we used 365.25 for days in a year.

albrja: Ok I am wrong there is only one instance where we do that here: https://github.com/ihmeuw/vivarium/blob/main/src/vivarium/framework/lookup/table.py#L179 but everywhere else is 365.0

albrja (approved):

src/vivarium/framework/values/pipeline.py
rmudambi: Take a look at `_get_property()` below. It has an input type of `T | None` and a return type of `T`. This means that can take any type or None as an input and return exactly the same type that was input. i.e. if type `Foo` was passed in, you're guaranteed to get `Foo` out.

hussain-jafari (approved):

patricktnast (approved):

PR #519, opened 2024-10-29
table typing

hussain-jafari:
### Description
- *Category*: typing
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5396

### Changes and notes
Typing updates in framework/lookup/table.py

### Testing
Tests and mypy pass.


src/vivarium/framework/lookup/table.py
```diff
@@ -136,16 +139,23 @@ def __init__(
 
         self.parameter_columns_with_edges = param_cols_with_edges
 
-        extra_columns = self.data.columns.difference(
+        if not isinstance(self.data, pd.DataFrame):
```
albrja, line 142: Why did you add this? Is this so mypy knowns data is a dataframe?

src/vivarium/framework/lookup/table.py
```diff
@@ -237,6 +260,11 @@ def call(self, index: pd.Index) -> pd.DataFrame:
         -------
             A table with the mapped values for the population requested.
         """
+        if not isinstance(self.data, pd.DataFrame):
```
albrja, line 263: Same question as above.

src/vivarium/framework/lookup/table.py
```diff
@@ -136,16 +139,23 @@ def __init__(
...
             | set([col for p in self.parameter_columns_with_edges for col in p])
             | set(self.value_columns)
         )
+        extra_columns = self.data.columns.difference(list(required_cols))
 
         if not self.value_columns:
             self.value_columns = list(extra_columns)
         else:
-            self.data = self.data.drop(columns=extra_columns)
+            if isinstance(self.data, pd.DataFrame):
```
rmudambi, line 157: This could be `elif`, but if you are throwing if `self.data` is not a DataFrame above, you don't need this check here.

src/vivarium/framework/lookup/table.py
```diff
@@ -42,20 +44,20 @@ class LookupTable(ABC):
...
-    parameter_columns: Union[List[str], Tuple] = ()
+    parameter_columns: list[str] | tuple[str, ...] = ()
     """Column names to be used as continuous parameters in Interpolation."""
-    value_columns: Union[List[str], Tuple[str]] = ()
+    value_columns: list[str] | tuple[str, ...] = ()
     """Names of value columns to be interpolated over."""
     interpolation_order: int = 0
     """Order of interpolation. Used to decide interpolation strategy."""
-    clock: Callable = None
+    clock: Callable[[], ClockTime] | None = None
```
rmudambi, line 60: `clock` is only not None if it's an `InterpolatedTable`. It should probably be an attribute of `InterpolatedTable`, not `LookupTable`. This will remove the need for some null checks later.

src/vivarium/framework/lookup/table.py
```diff
@@ -173,11 +183,19 @@ def call(self, index: pd.Index) -> pd.DataFrame:
         """
         pop = self.population_view.get(index)
         del pop["tracked"]
+        if self.clock is None:
```
rmudambi, line 186: If `clock` is defined as an attribute of `InterpolatedTable`, it can be type-hinted as not being `Optional`, so this check would not be necessary.

src/vivarium/framework/lookup/table.py
```diff
@@ -173,11 +183,19 @@ def call(self, index: pd.Index) -> pd.DataFrame:
...
-            fractional_year += current_time.timetuple().tm_yday / 365.25
-            pop["year"] = fractional_year
+            # TODO: [MIC-5478] handle Number output from clock
+            if isinstance(current_time, pd.Timestamp) or isinstance(current_time, datetime):
+                fractional_year: float = float(current_time.year)
+                fractional_year += current_time.timetuple().tm_yday / 365.25
+                pop["year"] = fractional_year
+            else:
+                raise ValueError(
+                    "The output of your clock is incompatible with your simulation data."
```
rmudambi, line 197: Can you improve this error message? Maybe something like "You cannot use the column 'year' in a simulation unless your simulation uses a DateTimeClock."

src/vivarium/framework/lookup/table.py
```diff
@@ -173,11 +183,19 @@ def call(self, index: pd.Index) -> pd.DataFrame:
...
+        if self.clock is None:
+            raise ValueError("An interpolated table must be initialized with a clock.")
         if "year" in [col for col in self.parameter_columns]:
             current_time = self.clock()
-            fractional_year = current_time.year
-            fractional_year += current_time.timetuple().tm_yday / 365.25
-            pop["year"] = fractional_year
+            # TODO: [MIC-5478] handle Number output from clock
+            if isinstance(current_time, pd.Timestamp) or isinstance(current_time, datetime):
+                fractional_year: float = float(current_time.year)
```
rmudambi, line 192: Do you need the `:float` since you're calling `float(...)`?

src/vivarium/framework/lookup/table.py
```diff
@@ -200,11 +218,11 @@ class CategoricalTable(LookupTable):
     def __init__(
         self,
         table_number: int,
-        data: Union[ScalarValue, pd.DataFrame, List[ScalarValue], Tuple[ScalarValue]],
-        population_view_builder: Callable,
-        key_columns: Union[List[str], Tuple[str]],
-        value_columns: Union[List[str], Tuple[str]],
-        **kwargs,
+        data: pd.DataFrame,
+        population_view_builder: Callable[..., PopulationView],
```
rmudambi, line 222: This doesn't seem restrictive enough? We're calling this a few lines later, and it is taking a list of strings.

src/vivarium/framework/lookup/table.py
```diff
@@ -42,20 +44,20 @@ class LookupTable(ABC):
...
     table_number: int
     """Unique identifier of the table."""
-    data: Union[ScalarValue, pd.DataFrame, List[ScalarValue], Tuple[ScalarValue]]
+    data: ScalarValue | pd.DataFrame | list[ScalarValue] | tuple[ScalarValue]
     """The data from which to build the interpolation."""
-    population_view_builder: Callable = None
+    population_view_builder: Callable[..., PopulationView] | None = None
     """Callable to get a population view to be used by the lookup table."""
-    key_columns: Union[List[str], Tuple[str]] = ()
+    key_columns: list[str] | tuple[str, ...] = ()
```
rmudambi, line 51: Do we want these to be Sequence[str]?

rmudambi (approved):

albrja (approved):

PR #521, opened 2024-10-29
Maintain lookup table value column order

stevebachmeier:
### Description

- *Category*:  bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5205

### Testing

added a new test and it passes

### DIFF
--- a/src/vivarium/component.py
+++ b/src/vivarium/component.py
@@ -11,9 +11,10 @@ simulations.
 
 import re
 from abc import ABC
+from collections.abc import Sequence
 from importlib import import_module
 from inspect import signature
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Sequence, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Union
 
 import pandas as pd
 from layered_config_tree import ConfigurationError, LayeredConfigTree
@@ -629,7 +630,7 @@ class Component(ABC):
         return builder.lookup.build_table(data)
 
     def _get_columns(
-        self, value_columns: Optional[Sequence[str]], data: float | pd.DataFrame
+        self, value_columns: Sequence[str] | None, data: float | pd.DataFrame
     ) -> tuple[list[str], list[str], list[str]]:
         all_columns = list(data.columns)
         if value_columns is None:
--- a/src/vivarium/component.py
+++ b/src/vivarium/component.py
@@ -13,7 +13,7 @@ import re
 from abc import ABC
 from importlib import import_module
 from inspect import signature
-from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Sequence, Union
 
 import pandas as pd
 from layered_config_tree import ConfigurationError, LayeredConfigTree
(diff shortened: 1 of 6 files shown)


tests/helpers.py
```diff
@@ -183,9 +183,46 @@ def load_baking_time(_builder: Builder) -> float:
 
 
 class SingleLookupCreator(ColumnCreator):
+    pass
+
+
+class OrderedColumnsLookupCreator(Component):
     @property
-    def standard_lookup_tables(self) -> List[str]:
```
stevebachmeier, line 187: it didn't seem to be using this?

src/vivarium/component.py
```diff
@@ -610,31 +610,38 @@ def build_lookup_table(
         if isinstance(data, list):
             return builder.lookup.build_table(data, value_columns=list(value_columns))
         if isinstance(data, pd.DataFrame):
-            all_columns = set(data.columns)
+            duplicated_columns = set(data.columns[data.columns.duplicated()])
+            if duplicated_columns:
```
albrja, line 614: Nice

src/vivarium/component.py
```diff
@@ -610,31 +610,38 @@ def build_lookup_table(
...
-                    bin_edge_columns.update([f"{column}_start", f"{column}_end"])
+                    parameter_columns.append(column)
+                    bin_edge_columns += [f"{column}_start", f"{column}_end"]
 
-            key_columns = all_columns - value_columns - bin_edge_columns
+            key_columns = [
+                col
+                for col in all_columns
+                if col not in value_columns and col not in bin_edge_columns
+            ]
```
albrja, line 638: Aren't lists not ordered? So if weh ave a different seed they could be in a different order?

albrja: I'd also say make all this logic for getting the different columns a helper function/

src/vivarium/framework/lookup/manager.py
```diff
@@ -64,9 +64,9 @@ def setup(self, builder: "Builder") -> None:
     def build_table(
         self,
         data: LookupTableData,
-        key_columns: Union[List[str], Tuple[str, ...]],
-        parameter_columns: Union[List[str], Tuple[str, ...]],
-        value_columns: Union[List[str], Tuple[str, ...]],
+        key_columns: Sequence[str],
```
albrja, line 67: I'm confused how this is a Sequence if you originally defined it as [] but I haven't really used Sequence.

tests/helpers.py
albrja: I don't remember what this was for but Rajan and I had tried to add some feature like this but I think this is deprecated now.

src/vivarium/component.py
stevebachmeier: lists are ordered

src/vivarium/framework/lookup/manager.py
stevebachmeier: ultimately it can be any sequence (i.e. list or tuple). But it's kinda weird b/c as you said we instantiate it specifically as a list. But we were _already_ saying `Union[list, tuple]` anyway! So I dunno - I think we're just a little inconsistent here in general

hussain-jafari (approved):

albrja (approved):

src/vivarium/component.py
```diff
@@ -610,35 +610,51 @@ def build_lookup_table(
...
-                value_columns=list(value_columns),
+                key_columns=key_columns,
+                parameter_columns=parameter_columns,
+                value_columns=value_columns,
             )
 
         return builder.lookup.build_table(data)
 
+    def _get_columns(
+        self, value_columns: Optional[Sequence[str]], data: float | pd.DataFrame
```
rmudambi, line 632: Don't use Optional

rmudambi (approved):

PR #526, opened 2024-11-01
Albrja/Mic-5462/mypy-results-observation

albrja:
### Remove mypy errors from results/observation.py
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5472

### DIFF
--- a/src/vivarium/framework/results/observation.py
+++ b/src/vivarium/framework/results/observation.py
@@ -139,8 +139,13 @@ class UnstratifiedObservation(BaseObservation):
         to_observe: Callable[[Event], bool] = lambda event: True,
     ):
         def _wrap_results_gatherer(
-            df: pd.DataFrame, _: tuple[str, ...] | None
+            df: pd.DataFrame | DataFrameGroupBy[str], _: tuple[str, ...] | None
         ) -> pd.DataFrame:
+            if isinstance(df, DataFrameGroupBy):
+                raise TypeError(
+                    "Must provide a dataframe to an UnstratifiedObservation. "
+                    f"Provided DataFrameGroupBy instead."
+                )
             return results_gatherer(df)
 
         super().__init__(
@@ -148,7 +153,7 @@ class UnstratifiedObservation(BaseObservation):
             pop_filter=pop_filter,
             when=when,
             results_initializer=self.create_empty_df,
-            results_gatherer=_wrap_results_gatherer,  # type: ignore [arg-type]
+            results_gatherer=_wrap_results_gatherer,
             results_updater=results_updater,
             results_formatter=results_formatter,
             stratifications=None,
--- a/src/vivarium/framework/results/observation.py
+++ b/src/vivarium/framework/results/observation.py
@@ -22,7 +22,7 @@ from __future__ import annotations
 
 import itertools
 from abc import ABC
-from collections.abc import Callable, Iterable, Sequence
+from collections.abc import Callable
 from dataclasses import dataclass
 from typing import Any
 
--- a/src/vivarium/framework/results/observation.py
+++ b/src/vivarium/framework/results/observation.py
@@ -19,10 +19,12 @@ provided as attributes to the parent class.
 
 """
 
+from __future__ import annotations
+
 import itertools
 from abc import ABC
 from dataclasses import dataclass
-from typing import Callable, Iterable, Optional, Sequence, Tuple, Union
+from typing import Callable, Iterable, Sequence
 
 import pandas as pd
 from pandas.api.types import CategoricalDtype
@@ -57,9 +59,8 @@ class BaseObservation(ABC):
     prior to starting the simulation. This could return, for example, an empty
     DataFrame or one with a complete set of stratifications as the index and
     all values set to 0.0."""
-    results_gatherer: Union[
-        Callable[[pd.DataFrame, Sequence[str]], pd.DataFrame],
-        Callable[[pd.DataFrame], pd.DataFrame],
+    results_gatherer: Callable[[pd.DataFrame, Sequence[str]], pd.DataFrame] | Callable[
+        [pd.DataFrame], pd.DataFrame
     ]
     """Method or function that gathers the new observation results."""
     results_updater: Callable[[pd.DataFrame, pd.DataFrame], pd.DataFrame]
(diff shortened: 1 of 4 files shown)


src/vivarium/framework/results/observation.py
```diff
@@ -19,10 +19,12 @@
 
 """
 
+from __future__ import annotations
+
 import itertools
 from abc import ABC
 from dataclasses import dataclass
-from typing import Callable, Iterable, Optional, Sequence, Tuple, Union
+from typing import Callable, Iterable, Sequence
```
rmudambi, line 27: These should be imported from `collections.abc`

rmudambi (changes requested):
You don't see any errors because you forgot to remove the first line in the file, so all errors are being skipped.

src/vivarium/framework/results/observation.py
```diff
@@ -140,12 +139,17 @@ def __init__(
...
+        ) -> pd.DataFrame:
+            return results_gatherer(df)
+
         super().__init__(
             name=name,
             pop_filter=pop_filter,
             when=when,
             results_initializer=self.create_empty_df,
-            results_gatherer=results_gatherer,
+            results_gatherer=_wrap_results_gatherer,  # type: ignore [arg-type]
```
albrja, line 152: This is being ignored because this creates a results gatherer that is a `Callable[[pd.DataFrame], pd.DataFrame]` where mypy thinks the typing is either a dataframe or a groupby and then the second argument is an optional tuple. We however know that it is one case or the other and have implemented the wrapper above to handle this.

Alternatively, the options I do not like is we can have the two separate callables as type hints in the base class and then we would need to ignore arg type in the observe method where we call the results gatherer because mypy does not know if df is a pd.DataFrame or groupby object.

src/vivarium/framework/results/observation.py
```diff
@@ -214,17 +218,17 @@ def __init__(
...
+        aggregator: Callable[[pd.DataFrame], float | pd.Series[float]],
         to_observe: Callable[[Event], bool] = lambda event: True,
     ):
         super().__init__(
             name=name,
             pop_filter=pop_filter,
             when=when,
             results_initializer=self.create_expanded_df,
-            results_gatherer=self.get_complete_stratified_results,
+            results_gatherer=self.get_complete_stratified_results,  # type: ignore [arg-type]
```
albrja, line 231: Same comment as above

src/vivarium/framework/results/observation.py
```diff
@@ -313,22 +317,22 @@ def get_complete_stratified_results(
...
+    ) -> pd.Series[float] | pd.DataFrame:
         """Apply the `aggregator` to the population groups and their
         `aggregator_sources` columns.
         """
         aggregates = (
-            pop_groups[aggregator_sources].apply(aggregator).fillna(0.0)
+            pop_groups[aggregator_sources].apply(aggregator).fillna(0.0)  # type: ignore [arg-type]
             if aggregator_sources
-            else pop_groups.apply(aggregator)
+            else pop_groups.apply(aggregator)  # type: ignore [arg-type]
```
albrja, line 330: These two ignores are cause by the DataFrameGroupBy objects needing a parameter type hint. In reality, it isn't any but it is a lot of things and so I am using any. I also chose to ignore it here so the typing for the function is a little more help to users instead of having an arg and the return type be Any.

src/vivarium/framework/results/observation.py
rmudambi: What if you type-hinted your wrapped method like this:
```python
def _wrap_results_gatherer(
    df: pd.DataFrame | DataFrameGroupBy[Any], _: tuple[str, ...] | None
) -> pd.DataFrame:
    if isinstance(df, DataFrameGroupBy):
        raise TypeError(...)
    return results_gatherer(df)
```

src/vivarium/framework/results/observation.py
```diff
@@ -52,32 +54,32 @@ class BaseObservation(ABC):
...
     prior to starting the simulation. This could return, for example, an empty
     DataFrame or one with a complete set of stratifications as the index and
     all values set to 0.0."""
-    results_gatherer: Union[
-        Callable[[pd.DataFrame, Sequence[str]], pd.DataFrame],
-        Callable[[pd.DataFrame], pd.DataFrame],
+    results_gatherer: Callable[
+        [pd.DataFrame | DataFrameGroupBy[Any], tuple[str, ...] | None], pd.DataFrame
     ]
+    # results_gatherer: Callable[[DataFrameGroupBy[Any], tuple[str, ...]], pd.DataFrame] | Callable[[pd.DataFrame, tuple[str, ...] | None], pd.DataFrame]
```
stevebachmeier, line 65: delete

src/vivarium/framework/results/observation.py
stevebachmeier: @rmudambi Is it reasonable to type hint the DataFrameGroupBys as `DataFrameGroupBy[str]`? I'm trying to think of the types of things we stratify by and they're all strings (age group, sex, location, disease states, etc). But I guess we might one day want to stratify by other things?

Or maybe `DataFrameGroupBy[str, int]` just to allow for IDs?

src/vivarium/framework/results/observation.py
```diff
@@ -340,10 +344,11 @@ def _format(aggregates: Union[pd.Series, pd.DataFrame]) -> pd.DataFrame:
     @staticmethod
     def _expand_index(aggregates: pd.DataFrame) -> pd.DataFrame:
         """Include all stratifications in the results by filling missing values with 0."""
-        if isinstance(aggregates.index, pd.MultiIndex):
-            full_idx = pd.MultiIndex.from_product(aggregates.index.levels)
-        else:
-            full_idx = aggregates.index
+        full_idx = (
```
stevebachmeier, line 347: so silly this works but the original didn't

stevebachmeier (commented):
Might be able to use str and int for DataFrameGroupBys? Let's talk at standup or midday

rmudambi (approved):
As a reminder, please don't force-push to branches that are already under review. This rewrites the history and makes it harder for a reviewer to keep track of what has changed since their last review.

src/vivarium/framework/results/observation.py
stevebachmeier: @albrja Did Rajan's suggestion just not work?

stevebachmeier (approved):

PR #527, opened 2024-11-04
wire up managers to be able to create columns

rmudambi:
### Description

- *Category*: refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5464

### Changes and notes
- Gives `PluginManager` a name
- Adds abstract `name` property to `Manager`
- Adds `columns_created` property to `Manager`
- Adds `on_initialize_simulants()` method to `Manager`

### Testing
All tests pass

### DIFF
--- a/src/vivarium/framework/plugins.py
+++ b/src/vivarium/framework/plugins.py
@@ -95,6 +95,10 @@ class PluginConfigurationError(VivariumError):
 
 
 class PluginManager(Manager):
+    @property
+    def name(self) -> str:
+        return "plugin_manager"
+
     def __init__(
         self,
         plugin_configuration: (
--- a/src/vivarium/manager.py
+++ b/src/vivarium/manager.py
@@ -7,15 +7,18 @@ A base Manager class to be used to create manager for use in ``vivarium``
 simulations.
 
 """
+from __future__ import annotations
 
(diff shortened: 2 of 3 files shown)


src/vivarium/framework/plugins.py
```diff
@@ -95,6 +95,10 @@ class PluginConfigurationError(VivariumError):
 
 
 class PluginManager(Manager):
+    @property
```
albrja, line 98: Do managers not have the same name parsing the component has?

src/vivarium/manager.py
```diff
@@ -26,25 +29,57 @@ class Manager:
     ##############
 
     @property
-    def configuration_defaults(self) -> Dict[str, Any]:
+    @abstractmethod
```
albrja, line 32: So is there a reason we wouldn't want to have the same name parsing as what is in component so we know any NAMEManager has the property name_manager?

albrja (approved):

src/vivarium/manager.py
rmudambi: I just didn't want to make this more complicated. We could definitely introduce that.

stevebachmeier (approved):

PR #528, opened 2024-11-04
specify the component that creates resources and resource groups

rmudambi:
### Description

- *Category*: refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5458

### Changes and notes
Apologies for the large PR. I couldn't figure out how to break it up.

- Instead of passing the initializer method when adding a resource,
  pass the component/manager, since we know the initializer will be
  `component.on_initialize_simulants()`
- In this vein, adds component to Pipeline and RandomnessStream
- Gets the initializer of a `ResourceGroup` from its resources (which
  all must have the same component and type).
- Improve error message when multiple components try to create the
  same resource.
- Added methods `set_attributes()` and `get_value_modifier()` to simplify
  creation of pipelines and their modifiers.
- 

### Testing
Added tests of new functionality
Ensured all tests pass

### DIFF
--- a/src/vivarium/framework/resource/resource.py
+++ b/src/vivarium/framework/resource/resource.py
@@ -17,8 +17,9 @@ class Resource:
     """The type of the resource."""
     name: str
     """The name of the resource."""
+    # todo why are we attaching the component to the resource group instead of the resource itself?
     resource_group: ResourceGroup | None = None
-    """The component that produces the resource."""
+    """The resource group of the resource."""
 
     @property
     def resource_id(self) -> str:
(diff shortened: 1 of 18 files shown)


src/vivarium/framework/resource/resource.py
```diff
@@ -11,14 +16,15 @@ class Resource:
     """The type of the resource."""
     name: str
     """The name of the resource."""
+    # TODO [MIC-5452]: all resources should have a component
+    component: Component | Manager | None
```
albrja, line 20: Why is this an optional argument if don't want it to be optional?

stevebachmeier (approved):

hussain-jafari (approved):

PR #533, opened 2024-11-08
fix broken vph tests

rmudambi:
### Description

- *Category*: bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5403

### Changes and notes
- Move get_data data validation into the get lookup table function 
- Simplify default arguments for state initialization weights
- Add hash function to Pipeline

### Testing
All tests pass
All VPH tests pass without any VPH changes

### DIFF
--- a/src/vivarium/framework/state_machine.py
+++ b/src/vivarium/framework/state_machine.py
@@ -502,12 +502,6 @@ class Machine(Component):
     # Lifecycle methods #
     #####################
 
-    # todo:
-    #  get VPH builds passing with the fewest possible changes
-    #  introduce residual_state arg to DiseaseModel and deprecate usage of initial_state
-    #  change meaning of initial_state in DiseaseModel to match the usage in Machine
-    #  move functionality from VPH to Vivarium
-
     def __init__(
         self,
         state_column: str,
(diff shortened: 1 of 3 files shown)


src/vivarium/framework/state_machine.py
```diff
@@ -516,7 +521,7 @@ def __init__(
             self.add_states(states)
 
         states_with_initialization_weights = [
-            s for s in self.states if s.initialization_weights != default_initializer
+            s for s in self.states if s.initialization_weights
```
stevebachmeier, line 524: Give a more descriptive variable name. Or at the _least_ change it away from "s" b/c that collides w/ pdb's "s" command.

src/vivarium/framework/values/pipeline.py
```diff
@@ -186,6 +186,9 @@ def _call(self, *args: Any, skip_post_processor: bool = False, **kwargs: Any) ->
     def __repr__(self) -> str:
         return f"_Pipeline({self.name})"
 
+    def __hash__(self) -> int:
```
stevebachmeier, line 189: What's this going to be used for?

stevebachmeier (approved):

src/vivarium/framework/values/pipeline.py
rmudambi: This makes sure pipelines remain hashable, which they were previously. Since pipelines are guaranteed to have unique names in a given simulation, this seems sufficient to disambiguate for a hash function. Pipelines not being hashable breaks some VPH components which want to put pipelines into a set.

albrja (approved):

PR #538, opened 2024-11-13
Sbachmei/mic 5549/mypy results context

stevebachmeier:
### Description

- *Category*:  other
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5549


src/vivarium/framework/results/context.py
```diff
@@ -356,7 +356,7 @@ def _get_groups(
             )
         else:
             pop_groups = filtered_pop.groupby(lambda _: "all")
-        return pop_groups
+        return pop_groups  # type: ignore[return-value]
```
stevebachmeier, line 359: Happy to talk about this. I think it's a discrepancy between mypy requiring DataFrameGroupBys to be generic and python RuntimeErroring w/ DataFrameGroupBy is not subscriptable. `from __future__ import annotations` does not fix like it does for pd.Series.

src/vivarium/framework/results/observation.py
```diff
@@ -35,6 +35,9 @@
 
 VALUE_COLUMN = "value"
 
+if TYPE_CHECKING:
+    _PandasGroup = pd.DataFrame | DataFrameGroupBy[tuple[str, ...] | str]
```
stevebachmeier, line 39: Figures this isn't used anywhere else so just made it private here instead of adding to types.py

src/vivarium/framework/results/manager.py
```diff
@@ -301,7 +302,7 @@ def _bin_data(data: Union[pd.Series, pd.DataFrame]) -> pd.Series:
 
     def register_observation(
         self,
-        observation_type,
+        observation_type: Type[BaseObservation],
```
stevebachmeier, line 305: This wasn't causing an error but it was just wrong. At this point we're sending in the class itself, NOT an instantiation of it (actually, it might be worth just saying `observation_type: Type[StratifiedObservation | UnstratifiedObservation | AddingObservation | ConcatenatingObservation]` if folks prefer b/c that's what it really is. But they all inherit from BaseObservation so this is less typing.

rmudambi: Should `BaseObservation` be called `Observation`?

src/vivarium/framework/results/observation.py
```diff
@@ -60,7 +59,8 @@ class BaseObservation(ABC):
     DataFrame or one with a complete set of stratifications as the index and
     all values set to 0.0."""
     results_gatherer: Callable[
-        [pd.DataFrame | DataFrameGroupBy[str], tuple[str, ...] | None], pd.DataFrame
```
rmudambi, line 63: I'm assuming the things in this file were just type-hinted incorrectly?

rmudambi (approved):

src/vivarium/framework/results/manager.py
stevebachmeier: I think we had it called that at once and changed it to this but now can't recall why. It does subclass from abc, though there are no abstract methods or anything on it. Probably doesn't matter?

src/vivarium/framework/results/observation.py
stevebachmeier: Yeah, though I'm not gonna lie - I'm having trouble understanding how to type-hint DataFrameGroupBys...

src/vivarium/framework/results/manager.py
patricktnast: Type[C] is covaiant, so I think putting in the base observation makes sense. It looks like there is a builtin `type` you could use instead, though

PR #542, opened 2024-11-14
allow users to define initialization weights as LookupTableData or an…

rmudambi:
… artifact key

### Description

- *Category*: feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5580

### Changes and notes
- Allow users to define initialization weights as LookupTableData or an artifact key

### Testing
Tests pass


albrja (approved):

src/vivarium/framework/state_machine.py
```diff
@@ -236,6 +239,20 @@ def __init__(
     # Public methods #
     ##################
 
+    def is_initial_state(self) -> bool:
+        """Determines if simulants could be initialized into this state.
+
+        Note: this will incorrectly return True if initialization_weights is a
```
stevebachmeier, line 245: This kinda seems like an important bug we shouldn't introduce. Is this case handled downstream at all?

stevebachmeier (commented):
I'm concerned about the case where an artifact or function returns all 0s

src/vivarium/framework/state_machine.py
rmudambi: The only implication of this happening is that we will throw a validation error if you specify an `initial_state` for your model if you specify a callable or artifact key for any state in that model even if that callable will always return 0, or if the key in the artifact only contains 0s. This doesn't seem like a big deal to me.

hussain-jafari (approved):

src/vivarium/framework/state_machine.py
stevebachmeier: I'd recommend at least pointing out in the docstring that it's not a big deal

src/vivarium/framework/state_machine.py
```diff
@@ -236,6 +239,20 @@ def __init__(
...
+        callable or an artifact key that will always return 0.
+        """
+        if self.initialization_weights is None:
```
stevebachmeier, line 248: I don't think you need these two lines

stevebachmeier (approved):

stevebachmeier (approved):

PR #544, opened 2024-11-18
mypy fixes in results/manager.py

hussain-jafari:
### Description
- *Category*: mypy
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5562

### Changes and notes
mypy fixes

### Testing
Ran mypy and pytest.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,7 +52,6 @@ exclude = [
     'src/vivarium/framework/population/manager.py',
     'src/vivarium/framework/population/population_view.py',
     'src/vivarium/framework/results/interface.py',
-    #'src/vivarium/framework/results/manager.py',
     'src/vivarium/framework/results/observer.py',
     'src/vivarium/framework/state_machine.py',
     'src/vivarium/interface/cli.py',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,7 +52,7 @@ exclude = [
     'src/vivarium/framework/population/manager.py',
     'src/vivarium/framework/population/population_view.py',
     'src/vivarium/framework/results/interface.py',
-    'src/vivarium/framework/results/manager.py',
+    #'src/vivarium/framework/results/manager.py',
     'src/vivarium/framework/results/observer.py',
     'src/vivarium/framework/state_machine.py',
     'src/vivarium/interface/cli.py',
--- a/src/vivarium/framework/results/manager.py
+++ b/src/vivarium/framework/results/manager.py
@@ -265,23 +265,24 @@ class ResultsManager(Manager):
         **cut_kwargs
             Keyword arguments for :meth: pandas.cut.
         """
+        if not isinstance(labels, list) or not all(
+            [isinstance(label, str) for label in labels]
+        ):
+            raise ValueError(
+                f"Labels must be a list of strings when registering a binned stratification, but labels was {labels} when registering the {binned_column} stratification."
+            )
 
         def _bin_data(data: pd.Series[int | float] | pd.DataFrame) -> pd.Series[Any]:
             """Use pandas.cut to bin continuous values"""
-            # mypy wants this squeeze to return a Series or DataFrame
-            if isinstance(data, pd.DataFrame):
-                data = data.squeeze()
+            data = data.squeeze() if isinstance(data, pd.DataFrame) else data
             if not isinstance(data, pd.Series):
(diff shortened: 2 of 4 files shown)


src/vivarium/framework/results/manager.py
```diff
@@ -271,14 +266,22 @@ def register_binned_stratification(
...
                 raise ValueError(f"Expected a Series, but got type {type(data)}.")
-            return pd.cut(
+            data = pd.cut(
                 data, bin_edges, labels=labels, right=False, include_lowest=True, **cut_kwargs
-            )
+            )  # type: ignore [call-overload]
+            if isinstance(data, pd.Series):
+                return data
+            else:
+                raise ValueError(
```
hussain-jafari, line 282: I believe the only way this could happen would be if labels were False, but it's typed and described in the doctstring as always being a list of strings.

pyproject.toml
```diff
@@ -52,7 +52,7 @@ exclude = [
     'src/vivarium/framework/population/manager.py',
     'src/vivarium/framework/population/population_view.py',
     'src/vivarium/framework/results/interface.py',
-    'src/vivarium/framework/results/manager.py',
+    #'src/vivarium/framework/results/manager.py',
```
patricktnast, line 55: delete entirely

src/vivarium/framework/results/manager.py
```diff
@@ -244,7 +239,7 @@ def register_binned_stratification(
         labels: list[str],
         excluded_categories: list[str] | None,
         target_type: str,
-        **cut_kwargs,
+        **cut_kwargs: dict[str, int | str | bool],
```
patricktnast, line 242: I think you should be able to just do int | str | bool here?
https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

hussain-jafari: Yeah that works as well

src/vivarium/framework/results/manager.py
```diff
@@ -271,14 +266,22 @@ def register_binned_stratification(
             Keyword arguments for :meth: pandas.cut.
         """
 
-        def _bin_data(data: pd.Series | pd.DataFrame) -> pd.Series:
+        def _bin_data(data: pd.Series[int | float] | pd.DataFrame) -> pd.Series[Any]:
             """Use pandas.cut to bin continuous values"""
-            data = data.squeeze()
+            # mypy wants this squeeze to return a Series or DataFrame
+            if isinstance(data, pd.DataFrame):
+                data = data.squeeze()
```
rmudambi, line 273: You should do 
```python
data = data.squeeze(axis=1)
```
This guarantees you won't get a scalar.

src/vivarium/framework/results/observation.py
```diff
@@ -163,7 +165,7 @@ def _wrap_results_gatherer(
 
     @staticmethod
     def create_empty_df(
-        requested_stratification_names: set[str],
+        requested_stratification_names: tuple[str, ...] | None,
```
rmudambi, line 168: Why is this changed from a set to a tuple?

src/vivarium/framework/results/observation.py
```diff
@@ -267,11 +269,15 @@ def create_expanded_df(
         """
 
         # Set up the complete index of all used stratifications
-        requested_and_registered_stratifications = [
-            stratification
-            for stratification in registered_stratifications
-            if stratification.name in requested_stratification_names
-        ]
+        requested_and_registered_stratifications = (
```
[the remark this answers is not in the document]
albrja: + 1

src/vivarium/framework/results/manager.py
albrja: If labels is False doesn't that mean the user implemented something incorrectly then?

src/vivarium/framework/results/observation.py
hussain-jafari: I think this was originally typed incorrectly, since we actually pass in a tuple[str, ...] or None to this function.

src/vivarium/framework/results/manager.py
hussain-jafari: mypy unfortunately still complains with this approach (with the original error and a new one about passing 1 to the axis argument) so I went with a one liner instead

rmudambi (approved):

stevebachmeier (approved):

PR #547, opened 2024-11-20
Lookup Manager Typing

patricktnast:
### Description

- *Category*: 
typing
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5548

### DIFF
--- a/src/vivarium/framework/lookup/manager.py
+++ b/src/vivarium/framework/lookup/manager.py
@@ -1,4 +1,3 @@
-# mypy: ignore-errors
 """
 =============
 Lookup Tables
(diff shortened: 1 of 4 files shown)


stevebachmeier (approved):

rmudambi (approved):

src/vivarium/framework/lookup/manager.py
```diff
@@ -52,7 +51,7 @@ def name(self) -> str:
         return "lookup_table_manager"
 
     def setup(self, builder: "Builder") -> None:
-        self.tables = {}
+        self.tables: dict[int, LookupTable] = {}
```
albrja, line 54: I would have guessed this is a str and not int with the way we set up lookup tables. Where do we index them with ints?

albrja (approved):

src/vivarium/framework/lookup/manager.py
patricktnast: ln 127

PR #549, opened 2024-12-06
fix mypy errors: component.py

stevebachmeier:
### Description

- *Category*:  other
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5599

### DIFF
--- a/src/vivarium/component.py
+++ b/src/vivarium/component.py
@@ -1,4 +1,3 @@
-# mypy: ignore-errors
 """
 =========
 Component
@@ -22,12 +21,13 @@ from typing import TYPE_CHECKING, Any
 
 import pandas as pd
 from layered_config_tree import ConfigurationError, LayeredConfigTree
-from loguru._logger import Logger
 
 from vivarium.framework.artifact import ArtifactException
 from vivarium.framework.population import PopulationError
 
 if TYPE_CHECKING:
+    from loguru import Logger
+
     from vivarium.framework.engine import Builder
     from vivarium.framework.event import Event
     from vivarium.framework.lookup import LookupTable
@@ -92,7 +92,24 @@ class Component(ABC):
     component. An empty dictionary indicates no managed configurations.
     """
 
-    def __repr__(self):
+    def __init__(self) -> None:
+        """Initializes a new instance of the Component class.
+
+        This method is the initializer for the Component class. It initializes
+        logger of type Logger and population_view of type PopulationView to None.
+        These attributes will be fully initialized in the setup_component method
+        of this class.
+        """
+        self._repr: str = ""
+        self._name: str = ""
+        self._sub_components: list["Component"] = []
+        self.logger: Logger | None = None
+        self.get_value_columns: Callable[[str | pd.DataFrame], list[str]] | None = None
+        self.configuration: LayeredConfigTree | None = None
+        self._population_view: PopulationView | None = None
+        self.lookup_tables: dict[str, LookupTable] = {}
+
+    def __repr__(self) -> str:
         """Returns a string representation of the __init__ call made to create this
         object.
 
@@ -111,16 +128,17 @@ class Component(ABC):
             object.
         """
         if not self._repr:
-            args = [
-                f"{name}={value.__repr__() if isinstance(value, Component) else value}"
-                for name, value in self.get_initialization_parameters().items()
-            ]
-            args = ", ".join(args)
+            args = ", ".join(
+                [
+                    f"{name}={value.__repr__() if isinstance(value, Component) else value}"
+                    for name, value in self.get_initialization_parameters().items()
+                ]
+            )
             self._repr = f"{type(self).__name__}({args})"
 
         return self._repr
 
-    def __str__(self):
+    def __str__(self) -> str:
         return self._repr
 
     ##############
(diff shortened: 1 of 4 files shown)


src/vivarium/component.py
```diff
@@ -92,7 +92,24 @@ class Component(ABC):
     component. An empty dictionary indicates no managed configurations.
     """
 
-    def __repr__(self):
+    def __init__(self) -> None:
```
stevebachmeier, line 95: This is a directly copy/paste that moves the constructor to the top. mypy really didn't like the init being at the bottom of the file.

src/vivarium/component.py
```diff
@@ -791,7 +798,7 @@ def _register_simulant_initializer(self, builder: Builder) -> None:
 
         if type(self).on_initialize_simulants != Component.on_initialize_simulants:
             builder.population.initializes_simulants(
-                self, creates_columns=self.columns_created, **initialization_requirements
+                self, creates_columns=self.columns_created, **initialization_requirements  # type: ignore[arg-type]
```
stevebachmeier, line 801: This block is temporary anyway since we're deprecating old functionalty here. It does highlight the fact, however, that we need to figure out how to best type hint kwargs that get unpacked at the signature. the current solution seems complicated. https://typing.readthedocs.io/en/latest/spec/callables.html#unpack-kwargs

src/vivarium/component.py
```diff
@@ -697,15 +704,15 @@ def get_data(self, builder: Builder, data_source: DataInput) -> Any:
...
+                data = data_source_callable(builder)
             else:
                 try:
                     data = builder.data.load(data_source)
                 except ArtifactException:
                     raise ConfigurationError(
                         f"Failed to find key '{data_source}' in artifact."
                     )
-        elif isinstance(data_source, Callable):
+        elif callable(data_source):
```
stevebachmeier, line 715: today I learned about the built-in `callable` which made a bunch of mypy errors go away. Not sure why mypy didn't like `isinstance(data_source, Callable)`. It was throwing `src/vivarium/component.py:718: error: Argument 2 to "isinstance" has incompatible type "<typing special form>"; expected "_ClassInfo"  [arg-type]` as well a bunch of operation errors at the next line (the call itself). This one change got rid of all of them.

src/vivarium/component.py
```diff
@@ -244,15 +262,15 @@ def initialization_requirements(
...
         """Provides a query to use when filtering the component's `PopulationView`.
 
         Returns
         -------
             A pandas query string for filtering the component's `PopulationView`.
-            Returns `None` if no filtering is required.
+            Returns an empty string if no filtering is required.
         """
-        return None
+        return ""
```
albrja, line 273: This seems like a big change based on validation for empty strings vs None. What is the reason for this change or was the type hinting just wrong?

src/vivarium/component.py
```diff
@@ -606,7 +607,9 @@ def build_lookup_table(
             raise ConfigurationError(f"Data '{data}' must be a LookupTableData instance.")
 
         if isinstance(data, list):
-            return builder.lookup.build_table(data, value_columns=list(value_columns))
+            return builder.lookup.build_table(
+                data, value_columns=list(value_columns) if value_columns else ()
```
albrja, line 611: Shouldn't this pass an empty list if value columns is None?

src/vivarium/component.py
stevebachmeier: Still working on this (the PR is in draft)

src/vivarium/component.py
stevebachmeier: doesn't really batter b/c build_table takes a Sequence (in fact it defaults to ())

src/vivarium/component.py
stevebachmeier: I think it's fine, though. It did indeed get flagged by mypy but all of the usage of the query are handling strings

stevebachmeier: I guess there's a chance that some downstream code is assuming this could be None. I can handle it if we think that's actually the case. but vph isn't using this property at all and everything in vivarium handles it as a str. (all of the defaults, e.g. are `""`, not `None`)

src/vivarium/component.py
```diff
@@ -627,10 +630,14 @@ def build_lookup_table(
         return builder.lookup.build_table(data)
 
     def _get_columns(
-        self, value_columns: Sequence[str] | None, data: float | pd.DataFrame
-    ) -> tuple[list[str], list[str], list[str]]:
+        self, value_columns: Sequence[str] | None, data: pd.DataFrame
+    ) -> tuple[Sequence[str], list[str], list[str]]:
         all_columns = list(data.columns)
         if value_columns is None:
+            if not self.get_value_columns:
```
rmudambi, line 637: `self.get_value_columns` will never actually be `None` at this point, since it's being set earlier in `setup_component()`. I think it's better to ignore the mypy error than introduce this check that will never trigger.

src/vivarium/framework/logging/manager.py
```diff
@@ -38,7 +37,7 @@ def _terminal_logging_not_configured() -> bool:
         # fragile since it depends on a loguru's internals as well as the stability of code
         # paths in vivarium, but both are quite stable at this point, so I think it's pretty,
         # low risk.
-        return 1 not in logger._core.handlers  # type: ignore[attr-defined]
+        return 1 not in loguru.logger._core.handlers  # type: ignore[attr-defined]
```
rmudambi, line 40: Why was this change needed?

rmudambi (approved):

src/vivarium/framework/logging/manager.py
stevebachmeier: Just a generic cleanup (we're already importing all of loguru). And to make it consistent w/ te component.py

src/vivarium/component.py
stevebachmeier: I can cast?

albrja (approved):

patricktnast (approved):

PR #550, opened 2024-12-09
results/interface.py typing

hussain-jafari:
results/interface.py typing

### DIFF
--- a/src/vivarium/framework/results/interface.py
+++ b/src/vivarium/framework/results/interface.py
@@ -35,7 +35,7 @@ ResultsUpdater = Callable[[pd.DataFrame, pd.DataFrame], pd.DataFrame]
 ResultsFormatter = Callable[[str, pd.DataFrame], pd.DataFrame]
 """This is a Callable that takes a measure as a string and a DataFrame of observation results and returns formatted results."""
 ResultsGathererInput = Union[
-    pd.DataFrame, DataFrameGroupBy[tuple[str, ...] | str, bool], tuple[str, ...], None
+    pd.DataFrame, DataFrameGroupBy, tuple[str, ...], None  # type: ignore [type-arg]
 ]
 ResultsGatherer = Callable[[ResultsGathererInput], pd.DataFrame]
 """This is a Callable that optionally takes a possibly stratified population and returns new observation results."""
--- a/src/vivarium/framework/results/interface.py
+++ b/src/vivarium/framework/results/interface.py
@@ -14,7 +14,6 @@ from collections.abc import Callable
 from typing import TYPE_CHECKING, Any, Union
 
 import pandas as pd
-
 from pandas.core.groupby.generic import DataFrameGroupBy
 
 from vivarium.framework.event import Event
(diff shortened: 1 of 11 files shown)


src/vivarium/framework/results/interface.py
```diff
@@ -23,13 +23,13 @@
...
 from vivarium.manager import Interface
-from vivarium.types import ScalarValue
+from vivarium.types import ScalarMapper, VectorMapper
 
 if TYPE_CHECKING:
     from vivarium.framework.results.manager import ResultsManager
 
 
-def _required_function_placeholder(*args, **kwargs) -> pd.DataFrame:
+def _required_function_placeholder(*args: Any, **kwargs: Any) -> pd.DataFrame:
```
stevebachmeier, line 32: This is interesting! In reality, we do know the argument types here (b/c we only use this for required_updated, _formatter, and _gatherer), but since this is a sort of sentinel value it just doesn't matter.

src/vivarium/framework/results/interface.py
```diff
@@ -182,7 +180,7 @@ def register_stratified_observation(
         additional_stratifications: list[str] = [],
         excluded_stratifications: list[str] = [],
         aggregator_sources: list[str] | None = None,
-        aggregator: Callable[[pd.DataFrame], float | pd.Series] = len,
+        aggregator: Callable[[pd.DataFrame], float | pd.Series[Any]] = len,
```
stevebachmeier, line 183: Arent the series' always number likes?

src/vivarium/framework/results/interface.py
```diff
@@ -323,7 +321,7 @@ def register_adding_observation(
...
```
stevebachmeier, line 324: Aren't the series' always number-likes?

src/vivarium/framework/results/interface.py
```diff
@@ -438,7 +436,7 @@ def register_concatenating_observation(
 
     @staticmethod
     def _check_for_required_callables(
-        observation_name: str, required_callables: dict[str, Callable]
+        observation_name: str, required_callables: dict[str, Callable[..., pd.DataFrame]]
```
stevebachmeier, line 439: This makes me think we _should_ properly type hint the placeholder callable; we can definitely do better than `...`, no?

src/vivarium/framework/results/interface.py
rmudambi: I think they are in practice, but in principle they don't need to be? Here's an example:

We're recording blood pressure for a population and we're stratifying by age group. We don't actually care about the actual mean blood pressure - we want to categorize population-level mean blood pressure into low, medium, and high. Then the aggregator function would return a string, not a number.

src/vivarium/framework/results/interface.py
rmudambi: In this case they should always be number-like, since the results_updater is performing addition.

src/vivarium/framework/results/interface.py
rmudambi: I agree, we should type-hint the placeholder, since we do know something about the arguments the provided callable should have.

src/vivarium/framework/results/interface.py
hussain-jafari: In that case I'll leave it as Any. Does that mean we want to update the scalar type as well?

src/vivarium/framework/results/interface.py
```diff
@@ -289,11 +297,11 @@ def register_unstratified_observation(
         ValueError
             If any required callable arguments are missing.
         """
-        required_callables = {
+        required_callables: dict[str, Callable[..., pd.DataFrame]] = {
```
rmudambi, line 300: Shouldn't this be `dict[str, ResultsUpdater | ResultsFormatter | ResultsGatherer]`?

rmudambi (approved):

src/vivarium/framework/results/interface.py
hussain-jafari: I tried that, as well as the following (the arguments can either be one DataFrame or two DataFrames)

`Callable[[pd.DataFrame, ...], pd.DataFrame]`
`Callable[[pd.DataFrame], pd.DataFrame] | Callable[[pd.DataFrame, pd.DataFrame], pd.DataFrame]`
`Callable[[pd.DataFrame | [pd.DataFrame, pd.DataFrame]], pd.DataFrame]`

but they all gave me errors like the following (this one is from Callable[[pd.DataFrame,...], pd.DataFrame]

Argument 2 to "_check_for_required_callables" of "ResultsInterface" has incompatible type "dict[str, Callable[[DataFrame, Any], DataFrame]]"; expected "dict[str, Callable[[str, DataFrame], DataFrame] | Callable[[DataFrame], DataFrame] | Callable[[DataFrame, DataFrame], DataFrame]]"  [arg-type]

stevebachmeier (approved):

PR #554, opened 2024-12-11
Mypy State Machine

patricktnast:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5602

### Changes and notes
 
A few notes:
mypy doesn't recognize numbers.Number as a valid type for the most part. see discussion here:
https://github.com/python/mypy/issues/3186
in most cases, I changed it to float | int or just float (in typehinting, 'int' counts as float but not in isinstance() https://peps.python.org/pep-0484/#the-numeric-tower

We needed to use sequence in Component so that we could have subclasses of Components be valid.

### Testing

tests pass

### DIFF
--- a/src/vivarium/component.py
+++ b/src/vivarium/component.py
@@ -102,7 +102,7 @@ class Component(ABC):
         """
         self._repr: str = ""
         self._name: str = ""
-        self._sub_components: list["Component"] = []
+        self._sub_components: Sequence["Component"] = []
         self.logger: loguru.Logger | None = None
         self.get_value_columns: Callable[[str | pd.DataFrame], list[str]] | None = None
         self.configuration: LayeredConfigTree | None = None
@@ -205,7 +205,7 @@ class Component(ABC):
         return self._population_view
 
     @property
-    def sub_components(self) -> list["Component"]:
+    def sub_components(self) -> Sequence["Component"]:
         """Provide components managed by this component.
 
         Returns
--- a/src/vivarium/component.py
+++ b/src/vivarium/component.py
@@ -16,7 +16,6 @@ from collections.abc import Callable, Sequence
 from datetime import datetime, timedelta
 from importlib import import_module
 from inspect import signature
-from numbers import Number
 from typing import TYPE_CHECKING, Any, cast
 
 import pandas as pd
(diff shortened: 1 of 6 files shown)


src/vivarium/framework/state_machine.py
```diff
@@ -72,8 +72,8 @@ def _next_state(
 
 
 def _groupby_new_state(
-    index: pd.Index, outputs: list, decisions: pd.Series
-) -> list[tuple[str, pd.Index]]:
+    index: pd.Index[int], outputs: list[State | str], decisions: pd.Series[Any]
```
patricktnast, line 75: mypy got mad at me for trying pd.Series[State | str]

src/vivarium/framework/state_machine.py
```diff
@@ -443,7 +448,9 @@ def extend(self, transitions: Iterable[Transition]) -> None:
     # Helper methods #
     ##################
 
-    def _normalize_probabilities(self, outputs, probabilities):
+    def _normalize_probabilities(
+        self, outputs: list[State | str], probabilities: NumericArray
```
patricktnast, line 452: you could avoid some of this `| str` business by making "null_transition" a `State` instead of just a string, but i wasn't sure whether it was worth the hassle.

src/vivarium/types.py
```diff
@@ -17,7 +16,7 @@
 ClockTime = Time | int
 ClockStepSize = Timedelta | int
 
-ScalarValue = Number | Timedelta | Time
+ScalarValue = float | Timedelta | Time
```
albrja, line 19: Shouldn't this be float or int?

albrja (approved):

src/vivarium/types.py
stevebachmeier: Good catch

src/vivarium/component.py
```diff
@@ -16,7 +16,6 @@
 from datetime import datetime, timedelta
 from importlib import import_module
 from inspect import signature
-from numbers import Number
```
stevebachmeier, line 19: I probably would have created a Number type that is `int | float` to minimize the changes needed. Since you've already done this it's probably fine, but _might_ be worth looking through the rest of the code base to see if we are importing Number from numbers elsewhere

stevebachmeier (approved):

src/vivarium/component.py
patricktnast: I already looked; this PR removes every instance of `from numbers import Number`

src/vivarium/types.py
patricktnast: in typehinting, 'int' counts as float but not in isinstance() https://peps.python.org/pep-0484/#the-numeric-tower

stevebachmeier: but I think it's important to relay to us that it can also be an int

PR #560, opened 2024-12-17
Type-hinting: component manager

rmudambi:
### Description

- *Category*: type-hinting
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5601

### Changes and notes
- Type-hinted vivarium/framework/components/ manager

### Testing
Builds pass

### DIFF
--- a/src/vivarium/framework/components/manager.py
+++ b/src/vivarium/framework/components/manager.py
@@ -1,4 +1,3 @@
-# mypy: ignore-errors
 """
 ============================
 The Component Manager System
@@ -16,6 +15,7 @@ and managers and components are given to it by the context. It is called on to
 setup everything it holds when the context itself is setup.
 
 """
+from __future__ import annotations
 
 import inspect
 from collections.abc import Iterator, Sequence
@@ -55,7 +55,7 @@ class OrderedComponentSet:
         if args:
             self.update(args)
 
-    def add(self, component: Component) -> None:
+    def add(self, component: Component | Manager) -> None:
         if component in self:
             raise ComponentConfigError(
                 f"Attempting to add a component with duplicate name: {component}"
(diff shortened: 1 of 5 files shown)


src/vivarium/framework/components/manager.py
```diff
@@ -193,9 +200,12 @@ def get_components_by_type(
             A list of components of type ``component_type``.
         """
         # Convert component_type to a tuple for isinstance
-        return [c for c in self._components if isinstance(c, tuple(component_type))]
+        component_type = (
```
stevebachmeier, line 203: Was this required or is it adding functionality? This is less readable to me than the original.

src/vivarium/framework/components/manager.py
```diff
@@ -254,7 +264,10 @@ def apply_configuration_defaults(self, component: Component | Manager) -> None:
             )
         except DuplicatedConfigurationError as e:
             new_name, new_file = component.name, self._get_file(component)
-            old_name, old_file = e.source, self._get_file(self.get_component(e.source))
+            if e.source:
+                old_name, old_file = e.source, self._get_file(self.get_component(e.source))
+            else:
+                old_name, old_file = "another component", "another file"
```
stevebachmeier, line 270: clever. It does lead to a slightly awkward error message "...but it has already been set by another component in file another file".

Not sure if the word "file" should be removed in the message or not below.

stevebachmeier (approved):

hussain-jafari (approved):

src/vivarium/framework/components/manager.py
rmudambi: The previous functionality was incorrect, as you can only call `tuple()` on an iterable.

PR #561, opened 2024-12-19
component parser typing

hussain-jafari:
### Description
- *Category*: typing
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5648

### Changes and notes
typing

### Testing
Ran mypy and pytest.

### DIFF
--- a/src/vivarium/framework/components/parser.py
+++ b/src/vivarium/framework/components/parser.py
@@ -1,4 +1,3 @@
-# mypy: ignore-errors
 """
 ==================================
 The Component Configuration Parser
@@ -22,12 +21,16 @@ There are three steps to this process.
 
 """
 
+from typing import Union
+
 from layered_config_tree.main import LayeredConfigTree
 
 from vivarium.framework.utilities import import_by_path
 
 from ... import Component
 from .manager import ComponentConfigError
+
+_NestedDict = dict[str, Union[str, list[str], "_NestedDict"]]
 
 
 class ParsingError(ComponentConfigError):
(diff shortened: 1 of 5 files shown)


albrja (approved):

src/vivarium/framework/components/parser.py
```diff
@@ -224,7 +227,7 @@ def prep_component(self, component_string: str) -> tuple[str, tuple]:
         return path, cleaned_args
 
     @staticmethod
-    def _clean_args(args: list, path: str) -> tuple:
+    def _clean_args(args: list[str], path: str) -> list[str]:
         """Transform component arguments into a tuple, validating that each argument
```
patricktnast, line 231: nit: if you're making these lists instead of tuples you should update the docstring.

tests/framework/components/test_parser.py
```diff
@@ -126,7 +126,7 @@ def test_parse_and_prep_components(parser):
     prepped_components = [
         parser.prep_component(component) for component in TEST_COMPONENTS_PARSED
     ]
-
+    prepped_components = [(x, tuple(y)) for (x, y) in prepped_components]
```
patricktnast, line 129: why not just change TEST_COMPONENTS_PREPPED?

src/vivarium/framework/components/parser.py
```diff
@@ -252,10 +255,10 @@ def _clean_args(args: list, path: str) -> tuple:
                 raise ParsingError(f"Invalid component argument {a} for component {path}")
 
             out.append(a[1:-1])
-        return tuple(out)
+        return out
```
patricktnast, line 258: could you explain why you changed this from a tuple to a list in the first place?

patricktnast (approved):

tests/framework/components/test_parser.py
hussain-jafari: We can't convert it into a set if we use lists instead of tuples

src/vivarium/framework/components/parser.py
hussain-jafari: I tried to reproduce the mypy error that led to this change and things ended up working without that change with a bit of typing actually so I've undone the list conversion and all related changes.

PR #571, opened 2025-01-10
Mypy errors test_crn.py

patricktnast:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5719

### DIFF
--- a/tests/framework/randomness/test_crn.py
+++ b/tests/framework/randomness/test_crn.py
@@ -5,7 +5,7 @@ Integration tests primarily meant to test the CRN guarantees for the Randomness 
 
 from collections.abc import Iterator
 from itertools import cycle
-from typing import Type
+from typing import Type, TypedDict, Callable
 
 import numpy as np
 import pandas as pd
@@ -22,11 +22,23 @@ from vivarium.interface import InteractiveContext
 
 
 @pytest.mark.parametrize("initializes_crn_attributes", [True, False])
-def test_basic_repeatability(initializes_crn_attributes):
+def test_basic_repeatability(initializes_crn_attributes: bool) -> None:
     test_idx = pd.Index(range(100))
     index_map = IndexMap()
 
-    stream_args = {
+    StreamArgs = TypedDict(
+        "StreamArgs",
+        {
+            "key": str,
+            "clock": Callable[[], pd.Timestamp],
+            "seed": str,
+            "index_map": IndexMap,
+            "initializes_crn_attributes": bool,
+        },
+        total=False,
+    )
+
+    stream_args: StreamArgs = {
         "key": "test",
         "clock": lambda: pd.Timestamp("2020-01-01"),
         "seed": "abc",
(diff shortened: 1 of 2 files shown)


albrja (approved):

tests/framework/randomness/test_crn.py
```diff
@@ -36,14 +48,14 @@ def test_basic_repeatability(initializes_crn_attributes):
...
+    arg_permutations: list[StreamArgs] = [
         {},
         {"key": "test2"},
         {"clock": lambda: pd.Timestamp("2020-01-02")},
         {"seed": "123"},
-    ]:
-        new_stream_args = {**stream_args, **arg_permutation}
+    ]
+    for arg_permutation in arg_permutations:
+        new_stream_args: StreamArgs = stream_args | arg_permutation
```
rmudambi, line 58: What does the `|` operator do here?

tests/framework/randomness/test_crn.py
```diff
@@ -22,11 +22,23 @@
 
 
 @pytest.mark.parametrize("initializes_crn_attributes", [True, False])
-def test_basic_repeatability(initializes_crn_attributes):
+def test_basic_repeatability(initializes_crn_attributes: bool) -> None:
     test_idx = pd.Index(range(100))
     index_map = IndexMap()
 
-    stream_args = {
+    StreamArgs = TypedDict(
```
rmudambi, line 29: Is this too much overhead and boilerplate to be making this worthwhile?

rmudambi (commented):
I'm just leaving some comments, but feel free to decide among the 4 of you what you want to do with it while I'm out.

tests/framework/randomness/test_crn.py
stevebachmeier: Same question. Is this really the same as combining like {**stream_args, **arg_permutation}?

tests/framework/randomness/test_crn.py
stevebachmeier: imo opinion it's not worth the overhead, but since it's already done 👍

stevebachmeier: That said, I've never seen the first arg ("StreamArgs") or the `total` arg. What are those?

stevebachmeier (approved):

tests/framework/randomness/test_crn.py
patricktnast: https://peps.python.org/pep-0584/

tests/framework/randomness/test_crn.py
patricktnast: It seemed to me that the natural scheme to just define all the args directly would be more verbose, and I think ideally we shouldn't ignore something unless it's actually impossible. Don't feel strongly though, esp. if somebody can come up with a more concise solution.

patricktnast: https://peps.python.org/pep-0589/#totality RE `total`

PR #574, opened 2025-01-23
Albrja/mic 5720/mypy tests randomness index map

albrja:
### Fix mypy errors in test_index_map.py
- *Category*: Type checking
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5720

### Changes and notes
-fix mypy errors

### DIFF
--- a/src/vivarium/framework/randomness/index_map.py
+++ b/src/vivarium/framework/randomness/index_map.py
@@ -192,7 +192,7 @@ class IndexMap:
         return new_map % len(self)
 
     def _convert_to_ten_digit_int(
-        self, column: pd.Series[datetime | int | float]
+        self, column: pd.Series[datetime] | pd.Series[int] | pd.Series[float]
     ) -> pd.Series[int]:
         """Converts a column of datetimes, integers, or floats into a column
         of 10 digit integers.
--- a/src/vivarium/framework/randomness/index_map.py
+++ b/src/vivarium/framework/randomness/index_map.py
@@ -192,7 +192,11 @@ class IndexMap:
         return new_map % len(self)
 
     def _convert_to_ten_digit_int(
-        self, column: pd.Series[datetime] | pd.Series[int] | pd.Series[float]
+        self,
+        column: pd.Series[pd.Timestamp]
+        | pd.Series[datetime]
+        | pd.Series[int]
+        | pd.Series[float],
     ) -> pd.Series[int]:
         """Converts a column of datetimes, integers, or floats into a column
         of 10 digit integers.
@@ -215,6 +219,7 @@ class IndexMap:
         if pdt.is_datetime64_any_dtype(column):
             integers = self._clip_to_seconds(column.astype(np.int64))
         elif pdt.is_integer_dtype(column):
+            column = column.astype(int)
             if not len(column >= 0) == len(column):
                 raise RandomnessError(
                     "Values in integer columns must be greater than or equal to zero."
--- a/tests/framework/randomness/test_index_map.py
+++ b/tests/framework/randomness/test_index_map.py
@@ -115,18 +115,14 @@ def test_shift_series() -> None:
 def test_convert_to_ten_digit_int() -> None:
     m = IndexMap()
     v = 1234567890
-    real_col = pd.date_range(
+    date_col = pd.date_range(
         pd.to_datetime(v, unit="s"), periods=10000, freq="ns"
     ).to_series()
-    datetime_val: pd.Timestamp = pd.to_datetime(v, unit="s")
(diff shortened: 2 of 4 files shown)


src/vivarium/framework/randomness/index_map.py
```diff
@@ -215,6 +219,7 @@ def _convert_to_ten_digit_int(
         if pdt.is_datetime64_any_dtype(column):
             integers = self._clip_to_seconds(column.astype(np.int64))
         elif pdt.is_integer_dtype(column):
+            column = column.astype(int)
```
albrja, line 222: This behavior is consistent with the other branch of the if/else block and makes mypy happy.

src/vivarium/framework/randomness/index_map.py
```diff
@@ -192,7 +192,11 @@ def _hash(self, keys: pd.Index[Any], salt: ClockTime = 0) -> pd.Series[int]:
         return new_map % len(self)
 
     def _convert_to_ten_digit_int(
-        self, column: pd.Series[datetime | int | float]
+        self,
+        column: pd.Series[pd.Timestamp]
```
albrja, line 196: This was typed incorrectly before but basically to make myppy happy, we have to have pd.Timestamp as the type to include datetime because pandas always converts datetimes to Timestamps. However, Timestamp is technically a part of datetime. See: https://pandas.pydata.org/docs/user_guide/timeseries.html

stevebachmeier: So even though pd.Timestamp inherits from datatime, you have to include it explicitly (despite _also_ having datetime as part of the signature)?

Also, glad you caught that incorrect mashing of series types!

tests/framework/randomness/test_index_map.py
```diff
@@ -38,7 +49,7 @@ def generate_keys(number, types=("int", "float", "datetime"), seed=123456):
             keys["int"] = np.arange(start, start + number, dtype=int)
 
     if "float" in types:
-        keys["float"] = rs.random_sample(size=number)
+        keys["float"] = rs.random_sample(number)
```
stevebachmeier, line 52: what drove removing the keyword?

tests/framework/randomness/test_index_map.py
```diff
@@ -149,58 +160,65 @@ def test_hash_uniformity(map_size_and_hashed_values):
...
     mock_index_map = IndexMap
 
-    def hash_mock(k, salt):
+    def hash_mock(
+        k: pd.Index[int], salt: pd.Series[datetime] | pd.Series[int] | pd.Series[float]
+    ) -> pd.Series[int]:
         seed = 123456
-        salt = IndexMap()._convert_to_ten_digit_int(pd.Series(salt, index=k))
-        rs = np.random.RandomState(seed=seed + salt)
+        digit_series = IndexMap()._convert_to_ten_digit_int(pd.Series(salt, index=k))
```
stevebachmeier, line 170: nit: maybe `salt_series`?

tests/framework/randomness/test_index_map.py
```diff
@@ -149,58 +160,65 @@ def test_hash_uniformity(map_size_and_hashed_values):
...
 
-def test_update_empty_good_keys(index_map):
+def test_update_empty_good_keys(index_map: type[IndexMap]) -> None:
     keys = generate_keys(1000)
     m = index_map(key_columns=list(keys.columns))
     m.update(keys, pd.to_datetime("2023-01-01"))
     key_index = keys.set_index(list(keys.columns)).index
-    assert len(m._map) == len(keys), "All keys not in mapping"
+    map = m._map
+    assert isinstance(map, pd.Series)
```
stevebachmeier, line 200: I wonder if you need this assertion if you typed map specifically, `map: pd.Series = m._map`

stevebachmeier (approved):

src/vivarium/framework/randomness/index_map.py
albrja: Yes - I think technically we could just make it pd.Timestamp and not have a series as datetime as an options since no matter what you do pandas will convert it but I t hink this is more intuitive since we do use datetime elsewhere and we have Time in types.py which is a Timestamp | datetime (and unfortunately I could not do a serious of that).

tests/framework/randomness/test_index_map.py
albrja: Not sure why I had deleted this but it is back in there.

tests/framework/randomness/test_index_map.py
albrja: I specifically had to do this because _map can either be pd.Series[int] or None but in this case it has to be a series. For some reason mypy doesnt like inline typing vs the assertion.

patricktnast (approved):

PR #582, opened 2025-02-04
Albrja/mic 5401/lookup dict

albrja:
### Add dictionary as data type that can create a lookup table
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5401

### Changes and notes
-add dict as type to LookupTableData


src/vivarium/component.py
```diff
@@ -25,6 +25,7 @@
 
 from vivarium.framework.artifact import ArtifactException
 from vivarium.framework.population import PopulationError
+from vivarium.types import ScalarValue
```
albrja, line 28: I don't understand this but when I put this in the type checking block it would error saying ScalarValue wasn't defined.

src/vivarium/framework/lookup/manager.py
```diff
@@ -204,7 +207,7 @@ def validate_build_table_parameters(
     ):
         raise ValueError("Must supply some data")
 
-    acceptable_types = (Numeric, datetime, timedelta, list, tuple, pd.DataFrame)
+    acceptable_types = (Numeric, datetime, timedelta, list, tuple, pd.DataFrame, dict)
```
rmudambi, line 210: Data will have been converted from a dict to a Dataframe in line 89, so we don't want to add it here.

rmudambi (approved):
I'm finding the required change to the `get_value_columns()` typing to be frustrating, but I think the only alternative is to type-hint it as `str | Any` or even just `Any`, which I guess is probably not desirable.

hussain-jafari (approved):

PR #592, opened 2025-02-26
change Event to a frozen dataclass

stevebachmeier:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5468

### Changes and notes
 
The only reason to freeze is b/c a single test asserted that an Event
is immutable. If we don't care about this, I can instead unfreeze and
remove that test. However, it makes philosophical sense to me that
you would never want to change an Event.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
 # default owners
-* @albrja @collijk @hussain-jafari @patricktnast @rmudambi @stevebachmeier
+* @albrja @hussain-jafari @patricktnast @rmudambi @stevebachmeier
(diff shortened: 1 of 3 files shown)


.github/CODEOWNERS
```diff
@@ -1,2 +1,2 @@
 # default owners
-* @albrja @collijk @hussain-jafari @patricktnast @rmudambi @stevebachmeier
```
stevebachmeier, line 2: Is it time?

albrja: Cut from the team 😭

albrja (approved):

patricktnast (approved):

stevebachmeier:
> On the philosophical question, I don't see why not.

Sorry @patricktnast my original comment that you replied to had a typo and was ambiguous. Are you saying that you think Events should be mutable or immutable?

PR #610, opened 2025-04-15
Albrja/mic 5373/all columns required and create column population view

albrja:
### Update Component to be able to create a column if all state table columns are required for the population view.
- *Category*: Bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5373

### Changes and notes
-allow Component to create a column if all state table columns are required.


albrja:
A very easy way to think about this change is if a component requires all state table columns for a population view and will also create a column, there is a new optional argument that can be changed to True to handle this use case otherwise it can be ignored. This also means it is not a breaking change.

rmudambi (changes requested):
This looks like it should work, but I think it'd be cleaner if the new argument passed to `get_view()` was `requires_all_columns` rather than `creates_and_requires_all`.

stevebachmeier (approved):

src/vivarium/component.py
```diff
@@ -746,12 +746,26 @@ def _set_population_view(self, builder: "Builder") -> None:
...
+            )
         if self.columns_required:
             # Get all columns created and required
             population_view_columns = self.columns_created + self.columns_required
-        elif self.columns_required == []:
+        elif self.columns_required == "all" or self.columns_required == []:
             # Empty list means population view needs all available columns
-            population_view_columns = []
+            # We cannot overwrite self.columns_required = "all" above because it does
+            # not have a setter.
```
rmudambi, line 763: We don't need this comment in code imo

src/vivarium/framework/population/manager.py
```diff
@@ -439,13 +464,17 @@ def get_view(self, columns: str | Sequence[str], query: str = "") -> PopulationV
             The query should be provided in a way that is understood by the
             :meth:`pandas.DataFrame.query` method and may reference state
             table columns not requested in the ``columns`` argument.
+        requires_all_columns
+            If True, all columns in the population state table will be
+            required for the population view and the columns will be
+            the columns created by the component.
```
rmudambi, line 470:
```suggestion
            If True, all columns in the population state table will be included in
            the population view.
```

src/vivarium/framework/population/manager.py
```diff
@@ -224,14 +230,26 @@ def get_view(self, columns: str | Sequence[str], query: str = "") -> PopulationV
             The query should be provided in a way that is understood by the
             :meth:`pandas.DataFrame.query` method and may reference state
             table columns not requested in the ``columns`` argument.
+        requires_all_columns
+            If True, all columns in the population state table are required for
+            the population view and the columns will be the columns created by the
+            component.
```
rmudambi, line 236:
```suggestion
            If True, all columns in the population state table will be included in
            the population view.
```

src/vivarium/framework/population/population_view.py
```diff
@@ -60,11 +61,15 @@ def __init__(
         query
             A :mod:`pandas`-style filter that will be applied any time this
             view is read from.
+        requires_all_columns
+            If True, all columns from the population state table are required and
+            the columns will be the columns created by the component.
```
rmudambi, line 66:
```suggestion
            If True, all columns in the population state table will be included in
            the population view.
```

src/vivarium/framework/population/population_view.py
```diff
@@ -60,11 +61,15 @@ def __init__(
         query
```
rmudambi, line 61: Need to update the docstring for `column` to read "The set of columns this view should have access to. If requires_all_columns is True, this is instead the columns created by the component containing this population view."

This also applies to all the relevant locations in the manager and interface.

src/vivarium/framework/population/population_view.py
```diff
@@ -79,9 +84,9 @@ def columns(self) -> list[str]:
         should be only be used in situations where the full state table is
         actually needed, like for some metrics collection applications.
         """
-        if not self._columns:
-            return list(self._manager.get_population(True).columns)
-        return list(self._columns)
+        if self.requires_all_columns:
+            return list(self._manager.get_population(True).columns) + self._columns
+        return self._columns
```
rmudambi, line 89: If you access this property after the initialization phase of the simulation, are the created columns duplicated? I suspect they are. If so, you should use sets to weed out the duplicates.

src/vivarium/framework/population/population_view.py
```diff
@@ -79,9 +84,9 @@ def columns(self) -> list[str]:
         should be only be used in situations where the full state table is
```
rmudambi, line 84: This docstring also needs to be updated.

rmudambi (changes requested):

src/vivarium/framework/population/manager.py
```diff
@@ -222,18 +222,17 @@ def get_view(
         ----------
         columns
             A subset of the state table columns that will be available in the
-            returned view. If empty, this view will have access to the entire
-            state table.
+            returned view. If requires_all_columns is True, this will be the
```
rmudambi, line 225: Nit: can you make this wording change everywhere we say this? I prefer that, because setting the this argument doesn't actually change what columns the component will create - we're just telling the user what they should put here. 
```suggestion
            returned view. If requires_all_columns is True, this should be set to the
```

PR #620, opened 2025-05-07
Albrja/mic-5664/update-rate-to-probability

albrja:
### Update utility functions for rate to probability and probability to rate conversions
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5664

### Changes and notes
-updates for rate and probability conversions
-Discussion and requested change can be found here: https://github.com/ihmeuw/vivarium_research/pull/1391

### DIFF
--- a/src/vivarium/framework/randomness/manager.py
+++ b/src/vivarium/framework/randomness/manager.py
@@ -101,6 +101,7 @@ class RandomnessManager(Manager):
         decision_point: str,
         component: Component | None,
         initializes_crn_attributes: bool = False,
+        rate_conversion_type: str | None = None,
     ) -> RandomnessStream:
         """Provides a new source of random numbers for the given decision point.
 
@@ -118,6 +119,10 @@ class RandomnessManager(Manager):
             in the Common Random Number framework. These streams cannot be
             copied and should only be used to generate the state table columns
             specified in ``builder.configuration.randomness.key_columns``.
+        rate_conversion_type
+            The type of conversion to use. Default is "linear" for a simple
+            multiplication of rate and time_scaling_factor. Other option is
+            "exponential".
 
         Returns
         -------
@@ -131,8 +136,10 @@ class RandomnessManager(Manager):
             If another location in the simulation has already created a randomness stream
             with the same identifier.
         """
+        if rate_conversion_type is None:
+            rate_conversion_type = str(self._rate_conversion_type)
         stream = self._get_randomness_stream(
-            decision_point, component, initializes_crn_attributes
+            decision_point, component, initializes_crn_attributes, rate_conversion_type
         )
         if not initializes_crn_attributes:
             # We need the key columns to be created before this stream can be called.
@@ -158,6 +165,7 @@ class RandomnessManager(Manager):
         decision_point: str,
         component: Component | None,
         initializes_crn_attributes: bool = False,
+        rate_conversion_type: str = "linear",
     ) -> RandomnessStream:
         if decision_point in self._decision_points:
             raise RandomnessError(
@@ -176,7 +184,7 @@ class RandomnessManager(Manager):
             index_map=self._key_mapping,
             component=component,
             initializes_crn_attributes=initializes_crn_attributes,
-            rate_conversion_type=self._rate_conversion_type,
+            rate_conversion_type=rate_conversion_type,
         )
         self._decision_points[decision_point] = stream
         return stream
--- a/src/vivarium/framework/randomness/manager.py
+++ b/src/vivarium/framework/randomness/manager.py
@@ -164,6 +164,11 @@ class RandomnessManager(Manager):
                 f"Two separate places are attempting to create "
                 f"the same randomness stream for {decision_point}"
             )
+        if self._rate_conversion_type not in ["linear", "exponential"]:
+            raise RandomnessError(
+                "The rate conversion type must be a linear or exponential. "
+                f"Got {self._rate_conversion_type} instead."
+            )
         stream = RandomnessStream(
             key=decision_point,
             clock=self._clock,
--- a/src/vivarium/framework/randomness/manager.py
+++ b/src/vivarium/framework/randomness/manager.py
@@ -172,11 +172,6 @@ class RandomnessManager(Manager):
                 f"Two separate places are attempting to create "
                 f"the same randomness stream for {decision_point}"
             )
-        if self._rate_conversion_type not in ["linear", "exponential"]:
-            raise RandomnessError(
-                "The rate conversion type must be a linear or exponential. "
-                f"Got {self._rate_conversion_type} instead."
-            )
         stream = RandomnessStream(
             key=decision_point,
             clock=self._clock,
--- a/src/vivarium/framework/randomness/manager.py
+++ b/src/vivarium/framework/randomness/manager.py
@@ -7,7 +7,7 @@ Randomness System Manager
 from __future__ import annotations
 
 from collections.abc import Callable
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Literal
 
 import pandas as pd
 
@@ -33,7 +33,7 @@ class RandomnessManager(Manager):
             "key_columns": [],
             "random_seed": 0,
             "additional_seed": None,
-            "rate_conversion_type": None,
+            "rate_conversion_type": "linear",
         }
     }
 
@@ -43,7 +43,7 @@ class RandomnessManager(Manager):
         self._key_columns: list[str] = []
         self._key_mapping_: IndexMap | None = None
         self._decision_points: dict[str, RandomnessStream] = dict()
-        self._rate_conversion_type: str | None = None
+        self._rate_conversion_type: Literal["linear", "exponential"] = "linear"
 
     @property
     def name(self) -> str:
@@ -76,11 +76,7 @@ class RandomnessManager(Manager):
         pop_size = builder.configuration.population.population_size
         map_size = max(map_size, 10 * pop_size)
         self._key_mapping_ = IndexMap(self._key_columns, map_size)
-        if builder.configuration.randomness.rate_conversion_type is not None:
-            self._rate_conversion_type = builder.configuration.randomness.rate_conversion_type
-        else:
-            self._rate_conversion_type = "linear"
-
+        self._rate_conversion_type = builder.configuration.randomness.rate_conversion_type
         self.resources = builder.resources
         self._add_constraint = builder.lifecycle.add_constraint
         self._add_constraint(self.get_seed, restrict_during=["initialization"])
@@ -101,7 +97,7 @@ class RandomnessManager(Manager):
         decision_point: str,
         component: Component | None,
         initializes_crn_attributes: bool = False,
-        rate_conversion_type: str | None = None,
+        rate_conversion_type: Literal["linear", "exponential"] = "linear",
     ) -> RandomnessStream:
         """Provides a new source of random numbers for the given decision point.
 
@@ -121,7 +117,7 @@ class RandomnessManager(Manager):
             specified in ``builder.configuration.randomness.key_columns``.
         rate_conversion_type
             The type of conversion to use. Default is "linear" for a simple
-            multiplication of rate and time_scaling_factor. Other option is
+            multiplication of rate and time_scaling_factor. The other option is
             "exponential".
 
         Returns
@@ -136,8 +132,6 @@ class RandomnessManager(Manager):
             If another location in the simulation has already created a randomness stream
             with the same identifier.
         """
-        if rate_conversion_type is None:
-            rate_conversion_type = str(self._rate_conversion_type)
         stream = self._get_randomness_stream(
             decision_point, component, initializes_crn_attributes, rate_conversion_type
         )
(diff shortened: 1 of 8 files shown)


src/vivarium/framework/utilities.py
```diff
@@ -28,20 +29,63 @@ def to_yearly(value: TimeValue, time_step: Timedelta) -> TimeValue:
...
+            clip = True
+    else:
+        if exceeds_one.any():
+            probability[exceeds_one] = 1.0
+            clip = True
+    if clip:
+        logger.warning(
+            "The rate to probability conversion resulted in a probability greater than 1.0. "
+            "The probability has been clipped to 1.0 and indicates the rate is too high. "
+        )
```
rmudambi, line 66: Can't we do this more or less the same way as we had done before?

```suggestion
    probability: NumericArray = np.array(rate) * time_scaling_factor
    exceeds_one = probability > 1.0
    if exceeds_one.any():
        # Clip to 1.0 if the probability is greater than 1.0.
        probability[exceeds_one] = 1.0
        logger.warning(
            "The rate to probability conversion resulted in a probability greater than 1.0. "
            "The probability has been clipped to 1.0 and indicates the rate is too high. "
        )
```

src/vivarium/framework/utilities.py
```diff
@@ -28,20 +29,63 @@ def to_yearly(value: TimeValue, time_step: Timedelta) -> TimeValue:
...
+    probability: Sequence[float] | NumberLike, time_scaling_factor: float | int = 1.0
+) -> NumericArray:
+    """Function to convert a probability to a rate.
+
+    Parameters
+    ----------
+    probability
+        The probability to convert to a rate.
+    time_scaling_factor
+        The time factor in to scale the rates. This is uusually the time step.
```
rmudambi, line 80: Same issue in the comment as in the other method.

tests/framework/test_utilities.py
```diff
@@ -29,10 +31,19 @@ def test_to_yearly() -> None:
...
+@pytest.mark.parametrize(
+    "rate, time_scaling_factor",
+    [
+        (0.5, 1),
+        (0.25, 1.0),
+        (0.5, 0.5),
+        (0.25, 0.5),
+    ],
+)
+def test_rate_to_probability(rate: float, time_scaling_factor: float) -> None:
```
rmudambi, line 43: Need tests of calls where the `time_scaling_factor` is not provided for both directions.

src/vivarium/framework/utilities.py
albrja: Because of the data types rate can be (float vs series for instance) probability > 1.0 will be a bool or an array so I have to do the separate checks here unless there is a better way to write this.

tests/framework/test_utilities.py
```diff
@@ -46,6 +46,16 @@ def test_rate_to_probability(rate: float, time_scaling_factor: float) -> None:
     expected == prob
 
 
+@pytest.mark.parametrize("rate", [0.75, pd.Series([0.75, 0.5, 0.25]), 0.1])
+def test_rate_to_probability_default_time_scaling(rate: float) -> None:
+    expected = rate
+    prob = rate_to_probability(rate)
+    if isinstance(expected, float):
```
rmudambi, line 53: You won't need this branching pattern, because all calls will return an `np.array`.

rmudambi, line 53: You won't need this branching pattern, because all calls will return an `np.array`.

src/vivarium/framework/utilities.py
albrja: Fixed this now

src/vivarium/framework/randomness/manager.py
```diff
@@ -33,6 +33,7 @@ class RandomnessManager(Manager):
             "key_columns": [],
             "random_seed": 0,
             "additional_seed": None,
+            "rate_conversion_type": None,
```
rmudambi, line 36: We want to find a better place to configure this, since this doesn't intrinsically have anything to do with randomness.

src/vivarium/framework/randomness/manager.py
```diff
@@ -74,6 +76,10 @@ def setup(self, builder: Builder) -> None:
         pop_size = builder.configuration.population.population_size
         map_size = max(map_size, 10 * pop_size)
         self._key_mapping_ = IndexMap(self._key_columns, map_size)
+        if builder.configuration.randomness.rate_conversion_type is not None:
```
rmudambi, line 79: Rather than have this if/else, you can just set the default value to be "linear" rather than None

src/vivarium/framework/randomness/manager.py
```diff
@@ -158,13 +164,19 @@ def _get_randomness_stream(
                 f"Two separate places are attempting to create "
                 f"the same randomness stream for {decision_point}"
             )
+        if self._rate_conversion_type not in ["linear", "exponential"]:
```
rmudambi, line 167: Nit: I think this error should be thrown in `RandomnessStream.__init__()`

src/vivarium/framework/utilities.py
```diff
@@ -40,30 +42,36 @@ def rate_to_probability(
...
-    if clip:
-        logger.warning(
-            "The rate to probability conversion resulted in a probability greater than 1.0. "
-            "The probability has been clipped to 1.0 and indicates the rate is too high. "
-        )
+            logger.warning(
+                "The rate to probability conversion resulted in a probability greater than 1.0. "
+                "The probability has been clipped to 1.0 and indicates the rate is too high. "
+            )
+    else:
```
rmudambi, line 67: You should throw a `NotImplemented` error if the user provides a conversion type other than linear or exponential (is exponential the right name for the second conversion strategy?).

src/vivarium/framework/randomness/manager.py
albrja: Wouldn't this make it not a major release since the key would be required then?

src/vivarium/framework/randomness/stream.py
```diff
@@ -114,6 +115,13 @@ def __init__(
         """A key-index mapping with a vectorized hash and vectorized lookups."""
         self.initializes_crn_attributes = initializes_crn_attributes
         """A boolean indicating whether the stream is used to initialize CRN attributes."""
+        self.rate_conversion_type = rate_conversion_type
+        """The type of rate conversion to use when converting rates to probabilities."""
+        if self.rate_conversion_type not in ["linear", "exponential"]:
+            raise NotImplementedError(
```
rmudambi, line 121: Upon reflection I think these should be ValueErrors

src/vivarium/framework/randomness/stream.py
```diff
@@ -102,6 +102,7 @@ def __init__(
         # TODO [MIC-5452]: all resources should have a component
         component: Component | None = None,
         initializes_crn_attributes: bool = False,
+        rate_conversion_type: str = "linear",
```
rmudambi, line 105: Maybe this is better?
```python
rate_conversion_type: Literal["linear" | "exponential"] = "linear",
```

src/vivarium/framework/randomness/stream.py
```diff
@@ -114,6 +115,13 @@ def __init__(
         """A key-index mapping with a vectorized hash and vectorized lookups."""
         self.initializes_crn_attributes = initializes_crn_attributes
         """A boolean indicating whether the stream is used to initialize CRN attributes."""
+        self.rate_conversion_type = rate_conversion_type
+        """The type of rate conversion to use when converting rates to probabilities."""
```
rmudambi, line 119: You should say in this docstring what the allowed values are

src/vivarium/framework/randomness/manager.py
```diff
@@ -125,8 +136,10 @@ def get_randomness_stream(
             If another location in the simulation has already created a randomness stream
             with the same identifier.
         """
+        if rate_conversion_type is None:
+            rate_conversion_type = str(self._rate_conversion_type)
```
rmudambi, line 140: See if mypy will allow you to remove these lines altogether.

src/vivarium/framework/utilities.py
```diff
@@ -28,20 +29,76 @@ def to_yearly(value: TimeValue, time_step: Timedelta) -> TimeValue:
...
+
+    Parameters
+    ----------
+    rate
+        The rate to convert to a probability.
+    time_scaling_factor
+        The time factor in to scale the rates by. This is usually the time step.
+    conversion_type
+        The type of conversion to use. Default is "linear" for a simple multiplcation
+        of rate and time_scaling_factor. Other option is "exponential" which should be
```
rmudambi, line 47:
```suggestion
        of rate and time_scaling_factor. The other option is "exponential" which should be
```

src/vivarium/framework/randomness/manager.py
```diff
@@ -112,6 +119,10 @@ def get_randomness_stream(
             in the Common Random Number framework. These streams cannot be
             copied and should only be used to generate the state table columns
             specified in ``builder.configuration.randomness.key_columns``.
+        rate_conversion_type
+            The type of conversion to use. Default is "linear" for a simple
+            multiplication of rate and time_scaling_factor. Other option is
```
rmudambi, line 124:
```suggestion
            multiplication of rate and time_scaling_factor. The other option is
```

src/vivarium/framework/utilities.py
```diff
@@ -28,20 +29,76 @@ def to_yearly(value: TimeValue, time_step: Timedelta) -> TimeValue:
...
-    rate[rate > 250] = 250.0
-    probability: NumericArray = 1 - np.exp(-rate)
     return probability
 
 
-def probability_to_rate(probability: Sequence[float] | NumberLike) -> NumericArray:
-    probability = np.array(probability)
-    rate: NumericArray = -np.log(1 - probability)
+def probability_to_rate(
+    probability: Sequence[float] | NumberLike, time_scaling_factor: float | int = 1.0
```
rmudambi, line 84: We need the additional parameter for rate conversion type for this method as well

src/vivarium/framework/utilities.py
rmudambi, line 47: Can you ask the RT if "exponential" is the right way to label this method.

src/vivarium/framework/randomness/manager.py
albrja: Updated

src/vivarium/framework/randomness/manager.py
albrja: Decided we will leave this here for now

rmudambi (approved):

src/vivarium/framework/utilities.py
```diff
@@ -28,20 +29,91 @@ def to_yearly(value: TimeValue, time_step: Timedelta) -> TimeValue:
...
+) -> NumericArray:
+    """Converts a rate to a probability.
+
+    Parameters
+    ----------
+    rate
+        The rate to convert to a probability.
+    time_scaling_factor
+        The time factor in to scale the rates by. This is usually the time step.
+    conversion_type
```
stevebachmeier, line 45: nit: "rate_conversion_type"

src/vivarium/framework/utilities.py
```diff
@@ -28,20 +29,91 @@ def to_yearly(value: TimeValue, time_step: Timedelta) -> TimeValue:
...
+        The type of conversion to use. Default is "linear" for a simple multiplcation
+        of rate and time_scaling_factor. The other option is "exponential" which should be
+        used for continuous time event driven models.
+
+    Returns
+    -------
+        An array of floats representing the probability of the converted rates
+    """
+    if rate_conversion_type not in ["linear", "exponential"]:
```
stevebachmeier, line 54: Is there a reason to have this check both here and in randomness.stream.py?

src/vivarium/framework/utilities.py
```diff
@@ -28,20 +29,91 @@ def to_yearly(value: TimeValue, time_step: Timedelta) -> TimeValue:
...
+) -> NumericArray:
+    """Function to convert a probability to a rate.
+
+    Parameters
+    ----------
+    probability
+        The probability to convert to a rate.
+    time_scaling_factor
+        The time factor in to scale the probability by. This is usually the time step.
+    conversion_type
```
stevebachmeier, line 96: nit: rate_conversion_type

src/vivarium/framework/utilities.py
```diff
@@ -28,20 +29,91 @@ def to_yearly(value: TimeValue, time_step: Timedelta) -> TimeValue:
...
+        of rate and time_scaling_factor. The other option is "exponential" which should be
+        used for continuous time event driven models.
+
+    Returns
+    -------
+        An array of floats representing the rate of the converted probabilities
+    """
+    # NOTE: The default behavior for randomness streams is to use a rate that is already
+    # scaled to the time step which is why the default time scaling factor is 1.0.
+    if rate_conversion_type not in ["linear", "exponential"]:
```
stevebachmeier, line 107: same question here

stevebachmeier (approved):

src/vivarium/framework/utilities.py
albrja: Yes because this is a utility function so if it called outside of the randomness stream we still want the check to happen.

PR #621, opened 2025-05-20
typing results/test_context.py

hussain-jafari:
typing results/test_context.py

### DIFF
--- a/tests/framework/results/test_context.py
+++ b/tests/framework/results/test_context.py
@@ -480,10 +480,10 @@ def test_gather_results_partial_stratifications_in_results(
     for results, _measure, _formatter in ctx.gather_results(
         population, lifecycle_phase, mocked_event
     ):
-        if results is not None:
-            unladen_results = results.reset_index().query('familiar=="unladen_swallow"')
-            assert len(unladen_results) > 0
-            assert (unladen_results[VALUE_COLUMN] == 0).all()
+        assert results is not None
+        unladen_results = results.reset_index().query('familiar=="unladen_swallow"')
+        assert len(unladen_results) > 0
+        assert (unladen_results[VALUE_COLUMN] == 0).all()
 
 
 def test_gather_results_with_empty_pop_filter(mocked_event: Mock) -> None:
@@ -694,8 +694,7 @@ def test_to_observe(mocked_event: Mock, mocker: MockerFixture) -> None:
     for result, _measure, _updater in ctx.gather_results(
         population, lifecycle_phase, mocked_event
     ):
-        if result is not None:
-            assert not result.empty
+        assert result is not None and not result.empty
 
     # Extract the observation from the context and patch it to not observe
     observation = list(ctx.observations["collect_metrics"].values())[0][0]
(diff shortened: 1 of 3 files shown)


tests/framework/results/test_context.py
```diff
@@ -670,7 +696,8 @@ def test_to_observe(mocked_event, mocker):
     for result, _measure, _updater in ctx.gather_results(
         population, lifecycle_phase, mocked_event
     ):
-        assert not result.empty
+        if result is not None:
+            assert not result.empty
```
rmudambi, line 700: this should be
```python
assert result is not None and not result.empty
```

tests/framework/results/test_context.py
```diff
@@ -456,12 +482,13 @@ def test_gather_results_partial_stratifications_in_results(
     for results, _measure, _formatter in ctx.gather_results(
         population, lifecycle_phase, mocked_event
     ):
-        unladen_results = results.reset_index().query('familiar=="unladen_swallow"')
-        assert len(unladen_results) > 0
-        assert (unladen_results[VALUE_COLUMN] == 0).all()
+        if results is not None:
```
rmudambi, line 485: You should `assert results is not None`. Otherwise we won't detect if a bug is introduced that makes `result` None

tests/framework/results/test_context.py
```diff
@@ -365,8 +385,9 @@ def test_concatenating_observation_gather_results(mocked_event):
     for result, _measure, _updater in ctx.gather_results(
         population, lifecycle_phase, mocked_event
     ):
-        assert result.equals(filtered_pop[included_cols])
-        i += 1
+        if result is not None:
+            assert result.equals(filtered_pop[included_cols])
+            i += 1
```
rmudambi, line 390:
```suggestion
        assert result is not None and result.equals(filtered_pop[included_cols])
        i += 1
```

tests/framework/results/test_context.py
```diff
@@ -326,15 +345,16 @@ def test_adding_observation_gather_results(
...
-            math.isclose(actual_result, expected_result, rel_tol=0.0001)
-            for actual_result in result.values
-        )
-        i += 1
+        if result is not None:
+            assert all(
+                math.isclose(actual_result, expected_result, rel_tol=0.0001)
+                for actual_result in result.values
+            )
+            i += 1
```
rmudambi, line 353:
```suggestion
        assert result is not None and all(
            math.isclose(actual_result, expected_result, rel_tol=0.0001)
            for actual_result in result.values
        )
        i += 1
```

rmudambi (changes requested):

rmudambi (approved):

tests/framework/results/test_context.py
```diff
@@ -326,15 +345,15 @@ def test_adding_observation_gather_results(
     for result, _measure, _updater in ctx.gather_results(
         population, lifecycle_phase, mocked_event
     ):
-        assert all(
+        assert result is not None and all(
```
albrja, line 348: I think it would be better to have an assert for result is not None and a separate one that does the fuzzy checker instead of the math isclose but I guess that is not really the point of the pr

albrja (approved):

tests/framework/results/test_context.py
hussain-jafari: I agree - will update

PR #624, opened 2025-05-22
Albrja/mic-5621/observer-get-configuration

albrja:
### Overrides get_configuration method in observer
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5621

### Changes and notes
-overrides get_configuration method in Observer class
-updates test component Observer classes

### DIFF
--- a/src/vivarium/framework/results/observer.py
+++ b/src/vivarium/framework/results/observer.py
@@ -1,7 +1,7 @@
 """
 =========
 Observers
-=========
+=========`
 
 An observer is a component that is responsible for registering
 :class:`observations <vivarium.framework.results.observation.Observation>`
@@ -16,6 +16,8 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Any
+
+from layered_config_tree.main import LayeredConfigTree
 
 from vivarium import Component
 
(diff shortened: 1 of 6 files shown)


tests/helpers.py
```diff
@@ -18,10 +18,6 @@ class MockComponentA(Observer):
     def name(self) -> str:
         return self._name
 
-    @property
```
albrja, line 21: We actually use the defaults now. It is also very weird we use this class for component and observer tests.

src/vivarium/framework/results/observer.py
```diff
@@ -51,6 +53,10 @@ def get_configuration_name(self) -> str:
         """Return the name of a concrete observer for use in the configuration"""
         return self.name.split("_observer")[0]
 
+    def get_configuration(self, builder: Builder) -> LayeredConfigTree:
+        config = builder.configuration["stratification"][self.get_configuration_name()]
```
rmudambi, line 57: Why don't we do this?
```python
return builder.configuration.get_tree(["stratification", self.get_configuration_name()])
```

rmudambi (approved):
We should modify the observers in VPH now to take advantage of this functionality

src/vivarium/framework/results/observer.py
albrja: Mypy

albrja: It doesnt like it if I return a dict[str, Any] since the super class returns LCT or None. And if I ignore overriding the super class error it interprets the return you specific as just Any which we know is not really correct

albrja: Oh I misread this comment. This works and I have updated it.

patricktnast (approved):

PR #625, opened 2025-05-27
Albrja/mic-5831/update-package-pins

albrja:
### Update package pins for isort and pandas stubs
- *Category*: CI
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6097

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -75,7 +75,7 @@ if __name__ == "__main__":
 
     lint_requirements = [
         "black==22.3.0",
-        "isort==5.13.2",
+        "isort",
         "mypy",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -53,7 +53,7 @@
         "pyarrow",
         "dill",
         # Type stubs
-        "pandas-stubs",
+        "pandas-stubs==2.2.3.250308",
```
hussain-jafari, line 56: Do we want an equals here rather than a lower bound?

hussain-jafari (approved):

PR #627, opened 2025-07-10
Pin vivarium_build_utils

stevebachmeier:
### Description

- *Category*:  feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6194

### Changes and notes
 

There's two bits to this task:
1. How to force Jenkins to use the pinned/resolved version of vbu
2. How to force both Jenkins, site-packages, and local -e install of vbu
to use the pinned/resolved versions of the vbu makefiles.

The Jenkins bit is simpler to understand, though difficult to test: I changed
from the global and compile-time `@Library` call to a local and
at runtime call to `library` and pass in the calculated version. Calculating this
version was a bit tricky...

The Makefiles are tricker because...python packaging I think? But we basically
need to import and run a new vbu function to get the location of the makefiles.

### Testing

This is tough to test. I did some manual testing but we may not know
100% until we just release and start using it.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,7 @@
 **3.4.6 - 07/15/25**
 
-  - Pin vivarium_build_utils>=1.1.0,<2.0.0
+  - Support pinning of vivarium_build_utils; pin vivarium_build_utils>=1.1.0,<2.0.0
+  - Pin mypy<1.17.0
 
 **3.4.5 - 05/30/25**
 


Jenkinsfile
```diff
@@ -2,20 +2,49 @@
 found at https://github.com/ihmeuw/vivarium_build_utils
 Due to Jenkins convention, importable modules must be stored
 in the 'vars' folder.
-Jenkins needs to be configured globally to use the correct branch.
```
stevebachmeier, line 5: We are no longer using the `@Library` directive in favor of `library` directive because @Library is run at compile time and so can't support a dynamic value like the version number.

Jenkinsfile
```diff
@@ -2,20 +2,48 @@
...
-The "_" denotes that all modules will be imported from the shared library.
 */ 
-@Library("vivarium_build_utils") _
-reusable_pipeline(scheduled_branches: ["main"], upstream_repos: ["layered_config_tree"])
+
+// Determine the vivarium_build_utils version and load the library
+library "vivarium_build_utils@${getVBUVersion()}"
+
+// Run the reusable pipeline
+reusable_pipeline(
```
stevebachmeier, line 13: this hypothetically remains the only things that need to be changed on a per-repo bases.

The Jenkinsfile is obviously more complicate now but:
1. I've kept the number of code lines _above_ this call the same. The library call itself has changed but it's still just one line.
2. I at least moved all of the extra code down below in a Helper functions section.
3. The getVBUVersion function is downloading a python script from vbu main and so it's version-controlled in a single place.

Jenkinsfile
```diff
@@ -13,9 +15,14 @@ To configure the repo/branch go to:
...
-Updating the shared repo will take affect on the next pipeline invocation.
-The "_" denotes that all modules will be imported from the shared library.
 */ 
-@Library("vivarium_build_utils") _
-reusable_pipeline(scheduled_branches: ["main"], upstream_repos: ["layered_config_tree"])
+
+// Load the get_vbu_version function from vivarium_build_utils/boostrap/
+@Library("get_vbu_version@main") _
+// Now load the full vivarium_build_utils library at the expected version
+library("vivarium_build_utils@${get_vbu_version()}")
```
stevebachmeier, line 23: only one new non-comment line!

Jenkinsfile
```diff
@@ -13,9 +15,15 @@ To configure the repo/branch go to:
...
         * Credentials: SSH key to access the repo
-
-Updating the shared repo will take affect on the next pipeline invocation.
-The "_" denotes that all modules will be imported from the shared library.
 */ 
-@Library("vivarium_build_utils") _
-reusable_pipeline(scheduled_branches: ["main"], upstream_repos: ["layered_config_tree"])
+
+// Load the get_vbu_version function from vivarium_build_utils/boostrap/
+@Library("get_vbu_version@main") _
```
stevebachmeier, line 21: this is the reason this build is failing...we need to get this into vivarium_build_utils main branch!

albrja (approved):

patricktnast (approved):

PR #632, opened 2025-07-23
Add 'make build-env'

stevebachmeier:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6114

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+PACKAGE_NAME = $(notdir $(CURDIR))
+
 # Check if we're running in Jenkins
 ifdef JENKINS_URL
 	# Files are already in workspace from shared library
@@ -6,8 +8,6 @@ else
 	# For local dev, use the installed vivarium_build_utils package
 	MAKE_INCLUDES := $(shell python -c "import vivarium_build_utils.resources" &>/dev/null  && python -c "from vivarium_build_utils.resources import get_makefiles_path; print(get_makefiles_path())")
 endif
-
-PACKAGE_NAME = vivarium
 
 # Include makefiles from vivarium_build_utils
 ifneq ($(MAKE_INCLUDES),) # if not an empty string
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ ifdef JENKINS_URL
 	MAKE_INCLUDES := .
 else
 	# For local dev, use the installed vivarium_build_utils package
-	MAKE_INCLUDES := $(shell python -c "import vivarium_build_utils.resources" &>/dev/null  && python -c "from vivarium_build_utils.resources import get_makefiles_path; print(get_makefiles_path())")
+	MAKE_INCLUDES := $(shell python -c "import vivarium_build_utils.resources" &>/dev/null && python -c "from vivarium_build_utils.resources import get_makefiles_path; print(get_makefiles_path())")
 endif
 
 # Include makefiles from vivarium_build_utils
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -4,12 +4,57 @@ ifdef JENKINS_URL
 	MAKE_INCLUDES := .
 else
 	# For local dev, use the installed vivarium_build_utils package
-	MAKE_INCLUDES := $(shell python -c "from vivarium_build_utils.resources import get_makefiles_path; print(get_makefiles_path())")
+	MAKE_INCLUDES := $(shell python -c "import vivarium_build_utils.resources" &>/dev/null && python -c "from vivarium_build_utils.resources import get_makefiles_path; print(get_makefiles_path())")
```
rmudambi, line 7: Why do we need this change?

Makefile
```diff
@@ -4,12 +4,57 @@ ifdef JENKINS_URL
...
-PACKAGE_NAME = vivarium
+PACKAGE_NAME = $(notdir $(CURDIR))
 
 # Include makefiles from vivarium_build_utils
+ifneq ($(MAKE_INCLUDES),) # if not an empty string
 include $(MAKE_INCLUDES)/base.mk
 include $(MAKE_INCLUDES)/test.mk
+endif
+
+ifeq ($(MAKE_INCLUDES),) # empty string
```
rmudambi, line 18: Can't this be an `else` in the above `if` block?

rmudambi (approved):

Makefile
stevebachmeier: This is related to the comment below (about adding the paths to `MAKE_INCLUDES`) to this if/else block.

The intent is to define a MAKE_INCLUDES which is nothing more that a filepath to VBU's makefiles. Historically, the MAKE_INCLUDES is either "." (when on jenkins) or where vbu is installed (local); we then use that filepath to add the base.mk and test.mk (both defined in vbu).

The problem is that we now need to support this vivarium makefile where we don't necessarily have vbu installed. There is probably several ways to do this but I figured that if vbu is not installed then basically you don't want to include anything else (there's nothing to include!).

So what this change on line 7 does is first tries to import vbu and then if that is successful (&&) it will do what it was originally doing (import vbu and call `get_makefiles_path()`). We send stdout/stderr to null and so if it fails on the first import, then MAKE_INCLUDES is effectively an empty string.

We can then use that fact down below to decide whether or not to include base.mk or test.mk.

thoughts:
1. I could probably change this if/else block to something like an if/elif/else block and copy the `include`s in two of the three paths. That just seemed more complicated and more lines of code.
2. I don't love the double-import of vbu. I tried to get a single-line `shell python` call to work using try/except but simply couldn't make the syntax work.

Makefile
stevebachmeier: whoops, yeah I misread the comment - I'll do this

Makefile
stevebachmeier, line 7: Save the vbu check into `vbu_installed` variable

Makefile
```diff
@@ -1,15 +1,70 @@
...
+# 	Files are already in workspace from shared library
 	MAKE_INCLUDES := .
 else
-	# For local dev, use the installed vivarium_build_utils package
-	MAKE_INCLUDES := $(shell python -c "from vivarium_build_utils.resources import get_makefiles_path; print(get_makefiles_path())")
+# 	For local dev, use the installed vivarium_build_utils package if it exists
+# 	First check if we can import vivarium_build_utils and assign 'yes' or 'no'.
+# 	This tries to import the package in python and redirects stderr to the null device
+# 	If the import is successful (&&), it will print 'yes', otherwise (||) it will print 'no'.
+	VIVARIUM_BUILD_UTILS_AVAILABLE := $(shell python -c "import vivarium_build_utils" 2>/dev/null && echo "yes" || echo "no")
```
hussain-jafari, line 10: This is way more readable!

hussain-jafari (approved):

rmudambi (approved):

PR #633, opened 2025-07-29
Sbachmei/mic 5918/common dependencies

stevebachmeier:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5918

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,17 +41,10 @@ if __name__ == "__main__":
         long_description = f.read()
 
     install_requirements = [
+        "vivarium_dependencies[numpy,pandas,pyyaml,scipy,click,tables,loguru,pyarrow]",
         "vivarium_build_utils>=2.0.1,<3.0.0",
         "layered_config_tree>=2.1.0",
-        "numpy<2.0.0",
-        "pandas>=2.2.0",
-        "pyyaml>=5.1",
-        "scipy",
-        "click",
-        "tables",
         "networkx",
-        "loguru",
-        "pyarrow",
         "dill",
         # Type stubs
         "pandas-stubs<=2.2.3.250308",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -41,53 +41,33 @@
         long_description = f.read()
 
     install_requirements = [
+        "vivarium_dependencies[numpy,pandas,pyyaml,scipy,click,tables,loguru,pyarrow,networkx]",
```
stevebachmeier, line 44: Big question...should I be pinning vivarium_dependencies just like we pinned `vivarium_build_utils` below? It seems like we should not pin by default and then only if needed in the future?

albrja (approved):

patricktnast (approved):

setup.py
rmudambi: No, I don't think we want to pin.

setup.py
```diff
@@ -41,53 +41,33 @@
...
-        "pytest-cov",
-        "pytest-mock",
+        "vivarium_dependencies[pytest]",
         "vivarium_testing_utils",
     ]
 
     lint_requirements = [
-        "black==22.3.0",
+        "vivarium_dependencies[black]",
         "isort",
```
rmudambi, line 63: Shouldn't this be 
```suggestion
        "vivarium_dependencies[format, mypy]",
```
And if mypy truly needs to be pinned below `1.17.0`, we should have that pin in viv dependencies?

rmudambi (approved):

setup.py
stevebachmeier: I put in vivarium_dependencies the most common pins. If anything deviates from that then it needs to be explicitly defined in the repo. The most common pin for mypy is no pin, and somewhere along the way it was decided that this pin is needed in vivarium so we modified it.

PR #638, opened 2025-08-28
Refactor lifecycle system into a package

rmudambi:
### Description

- *Category*: refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6329

### Changes and notes
Refactor lifecycle system into a package

Question: we're inconsistent in whether we break interface and
manager into separate modules. What do we prefer?

### Testing
All tests pass


src/vivarium/framework/lifecycle/manager.py
```diff
@@ -0,0 +1,280 @@
...
+        self._make_constraint(method, allow_during)
+
+    def __repr__(self) -> str:
+        return f"LifeCycleManager(state={self.current_state})"
+
+    def __str__(self) -> str:
+        return str(self.lifecycle)
+
+
+class LifeCycleInterface(Interface):
```
rmudambi, line 214: We're inconsistent in whether we break interface and manager into separate modules. What do we prefer?

src/vivarium/framework/lifecycle/entities.py
rmudambi, line 1: Is there a better name for this file?

src/vivarium/framework/lifecycle/manager.py
albrja: I support being consistent.

albrja (approved):

patricktnast (approved):

PR #646, opened 2025-09-11
Sbachmei/mic 6379/attribute pipelines

stevebachmeier:
### Description

- *Category*:  feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4656

### Changes and notes
 
This implements the new AttributePipeline class and a few related classes.
It also adds the registration/interface methods.

I can't really think of a good way to DRY this up. I feel like there must be
a way to overload the super method calls or something like that, but I 
have't figured it out yet.

### Testing

Various tests added and all pass.

### DIFF
--- a/src/vivarium/framework/values/manager.py
+++ b/src/vivarium/framework/values/manager.py
@@ -92,32 +92,17 @@ class ValuesManager(Manager):
         """
         self.logger.debug(f"Registering value pipeline {value_name}")
         pipeline = self.get_value(value_name)
-        pipeline.set_attributes(
-            component,
+        self._configure_pipeline(
+            pipeline,
             source,
+            component,
+            requires_columns,
+            requires_values,
+            requires_streams,
+            required_resources,
             preferred_combiner,
             preferred_post_processor,
-            self,
         )
-
-        # The resource we add here is just the pipeline source.
-        # The value will depend on the source and its modifiers, and we'll
-        # declare that resource at post-setup once all sources and modifiers
-        # are registered.
-        dependencies = self._convert_dependencies(
-            source, requires_columns, requires_values, requires_streams, required_resources
-        )
-        self.resources.add_resources(pipeline.component, [pipeline.source], dependencies)
-
-        self.add_constraint(
-            pipeline._call,
-            restrict_during=[
-                lifecycle_states.INITIALIZATION,
-                lifecycle_states.SETUP,
-                lifecycle_states.POST_SETUP,
-            ],
-        )
-
         return pipeline
 
     def register_attribute_producer(
@@ -141,32 +126,17 @@ class ValuesManager(Manager):
         """
         self.logger.debug(f"Registering attribute pipeline {value_name}")
         pipeline = self.get_attribute(value_name)
-        pipeline.set_attributes(
-            component,
+        self._configure_pipeline(
+            pipeline,
             source,
+            component,
+            requires_columns,
+            requires_values,
+            requires_streams,
+            required_resources,
             preferred_combiner,
             preferred_post_processor,
-            self,
         )
-
-        # The resource we add here is just the pipeline source.
-        # The value will depend on the source and its modifiers, and we'll
-        # declare that resource at post-setup once all sources and modifiers
-        # are registered.
-        dependencies = self._convert_dependencies(
-            source, requires_columns, requires_values, requires_streams, required_resources
-        )
-        self.resources.add_resources(pipeline.component, [pipeline.source], dependencies)
-
-        self.add_constraint(
-            pipeline._call,
-            restrict_during=[
-                lifecycle_states.INITIALIZATION,
-                lifecycle_states.SETUP,
-                lifecycle_states.POST_SETUP,
-            ],
-        )
-
         return pipeline
 
     def register_value_modifier(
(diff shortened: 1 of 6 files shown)


src/vivarium/framework/values/manager.py
```diff
@@ -153,17 +186,72 @@ def register_value_modifier(
...
+        """Marks a ``Callable`` as the modifier of a named attribute.
+
+        Parameters
+        ----------
+        value_name :
+            The name of the dynamic attribute pipeline to be modified.
+        modifier :
+            A function that modifies the source of the dynamic attribute pipeline
+            when called. If the pipeline has a ``replace_combiner``, the
+            modifier should accept the same arguments as the pipeline source
```
albrja, line 223: I'm not sure if you are just being consistent across docstrings but I don't like "should accept" when it really HAS to have the same arguments.

albrja (approved):

src/vivarium/framework/values/manager.py
```diff
@@ -153,17 +186,72 @@ def register_value_modifier(
...
+            component,
+            requires_columns,
+            requires_values,
+            requires_streams,
+            required_resources,
+        )
 
-        dependencies = self._convert_dependencies(
-            modifier, requires_columns, requires_values, requires_streams, required_resources
+    def register_attribute_modifier(
```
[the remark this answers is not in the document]
stevebachmeier: Good catch - that actually comes in the next PR. (I just forgot here but then actually needed it there).

src/vivarium/framework/values/manager.py
stevebachmeier: Updated "should accept" to "must have"

src/vivarium/framework/values/pipeline.py
```diff
@@ -232,8 +266,111 @@ def set_attributes(
...
+        Returns
+        -------
+            A pd.DataFrame of attributes for the simulants in `index`.
+
+        Raises
+        ------
+        DynamicValueError
+            If the pipeline is invoked without a source set.
+        """
+        # NOTE: must pass index in as arg (NOT kwarg!) to match signature of parent Pipeline._call()
```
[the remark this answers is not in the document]
stevebachmeier: As discussed, yes it is true

src/vivarium/framework/values/pipeline.py
```diff
@@ -232,8 +266,111 @@ def set_attributes(
...
+        attribute = self._call(index, skip_post_processor=skip_post_processor)
+        if not isinstance(attribute, pd.DataFrame):
+            raise DynamicValueError(
```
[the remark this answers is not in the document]
stevebachmeier: leave as is

src/vivarium/framework/values/post_processors.py
```diff
@@ -18,6 +18,13 @@ def __call__(self, value: Any, manager: ValuesManager) -> Any:
         ...
 
 
+class AttributePostProcessor(Protocol):
```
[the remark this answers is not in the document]
stevebachmeier: I'll look into it

src/vivarium/framework/values/pipeline.py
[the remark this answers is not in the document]
patricktnast: is this in only in line 333, or do components that call this pipeline also need to pass the index as an arg?

src/vivarium/framework/values/pipeline.py
```diff
@@ -232,8 +266,111 @@ def set_attributes(
         manager
             The simulation values manager.
         """
-        self.component = component
+        self._set_common_attributes(component, combiner, post_processor, manager)
         self.source = ValueSource(self, source, component)
+
+    def _set_common_attributes(
```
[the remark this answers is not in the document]
stevebachmeier: Irrelevant now that we've done away with AttributeSource type.

src/vivarium/framework/values/manager.py
```diff
@@ -50,7 +61,10 @@ def on_post_setup(self, _event: Event) -> None:
         """Finalizes dependency structure for the pipelines."""
         # Unsourced pipelines might occur when generic components register
         # modifiers to values that aren't required in a simulation.
-        unsourced_pipelines = [p for p, v in self._pipelines.items() if not v.source]
+        unsourced_pipelines = [p for p, v in self._value_pipelines.items() if not v.source]
```
patricktnast, line 64: can you not just use _all_pipelines here?

patricktnast: or even self.items(), i guess

src/vivarium/framework/values/post_processors.py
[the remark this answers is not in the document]
stevebachmeier: I think I like it. It doesn't add much in terms of lines of code and does provide nice type-hinting. The only negative is that hovering over the type hint doesn't really shed any light as to the protocol being defined.

src/vivarium/framework/values/manager.py
stevebachmeier: Ah, good catch, i wrote this before I'd implemented all_pipelines.

patricktnast (approved):

patricktnast (approved):

src/vivarium/framework/values/manager.py
```diff
@@ -138,12 +131,9 @@ def register_attribute_producer(
             pipeline,
             source,
             component,
-            requires_columns,
-            requires_values,
-            requires_streams,
-            required_resources,
-            preferred_combiner,
-            preferred_post_processor,
+            required_resources=required_resources,
```
rmudambi, line 134: Nit: why are these now kwargs but not the args above?

src/vivarium/framework/values/pipeline.py
```diff
@@ -266,21 +246,11 @@ def set_attributes(
...
-        combiner: ValueCombiner,
-        post_processor: PostProcessor | AttributePostProcessor | None,
-        manager: ValuesManager,
-    ) -> None:
-        """Set the common attributes for all pipeline types."""
         self.component = component
         self._combiner = combiner
         self.post_processor = post_processor
         self._manager = manager
+        self.source = ValueSource(self, source, component)
```
rmudambi, line 253: Nit: wasn't this the first attribute set before?

rmudambi (approved):
Looks great!

src/vivarium/framework/values/manager.py
stevebachmeier: Because the args are in the correct position but then we skip a bunch of optional args (requires_*) so need to add the keyword.

src/vivarium/framework/values/pipeline.py
stevebachmeier: Second, actually. I moved it back.

PR #647, opened 2025-09-12
Sbachmei/mic 6380/convert pipelines to attribute pipelines

stevebachmeier:
### Description

- *Category*:  refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6380

### Changes and notes
 

This does a few things:
1. (primary goal) converts the example pipelines to attribute pipelines
2. Adds the `register_attribute_modifier` interface method I forgot in the previous PR
3. Adds a `register_attribute_rate_producer` method. I wasn't going to do this at first
but realized I couldn't easily reuse the `register_rate_producer` since that immediately
calls `get_value()` whereas we need to be calling `get_attribute()`.

NOTE: I did NOT convert the one and only vivarium-level pipeline, "simulant_step_size".
I think some more discussion is needed about exactly what AttributePipelines (and
their sources and modifiers) need to return. The only thing potentially keeping this from
just being an AttributePipeline is that the source isn't returning a pd.Series, but rather a 
_list_ of pd.Series! But since the post-processor is a list combiner, the _callable_ does
indeed return a pd.Series. It should be easy to convert if we decide to. In fact, I tried it
and only one test failed (because the mocking was no longer being done correctly).

### Testing

Tests pass. 

NOTE: I did NOT run the boids example to make sure it works. I'm not really even sure how.

### DIFF
--- a/src/vivarium/examples/boids/forces.py
+++ b/src/vivarium/examples/boids/forces.py
@@ -32,9 +32,9 @@ class Force(Component, ABC):
         self.config = builder.configuration[self.__class__.__name__.lower()]
         self.max_speed = builder.configuration.movement.max_speed
 
-        self.neighbors = builder.value.get_value("neighbors")
+        self.neighbors = builder.value.get_attribute("neighbors")
 
-        builder.value.register_value_modifier(
+        builder.value.register_attribute_modifier(
             "acceleration",
             modifier=self.apply_force,
             required_resources=self.columns_required + [self.neighbors],
@@ -47,6 +47,8 @@ class Force(Component, ABC):
     def apply_force(self, index: pd.Index[int], acceleration: pd.DataFrame) -> pd.DataFrame:
         neighbors = self.neighbors(index)
         pop = self.population_view.get(index)
+        if not isinstance(neighbors, pd.Series):
+            raise ValueError("Neighbors must be a pd.Series of ints")
         pairs = self._get_pairs(neighbors, pop)
 
         raw_force = self.calculate_force(pairs)
@@ -68,7 +70,7 @@ class Force(Component, ABC):
     def calculate_force(self, neighbors: pd.DataFrame) -> pd.DataFrame:
         pass
 
-    def _get_pairs(self, neighbors: pd.Series[int], pop: pd.DataFrame) -> pd.DataFrame:
+    def _get_pairs(self, neighbors: pd.Series[Any], pop: pd.DataFrame) -> pd.DataFrame:
         pairs = (
             pop.join(neighbors.rename("neighbors"))
             .reset_index()
--- a/src/vivarium/examples/boids/forces.py
+++ b/src/vivarium/examples/boids/forces.py
@@ -70,7 +70,6 @@ class Force(Component, ABC):
         pass
 
     def _get_pairs(self, neighbors: pd.Series[int | float], pop: pd.DataFrame) -> pd.DataFrame:
-        breakpoint()
         pairs = (
             pop.join(neighbors.rename("neighbors"))
             .reset_index()
--- a/src/vivarium/examples/boids/forces.py
+++ b/src/vivarium/examples/boids/forces.py
@@ -70,7 +70,8 @@ class Force(Component, ABC):
     def calculate_force(self, neighbors: pd.DataFrame) -> pd.DataFrame:
         pass
 
-    def _get_pairs(self, neighbors: pd.Series[Any], pop: pd.DataFrame) -> pd.DataFrame:
+    def _get_pairs(self, neighbors: pd.Series[int | float], pop: pd.DataFrame) -> pd.DataFrame:
+        breakpoint()
         pairs = (
             pop.join(neighbors.rename("neighbors"))
             .reset_index()
--- a/src/vivarium/examples/boids/forces.py
+++ b/src/vivarium/examples/boids/forces.py
@@ -37,6 +37,7 @@ class Force(Component, ABC):
             "acceleration",
             modifier=self.apply_force,
             required_resources=self.columns_required + [self.neighbors],
+            component=self,
         )
 
     ##################################
--- a/src/vivarium/examples/boids/forces.py
+++ b/src/vivarium/examples/boids/forces.py
@@ -1,7 +1,6 @@
 
 from __future__ import annotations
 from abc import ABC, abstractmethod
-from typing import Any
 
 import numpy as np
 import pandas as pd
@@ -15,7 +14,7 @@ class Force(Component, ABC):
     # Properties #
     ##############
     @property
-    def configuration_defaults(self) -> dict[str, Any]:
+    def configuration_defaults(self) -> dict[str, dict[str, float]]:
         return {
             self.__class__.__name__.lower(): {
                 "max_force": 0.03,
--- a/src/vivarium/examples/boids/movement.py
+++ b/src/vivarium/examples/boids/movement.py
@@ -32,7 +32,7 @@ class Movement(Component):
         self.config = builder.configuration
 
         self.acceleration = builder.value.register_attribute_producer(
-            "acceleration", source=self.base_acceleration
+            "acceleration", source=self.base_acceleration, component=self
         )
 
     ##################################
--- a/src/vivarium/examples/boids/movement.py
+++ b/src/vivarium/examples/boids/movement.py
@@ -31,7 +31,7 @@ class Movement(Component):
     def setup(self, builder: Builder) -> None:
         self.config = builder.configuration
 
-        self.acceleration = builder.value.register_value_producer(
+        self.acceleration = builder.value.register_attribute_producer(
             "acceleration", source=self.base_acceleration
         )
 
@@ -66,6 +66,8 @@ class Movement(Component):
         acceleration = self.acceleration(event.index)
 
         # Accelerate and limit velocity
+        if not isinstance(acceleration, pd.DataFrame):
+            raise ValueError("Acceleration must be a pd.DataFrame")
         pop[["vx", "vy"]] += acceleration.rename(columns=lambda c: f"v{c}")
         speed = np.sqrt(np.square(pop.vx) + np.square(pop.vy))
         velocity_scaling_factor = np.where(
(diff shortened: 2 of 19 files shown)


src/vivarium/examples/disease_model/disease.py
```diff
@@ -34,13 +34,13 @@ def setup(self, builder: Builder) -> None:
...
         self.base_rate = lambda index: pd.Series(rate, index=index)
-        self.joint_population_attributable_fraction = builder.value.register_value_producer(
+        self.joint_population_attributable_fraction = builder.value.register_attribute_producer(
             f"{self.rate_name}.population_attributable_fraction",
             source=lambda index: [pd.Series(0.0, index=index)],
             preferred_combiner=list_combiner,
             preferred_post_processor=union_post_processor,
         )
-        self.transition_rate = builder.value.register_rate_producer(
+        self.transition_rate = builder.value.register_attribute_rate_producer(
```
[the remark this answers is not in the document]
stevebachmeier: it's a totally new method.

tests/framework/test_time.py
```diff
@@ -512,7 +512,7 @@ def test_move_simulants_to_end(base_config: LayeredConfigTree) -> None:
     assert step_modifier_component.ts_pipeline_value.index.equals(full_pop_index)
     assert np.all(
         sim._clock.simulant_next_event_times(evens)
-        == sim._clock.stop_time + sim._clock.minimum_step_size
+        == sim._clock.stop_time + sim._clock.minimum_step_size  # type: ignore [operator]
```
stevebachmeier, line 515: I just don't want to think about this here - not really sure the best way to handle the different types in stop time vs step_size. Happy to look closer if people want though

tests/framework/test_values.py
```diff
@@ -256,25 +256,36 @@ def bad_attribute_source(index: pd.Index[int]) -> pd.DataFrame:
...
-    pipeline = manager.register_attribute_producer(
-        "test_attribute", source=bad_attribute_source
+    series_pipeline = manager.register_attribute_producer(
+        "test_series_attribute", source=series_attribute_source
+    )
+    dataframe_pipeline = manager.register_attribute_producer(
+        "test_dataframe_attribute", source=dataframe_attribute_source
+    )
+    bad_pipeline = manager.register_attribute_producer(
+        "test_bad_attribute", source=bad_attribute_source  # type: ignore [arg-type]
```
stevebachmeier, line 278: Ignoring b/c this is precisely what we're testing here via pytest.raises

src/vivarium/framework/values/manager.py
```diff
@@ -676,3 +736,8 @@ def get_value(self, name: str) -> Pipeline:
 
         """
         return self._manager.get_value(name)
+
+    # TODO: [MIC-5452] Remove this method (attributes should be obtained via population views)
+    def get_attribute(self, name: str) -> AttributePipeline:
```
stevebachmeier, line 741: NOTE: this will NOT stick around for the long haul. The way we intend for users to get attributes is via the population_view.get() method. This is just temporarily needed in a few loations (pytests and examples) until that interface gets updated.

src/vivarium/framework/results/manager.py
```diff
@@ -26,11 +26,6 @@
     from vivarium.framework.engine import Builder
 
 
-class SourceType(Enum):
```
stevebachmeier, line 29: Not needed anymore after Rajan's recent updates

src/vivarium/framework/values/manager.py
```diff
@@ -513,19 +513,15 @@ def register_rate_producer(
         self,
         rate_name: str,
         source: Callable[..., Any],
-        # TODO [MIC-5452]: all calls should have a component
-        component: Component | None = None,
```
stevebachmeier, line 517: register_rate_producer is weird b/c it's an older interface metho (i.e. I didn't want to remove the old-style requires_* args, but it calls the new register_attribute_producer method which did have those stripped out.

I just updated it to require only required_resources and luckily I didn't need to fix much.

tests/framework/test_time.py
```diff
@@ -175,19 +176,29 @@ def __init__(
...
             f"test_rate_{self.name}",
             source=lambda idx: pd.Series(1.75, index=idx),
             preferred_post_processor=rescale_post_processor,
+            component=self,
         )
 
     def on_time_step(self, event: Event) -> None:
         self.ts_pipeline_value = self.rate_pipeline(event.index)
 
+    def __hash__(self) -> int:
```
stevebachmeier, line 192: This is needed because StepModifierWithRatePipeline gets used as a resource, and the class ResourceGroup needs to[ hash the resources during initialization](https://github.com/ihmeuw/vivarium/blob/main/src/vivarium/framework/resource/group.py#L45)

src/vivarium/examples/boids/movement.py
```diff
@@ -31,8 +31,8 @@ class Movement(Component):
     def setup(self, builder: Builder) -> None:
         self.config = builder.configuration
 
-        self.acceleration = builder.value.register_value_producer(
-            "acceleration", source=self.base_acceleration
+        self.acceleration = builder.value.register_attribute_producer(
+            "acceleration", source=self.base_acceleration, component=self
```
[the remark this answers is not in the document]
stevebachmeier: The _new_ interface methods now require component. The old ones (i.e. register_value_*) don't require it (though there's a ticket somewhere not related to population to require them)

src/vivarium/examples/boids/forces.py
```diff
@@ -47,6 +47,8 @@ def setup(self, builder: Builder) -> None:
     def apply_force(self, index: pd.Index[int], acceleration: pd.DataFrame) -> pd.DataFrame:
         neighbors = self.neighbors(index)
         pop = self.population_view.get(index)
+        if not isinstance(neighbors, pd.Series):
```
rmudambi, line 50: This is because it could be a dataframe? Do we need to narrow our focus even more to have a `SeriesAttributePipeline` and a `DataFrameAttributePipeline`?

src/vivarium/examples/boids/forces.py
```diff
@@ -68,7 +70,7 @@ def apply_force(self, index: pd.Index[int], acceleration: pd.DataFrame) -> pd.Da
     def calculate_force(self, neighbors: pd.DataFrame) -> pd.DataFrame:
         pass
 
-    def _get_pairs(self, neighbors: pd.Series[int], pop: pd.DataFrame) -> pd.DataFrame:
+    def _get_pairs(self, neighbors: pd.Series[int | float], pop: pd.DataFrame) -> pd.DataFrame:
```
rmudambi, line 73: I think these files are excluded from mypy, FYI.

src/vivarium/framework/values/pipeline.py
```diff
@@ -247,18 +247,18 @@ def set_attributes(
             The simulation values manager.
         """
         self.component = component
-        self.source = ValueSource(self, source, component)
         self._combiner = combiner
         self.post_processor = post_processor
         self._manager = manager
+        self.source = ValueSource(self, source, component)
```
rmudambi, line 253: This is a merge conflict I think. It should still be assigned above.

src/vivarium/framework/values/post_processors.py
```diff
@@ -25,7 +25,9 @@ def __call__(self, value: Any, manager: ValuesManager) -> pd.Series[Any] | pd.Da
         ...
 
 
-def rescale_post_processor(value: NumberLike, manager: ValuesManager) -> NumberLike:
+def rescale_post_processor(
+    value: pd.Series[Any] | pd.DataFrame, manager: ValuesManager
+) -> pd.Series[Any] | pd.DataFrame:
```
rmudambi, line 30: Are these types really Any? They're not `[int | float]`? In fact the return type looks like it will always be a float, since you're dividing by 365.0

src/vivarium/framework/time.py
```diff
@@ -14,14 +14,15 @@
...
-from collections.abc import Callable
+from collections.abc import Callable, Sequence
 from functools import partial
 from typing import TYPE_CHECKING, Any
 
 import numpy as np
 import pandas as pd
 
 from vivarium.framework.lifecycle import lifecycle_states
+from vivarium.framework.resource import Resource
```
rmudambi, line 25: Nit: Should you put this in the TYPE CHECKING block?

tests/framework/test_time.py
rmudambi: How are other components hashed in that line?

tests/framework/test_time.py
rmudambi: We ignore this everywhere else it comes up iirc

tests/framework/test_values.py
```diff
@@ -269,27 +269,44 @@ def bad_attribute_source(index: pd.Index[int]) -> pd.DataFrame:
...
     with pytest.raises(
         DynamicValueError,
         match=(
-            f"The dynamic attribute pipeline for {pipeline.name} returned a {type('foo')} "
+            f"The dynamic attribute pipeline for {bad_pipeline.name} returned a {type('foo')} "
             "but pd.Series' or pd.DataFrames are expected for attribute pipelines."
         ),
     ):
-        pipeline(index)
+        bad_pipeline(index)
```
rmudambi, line 309: Maybe create a pipeline with `source=str_attribute_source` and a post-processor that turns it into a Series and show that that doesn't fail?

rmudambi (approved):

src/vivarium/examples/boids/forces.py
stevebachmeier: Well, kinda but I think of it in the opposite direction. it's just that the next call `self._get_pairs(neighbors, pop)` type hints neighbors as being a series.

src/vivarium/framework/values/post_processors.py
stevebachmeier: Oh, I guess for this particular post processor yeah probably just floats?

src/vivarium/framework/time.py
stevebachmeier: We're not very strict about that, although I'd like to be.

tests/framework/test_time.py
stevebachmeier: Not sure I understand the question

stevebachmeier: I need to look into this

patricktnast (approved):

PR #650, opened 2025-09-16
Sbachmei/mic 6380/convert pipelines to attribute pipelines

stevebachmeier:
### Description

- *Category*:  refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6380

### Changes and notes
 

This primarily focuses on converting all get_value_* calls to get_attribute_* calls.

### Testing

Tests pass.

NOTE: I did NOT run the boids example to make sure it works. I'm not really even sure how.

### DIFF
--- a/src/vivarium/framework/values/manager.py
+++ b/src/vivarium/framework/values/manager.py
@@ -729,20 +729,3 @@ class ValuesInterface(Interface):
 
         """
         return self._manager.get_value(name)
-
-    def get_attribute(self, name: str) -> AttributePipeline:
-        """Retrieve the pipeline representing the named attribute.
-
-        Parameters
-        ----------
-        name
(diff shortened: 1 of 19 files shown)


src/vivarium/framework/values/manager.py
```diff
@@ -513,19 +513,15 @@ def register_rate_producer(
         self,
         rate_name: str,
         source: Callable[..., Any],
-        # TODO [MIC-5452]: all calls should have a component
```
stevebachmeier, line 516: register_rate_producer is weird b/c it's an older interface metho (i.e. I didn't want to remove the old-style requires_* args, but it calls the new register_attribute_producer method which did have those stripped out.

I just updated it to require only `required_resources` and luckily I didn't need to fix much.

PR #653, opened 2025-09-19
support passing column names as an AttributePipeline source

stevebachmeier:
### Description

- *Category*:  feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6382

### Testing

Added one new test; all pass

### DIFF
--- a/src/vivarium/framework/values/interface.py
+++ b/src/vivarium/framework/values/interface.py
@@ -46,7 +46,7 @@ class ValuesInterface(Interface):
         self,
         value_name: str,
         source: Callable[..., Any],
-        # TODO [MIC-5452]: all calls should have a component
+        # TODO [MIC-6433]: all calls should have a component
         component: Component | None = None,
         requires_columns: Iterable[str] = (),
         requires_values: Iterable[str] = (),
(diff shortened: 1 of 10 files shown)


src/vivarium/framework/values/interface.py
```diff
@@ -160,7 +163,7 @@ def register_attribute_producer(
     def register_rate_producer(
         self,
         rate_name: str,
-        source: Callable[..., Any],
+        source: Callable[[pd.Index[int]], Any] | list[str],
```
stevebachmeier, line 166: We previously agreed that "rate producers" are by definition special cases of attribute producers and so the source needs to be a callable or list of cols.

src/vivarium/framework/values/pipeline.py
```diff
@@ -46,14 +47,26 @@ def __init__(
...
+    def __call__(self, population_mgr: PopulationManager, *args: Any, **kwargs: Any) -> Any:
         if not self._source:
             raise DynamicValueError(
                 f"The dynamic {self._pipeline_type} pipeline for {self.name} has no source."
                 " This likely means you are attempting to modify a value that"
                 " hasn't been created."
             )
-        return self._source(*args, **kwargs)
+
+        if callable(self._source):
```
stevebachmeier, line 58: this is the meat of the whole thing

PR #656, opened 2025-09-22
Sbachmei/mic 6384/convert state table columns to attribute pipelines

stevebachmeier:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6384

### Changes and notes
 
This is kind of a weird PR, as I thought there'd be more to it. The goals
is to convert all vivarium state table cols into AttributePipelines without
removing them from the state table (that will come later).

I thought that this would include adding all state table cols to the creating
component's `columns_created` attr, but it turns out that in all cases
they were already there.

I did need to add a bit so Managers auto-register them must like components.

Note: vivarium framework has four "official" AttributePipelines now:

- `tracked` (PopulationManager.colums_created)
- `next_event_time` and `step_size` (SimulationClock.columns_created)
- `simulant_step_size` (explicitly registered in SimulantClock.setup() method)

### DIFF
--- a/src/vivarium/framework/time.py
+++ b/src/vivarium/framework/time.py
@@ -108,7 +108,7 @@ class SimulationClock(Manager):
         self._minimum_step_size: ClockStepSize | None = None
         self._standard_step_size: ClockStepSize | None = None
         self._clock_step_size: ClockStepSize | None = None
-        self._pop_view: PopulationView | None = None
+        self._population_view: PopulationView | None = None
         self._pipeline_name = "simulant_step_size"
         # TODO: Delegate this functionality to "tracked" or similar when appropriate
         self._simulants_to_snooze = pd.Index([])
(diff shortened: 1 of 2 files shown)


src/vivarium/framework/time.py
```diff
@@ -108,12 +108,13 @@ def __init__(self) -> None:
...
         self._standard_step_size: ClockStepSize | None = None
         self._clock_step_size: ClockStepSize | None = None
-        self._individual_clocks: PopulationView | None = None
+        self._pop_view: PopulationView | None = None
         self._pipeline_name = "simulant_step_size"
         # TODO: Delegate this functionality to "tracked" or similar when appropriate
         self._simulants_to_snooze = pd.Index([])
 
     def setup(self, builder: "Builder") -> None:
+        super().setup(builder)
```
stevebachmeier, line 117: Same as the PopulationManager - you need to super setup to get the automatic registering of columns_created.

patricktnast (approved):

PR #724, opened 2026-01-07
Sbachmei/mic 6606/multiple initializers

stevebachmeier:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6606

### Changes and notes
 
The fundamental goal with this PR is to allow Components (and Managers) 
to register any number of initializer methods; we no longer simply assume
that each component has a single one called `on_initialize_simulants`.

The reason for this is that there are occasions where a component's
single initializer introduces a cycle in the DAG and we need to thus
separate the resource groups per initializer.

### DIFF
--- a/docs/source/tutorials/boids.rst
+++ b/docs/source/tutorials/boids.rst
@@ -47,8 +47,8 @@ Imports
 +++++++
 
 .. literalinclude:: ../../../src/vivarium/examples/boids/population.py
-   :lines: 1-6
-   :linenos:
+   :start-after: # docs-start: imports
+   :end-before: # docs-end: imports
 
 `NumPy <http://www.numpy.org/>`_ is a library for doing high performance
 numerical computing in Python. `pandas <https://pandas.pydata.org/>`_ is a set
@@ -78,26 +78,44 @@ configuration information. Components typically expose the values they use in
 the ``configuration_defaults`` class attribute.
 
 .. literalinclude:: ../../../src/vivarium/examples/boids/population.py
-   :lines: 13-17
+   :start-after: # docs-start: configuration_defaults
+   :end-before: # docs-end: configuration_defaults
    :dedent: 4
-   :linenos:
-   :lineno-start: 13
 
 We'll talk more about configuration information later. For now observe that
 we're exposing a set of possible colors for our boids.
 
-The ``columns_created`` property
-++++++++++++++++++++++++++++++++
+The ``setup`` method
+++++++++++++++++++++
+
+Almost every component in Vivarium will have a setup method. The setup method
+gives the component access to an instance of the
+:class:`~vivarium.framework.engine.Builder` which exposes a handful of tools
+to help build components. The simulation framework is responsible for calling
+the setup method on components and providing the builder to them. We'll
+explore these tools that the builder provides in detail as we go.
 
 .. literalinclude:: ../../../src/vivarium/examples/boids/population.py
-   :lines: 18
+   :start-after: # docs-start: setup
+   :end-before: # docs-end: setup
    :dedent: 4
-   :linenos:
-   :lineno-start: 18
 
-The ``columns_created`` property tells Vivarium what columns (or "attributes")
-the component will add to the population table.
-See the next section for where we actually create these columns.
+Our setup method is pretty simple: we just save the configured colors for later use,
+get a randomness stream, and register some private columns. Regarding the colors,
+note that the component is accessing the subsection of the configuration that it cares about.
+The full simulation configuration is available from the builder as
+``builder.configuration``. You can treat the configuration object just like
+a nested python
+`dictionary <https://docs.python.org/3/tutorial/datastructures.html#dictionaries>`_
+that's been extended to support dot-style attribute access. Our access here
+mirrors what's in the ``configuration_defaults`` at the top of the class
+definition.
+
+Note that the setup method is registering two columns (``color`` and
+``entrance_time``) as private columns with a method called ``on_initialize_simulants``
+as their initializer and the randomness stream as their dependency. This tells 
+Vivarium what columns (or "attributes") the component will add to the population table
+and how. See the next section for where we actually create these columns.
 
 .. note::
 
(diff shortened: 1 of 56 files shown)


docs/source/tutorials/boids.rst
```diff
@@ -47,8 +47,8 @@ Imports
 +++++++
 
 .. literalinclude:: ../../../src/vivarium/examples/boids/population.py
```
stevebachmeier, line 49: I hate these hard-coded line numbers. I wouldn't have done anything to fix it here except that we started calling line numbers that are larger than the actual length of the file and so I took the opportunity to make it a bit more dynamic. 

It does mean plopping comments all over the actual code to be referenced, but I figured that's fine for example code. We wouldn't do this in production code.

src/vivarium/framework/population/manager.py
```diff
@@ -285,6 +282,10 @@ def setup(self, builder: Builder) -> None:
         self.resources = builder.resources
         self._add_constraint = builder.lifecycle.add_constraint
         self._get_attribute_pipelines = builder.value.get_attribute_pipelines()
+        self._register_attribute_producer = builder.value.register_attribute_producer
+        self._get_current_component_or_manager = (
```
stevebachmeier, line 301: This can indeed be a manager (e.g. the clock manager does have an initializer method (though it doens't initialize columns))

src/vivarium/framework/population/manager.py
```diff
@@ -385,38 +386,6 @@ def _get_view(
...
-            The state table columns that the given initializer provides the
-            initial state information for.
-        required_resources
-            The resources that the initializer requires to run. Strings are
-            interpreted as population attribute names.
-        """
-        if isinstance(creates_columns, str):
-            creates_columns = [creates_columns]
-
-        self._initializer_components.add(component.on_initialize_simulants, creates_columns)
```
stevebachmeier, line 411: I didn't like this indirection so just moved this bit of logic right into the new `register_initializer() method`

src/vivarium/framework/population/manager.py
```diff
@@ -464,39 +433,77 @@ def _create_simulants(
...
+                column_name,
+                source=[column_name],
+                source_is_private_column=True,
+            )
+
+        # Register private column metadata
+        self._private_column_metadata[component.name].extend(columns)
+
+        # Register the initializer as a resource
+        self._initializer_components.add(initializer, columns)
```
stevebachmeier, line 516: This bit is from the old `register_simulant_initializer()` method I deleted

src/vivarium/framework/resource/group.py
```diff
@@ -60,22 +61,18 @@ def __init__(
...
 
     @property
     def names(self) -> list[str]:
         """The long names (including type) of all resources in this group."""
         return list(self.resources)
 
-    @property
-    def initializer(self) -> Callable[[SimulantData], None]:
-        """The method that initializes this group of resources."""
-        return self.component.on_initialize_simulants
```
stevebachmeier, line 77: we can't assume this anymore! So we pass initializer in directory to ResourceGroup init

src/vivarium/framework/resource/interface.py
```diff
@@ -64,33 +66,31 @@ def add_resources(
...
+        columns
+            The state table columns that the given initializer provides the initial
+            state information for.
+        dependencies
+            The resources that the initializer requires to run. Strings are interpreted
+            as attributes.
         """
-        if isinstance(resources, str):
-            resources = [resources]
-        columns = [Column(col, component) for col in resources]
```
stevebachmeier, line 89: Just moved this logic into the new `PopulationManager.add_private_columns` called below

src/vivarium/framework/resource/resource.py
```diff
@@ -37,35 +37,20 @@ def resource_id(self) -> str:
         """The long name of the resource, including the type."""
         return f"{self.resource_type}.{self.name}"
 
-    @property
```
stevebachmeier, line 40: Resources themselves are no longer initialized. Instead, the resource _group_ is instantiated with or w/o an initializer

src/vivarium/framework/time.py
```diff
@@ -116,7 +116,9 @@ def setup(self, builder: Builder) -> None:
             builder.value.register_value_modifier,
             self._simulant_step_size_pipeline,
         )
-        builder.population.initializes_simulants(self)
+        builder.population.register_initializer(
+            initializer=self.on_initialize_simulants, columns=None
```
stevebachmeier, line 120: A rare instance of registering an initializer w/o any private columns

src/vivarium/component.py
```diff
@@ -362,11 +352,8 @@ def setup_component(self, builder: Builder) -> None:
         self._logger = builder.logging.get_logger(self.name)
         self.configuration = self.get_configuration(builder)
         self.setup(builder)
-        self._register_attribute_private_columns(builder)
```
stevebachmeier, line 365: This is important. All three of these Component setup methods have been combined into the new `PopulationManager.register_initializer()` method.

src/vivarium/framework/resource/group.py
```diff
@@ -27,6 +27,7 @@ def __init__(
         self,
         initialized_resources: Iterable[Column] | Resource,
         dependencies: Iterable[str | Resource],
+        initializer: Callable[[SimulantData], None] | None,
```
stevebachmeier, line 30: This is kind of the whole point of this ticket. ResourceGroups are now instantiated with discrete initializers; this allows a single component to register multiple initializers and not have a cycle in the event one initializer was being used to call `PopulationView.update()` twice.

tests/framework/components/test_component.py
```diff
@@ -282,17 +287,6 @@ def test_component_lookup_table_configuration(hdf_file_path: Path) -> None:
     )
     sim.setup()
 
-    # Assertions for specific lookup tables
-    expected_tables = {
```
stevebachmeier, line 286: this just wasn't actually being used

tests/framework/population/test_manager.py
```diff
@@ -73,15 +73,6 @@ def test_initializer_set_fail_attr() -> None:
         component_set.add(NonComponent().initializer, ["test_column"])
 
 
-def test_initializer_set_duplicate_component() -> None:
-    component_set = InitializerComponentSet()
-    component = InitializingComponent("test")
-
-    component_set.add(component.initializer, ["test_column1"])
-    with pytest.raises(PopulationError, match="multiple population initializers"):
-        component_set.add(component.other_initializer, ["test_column2"])
```
stevebachmeier, line 82: we now do allow a single component to add multiple initializers

tests/framework/population/test_manager.py
```diff
@@ -96,13 +87,6 @@ def test_initializer_set_duplicate_columns() -> None:
         component_set.add(component2.initializer, ["sneaky_column"] + columns)
 
 
-def test_initializer_set_population_manager() -> None:
```
stevebachmeier, line 99: i just don't think this is very useful

tests/framework/population/test_manager.py
```diff
@@ -381,30 +366,89 @@ def test_get_population_deduplicates_requested_columns(
     assert set(pop.columns) == {"pie"}
 
 
-def test_register_private_columns() -> None:
-    class ColumnCreator2(Component):
+def test_register_initializer(mocker: MockerFixture) -> None:
```
stevebachmeier, line 369: If you have the stomach for it, take a look at this test and confirm whether you think it's doing a good enough job for this new method

tests/framework/population/test_population_view.py
```diff
@@ -315,7 +316,9 @@ class TestGetAttributesReturnTypes:
     class SomeComponent(ColumnCreator, AttributePipelineCreator):
         """Class that creates multi-level column attributes and private columns."""
 
-        ...
+        def setup(self, builder: Builder) -> None:
```
stevebachmeier, line 319: I've never had a good grasp on multi-inheritance, but evidently when you inherit from more than one class it only calls the setup of one of them. Thanks copilot for this.

tests/framework/resource/test_manager.py
```diff
@@ -32,7 +32,9 @@ def resource_producers() -> dict[int, ResourceProducer]:
 
 @pytest.fixture
 def manager_with_resources(
-    manager: ResourceManager, resource_producers: dict[int, ResourceProducer]
+    manager: ResourceManager,
```
stevebachmeier, line 35: This `manager_with_resources` fixture is very confusing to me. I hope I didn't muck things up while making the tests that rely on it pass

docs/source/tutorials/boids.rst
rmudambi: oh this is much nicer. thanks for this!

src/vivarium/framework/population/population_view.py
```diff
@@ -69,6 +68,15 @@ def __init__(
...
         return f"population_view_{self._id}"
 
+    @property
+    def private_columns(self) -> list[str]:
+        """The private columns managed by this PopulationView."""
+        if self._component is None:
+            raise PopulationError(
+                "This PopulationView is read-only, so it doesn't have access to private_columns."
+            )
+        return self._manager._private_column_metadata[self._component.name]
```
rmudambi: Nit: `_private_column_metadata` shouldn't be a private attribute if we're accessing it like this (or we should have some other public access pattern)

rmudambi (approved):

src/vivarium/framework/population/population_view.py
```diff
@@ -69,6 +68,15 @@ def __init__(
     def name(self) -> str:
         return f"population_view_{self._id}"
 
+    @property
+    def private_columns(self) -> list[str]:
+        """The private columns managed by this PopulationView."""
+        if self._component is None:
```
albrja, line 74: How would component be None? I can see when we instantiate an instance it can be None but what is the use case or situation where this would actually happen?

albrja (approved):

src/vivarium/framework/population/population_view.py
stevebachmeier: The obvious use case is when a Manager needs a population view. In that case, the component arg would be set to None, indicating that the "component" attached to this particular view is not in charge of any private columns. e.g. https://github.com/ihmeuw/vivarium/blob/epic/mic-4656/population-redesign/src/vivarium/framework/results/manager.py#L73

rmudambi: Or a LookupTable

PR #726, opened 2026-01-08
Sbachmei/mic 6727/interactive context get columns method

stevebachmeier:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6727

### Changes and notes
 
The method actually already existed via `list_attributes` (the complement to 
`list_values`). I decided that we probably want users to think of them actually
as columns in the state table so renamed it.

I also slightly streamlined how we're getting it w/ a bit less indirection.
There might still be room for improvement here.

### DIFF
--- a/src/vivarium/interface/interactive.py
+++ b/src/vivarium/interface/interactive.py
@@ -195,13 +195,13 @@ class InteractiveContext(SimulationContext):
             attributes=returned_attributes, squeeze=squeeze
         )
 
+    def get_columns(self) -> list[str]:
+        """List all columns in the population state table."""
+        return self._population.get_all_attribute_names()
+
     def list_values(self) -> list[str]:
         """List the names of all value pipelines in the simulation."""
         return list(self._values.get_value_pipelines().keys())
-
-    def list_attributes(self) -> list[str]:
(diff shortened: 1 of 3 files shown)


src/vivarium/interface/interactive.py
```diff
@@ -195,14 +195,14 @@ def get_population(
...
+        """List all columns in the population state table."""
+        return self._population.get_all_attribute_names()
+
     def list_values(self) -> list[str]:
         """List the names of all value pipelines in the simulation."""
         return list(self._values.get_value_pipelines().keys())
 
-    def list_attributes(self) -> list[str]:
-        """List the names of all attribute pipelines in the simulation."""
-        return list(self._values.get_attribute_pipelines().keys())
```
stevebachmeier, line 204: `ValuesManager.get_attribute_pipelines()` returns a private dictionary `ValuesManager._attribute_pipelines`. That's fine.

But it seems to make more sense to me to instead go through the population manager here (just sort of semantically) and call this new method which returns the keys of the very same dict. It's more indirection but makes more sense I think.?

rmudambi (approved):

albrja (approved):

PR #727, opened 2026-01-12
fix issues after merge conflict resolution

stevebachmeier:
### Description

- *Category*: 
- *JIRA issue*: na

### DIFF
--- a/docs/source/concepts/lookup.rst
+++ b/docs/source/concepts/lookup.rst
@@ -145,10 +145,10 @@ Female  60         100      27
 Constructing Lookup Tables from a Component
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Components can register lookup tables to be built by specifying
-a ``data_sources`` block in their :attr:`~vivarium.component.Component.configuration_defaults` property. 
-As a basic example, DiseaseModel in ``vivarium_public_health`` has the following
-``data_sources`` configuration:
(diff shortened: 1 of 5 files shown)


docs/source/concepts/lookup.rst
```diff
@@ -145,10 +145,10 @@ Female  60         100      27
 Constructing Lookup Tables from a Component
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Components can register lookup tables to be built by specifying
```
stevebachmeier, line 148: I really didn't think terribly hard here so it's probably worth making sure what I said is actually correct.

src/vivarium/framework/values/post_processors.py
```diff
@@ -65,12 +65,24 @@ def rescale_post_processor(
             "datetime timedelta or pandas Timedelta object."
         )
     if isinstance(value, (int, float)):
-        return pd.Series(from_yearly(value, time_step), index=index)
+        result = from_yearly(value, time_step)
+        if not isinstance(result, (int, float)):
```
rmudambi: How would we get the wrong type, and what has changed to necessitate this handling? I assume it's just to make mypy happy.

stevebachmeier: Yeah, mypy. I'm honestly not sure why this was needed. It didn't seem to be any sort of merge conflict but it started failing after the rebase 🤷

hussain-jafari (approved):

rmudambi (approved):

PR #729, opened 2026-01-15
Sbachmei/mic 6743/expose component manager current component

stevebachmeier:
### Description

- *Category*:  other
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6743

### DIFF
--- a/src/vivarium/framework/components/interface.py
+++ b/src/vivarium/framework/components/interface.py
@@ -93,7 +93,7 @@ class ComponentInterface(Interface):
         This method is primarily used internally by the framework to support
         manual setting up of components.
         """
-        self._manager._current_component = component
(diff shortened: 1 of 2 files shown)


rmudambi (commented):
I think rather than creating and exposing these new methods, we should just have the risk component use the private methods, since it's a temporary hack, and we want to find a better way to do this anyway.

stevebachmeier:
> I think rather than creating and exposing these new methods, we should just have the risk component use the private methods, since it's a temporary hack, and we want to find a better way to do this anyway.

@rmudambi like this? Or did you mean truly expose nothing at all? I kinda tried that and it didn't seem to work but I didn't give it much thought.

PR #740, opened 2026-02-11
Albrja/mic 6187/update python versions

albrja:
### Add support for python 3.12 an 3.13
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6187

### Changes and notes
-add support for python 3.12 an 3.13
-update mocker behavior in test suites to use stricter behavior new to 3.13


patricktnast (approved):

tests/framework/artifact/test_artifact.py
```diff
@@ -321,7 +323,6 @@ def test_remove_bad_key(hdf_mock: MagicMock, artifact_path: Path) -> None:
     assert key not in a
     assert key not in a._cache
     hdf_mock.remove.assert_not_called()
-    hdf_mock.write.called_once_with(artifact_path, "metadata.keyspace", ["metadata.keyspace"])
```
stevebachmeier, line 324: why lose this?

tests/framework/artifact/test_artifact.py
```diff
@@ -445,7 +446,6 @@ def test_replace_nonexistent_key(hdf_mock: MagicMock, artifact_path: Path) -> No
     key = "new.key"
 
     a = Artifact(artifact_path, filter_terms=filter_terms)
-    hdf_mock.called_once_with(key)
```
stevebachmeier, line 448: why lose this?

tests/framework/artifact/test_artifact.py
```diff
@@ -102,23 +102,23 @@ def test_artifact_creation(
     assert a.filter_terms is None
     assert a._cache == {}
     assert a.keys == keys_mock
-    hdf_mock.load.called_once_with("metadata.keyspace")
```
stevebachmeier, line 105: what about incrememting support py versions demanded changes in this file?

tests/framework/test_configuration.py
```diff
@@ -23,7 +23,7 @@ def test_get_default_specification_user_config(
 
     default_spec = _get_default_specification()
 
-    assert expand_user_mock.called_once_with("~/vivarium.yaml")
```
stevebachmeier, line 26: these changes don't seem correct. Are we not using `~/vivarium.yaml` as a default anymore? And how/why are these changes showing up in a PR about increasing python version support?

.readthedocs.yml
```diff
@@ -11,7 +11,7 @@ sphinx:
 build:
   os: ubuntu-22.04
   tools:
-    python: "3.11"
+    python: "3.12"
```
stevebachmeier: Should this be 3.13 (the latest version we support)?

.github/workflows/deploy.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: "3.11"
+          python-version: "3.12"
```
stevebachmeier: 3.13?

stevebachmeier (changes requested):

tests/framework/artifact/test_artifact.py
albrja: In this case it is because we never write a key because we are checking for a non-existent key. According to copilot, this is part of the stricter usage of mocks with 3.13 but I have updated to match the call above it.

tests/framework/artifact/test_artifact.py
albrja: Stricter mock behavior and assertion. Those are the other arguments to load.

tests/framework/artifact/test_artifact.py
albrja: This mock was being called incorrectly but it has been updated.

tests/framework/test_configuration.py
albrja: expand_user_mock = mocker.patch("vivarium.framework.configuration.Path.expanduser") is a mocked Path object. This does not take an argument which is being specified here. And yes otherwise the rest of the test would fail.

albrja:
@stevebachmeier Here is a brief summary of mock updates from Copilot for python 3.13 (obviously not comprehensive):
"These errors are happening because Python 3.13 made changes to the unittest.mock library. In Python 3.13, using assertion attributes like called_once_with directly on a Mock object (e.g., mock.called_once_with) is no longer valid unless the mock was created with a spec that includes those attributes.

Previously, you could sometimes use mock.called_once_with as an assertion, but this was never the intended usage. The correct way is to call the method as an assertion, like mock.assert_called_once_with(...). In Python 3.13, trying to access mock.called_once_with as an attribute raises an AttributeError unless the mock is given a spec that defines it."

stevebachmeier (approved):

CHANGELOG.rst
```diff
@@ -47,6 +47,7 @@ Major changes
 
 Other changes
 -------------
+- Add support for python 3.12 and 3.13
```
stevebachmeier: Should either go under the Major Changes above or below under Miscellaneous. The `------` headers are level 1 and `~~~~~~~~` are level 2.

stevebachmeier: But we should discuss whether updating python version support is a minor or patch release (i.e. loosely "major" or "miscellaneous" updates in this changelog)

hussain-jafari (approved):

PR #744, opened 2026-02-12
Disallow tracked filtering on initialization or pop creation

stevebachmeier:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6801

### Changes and notes
 
We ran into issues with initializers requesting attributes from pop view
which by default tries to filter untracked simulants out but the column
not existing yet. We decided that on initialization (and, by extension, 
population generation) we should just never try to filter out untracked
simulants; users can post-process as desired if needed.

Note that this does not cause a problem in later lifecycle phases since
by definition all initializers would have set up all columns by then.

### DIFF
--- a/src/vivarium/framework/population/manager.py
+++ b/src/vivarium/framework/population/manager.py
@@ -176,6 +176,7 @@ class PopulationManager(Manager):
         self._get_current_component_or_manager = (
             builder.components.get_current_component_or_manager
         )
+        self._get_current_state = builder.lifecycle.current_state()
 
         builder.lifecycle.add_constraint(
             self.get_view,
--- a/src/vivarium/framework/population/manager.py
+++ b/src/vivarium/framework/population/manager.py
@@ -176,7 +176,7 @@ class PopulationManager(Manager):
         self._get_current_component_or_manager = (
             builder.components.get_current_component_or_manager
         )
-        self._get_current_state = builder.lifecycle.current_state()
+        self.get_current_state = builder.lifecycle.current_state()
 
         builder.lifecycle.add_constraint(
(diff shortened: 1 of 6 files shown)


src/vivarium/framework/population/manager.py
```diff
@@ -176,6 +176,7 @@ def setup(self, builder: Builder) -> None:
         self._get_current_component_or_manager = (
             builder.components.get_current_component_or_manager
         )
+        self._get_current_state = builder.lifecycle.current_state()
```
rmudambi: This shouldn't be private if it's being called by the population view?

src/vivarium/framework/time/__init__.py
```diff
@@ -1,2 +1,7 @@
 from vivarium.framework.time.interface import TimeInterface
-from vivarium.framework.time.manager import DateTimeClock, SimpleClock, SimulationClock
+from vivarium.framework.time.manager import (
+    DateTimeClock,
+    SimpleClock,
+    SimulationClock,
+    get_time_stamp,
```
rmudambi, line 6: Why do we have this change?

rmudambi (approved):

patricktnast (approved):

src/vivarium/framework/time/__init__.py
stevebachmeier: I split the time interface and manager into separate modules so this had to be imported differently

stevebachmeier: (and I just didn't notice until running the pregnancy model which actually imports this function)

PR #745, opened 2026-02-13
squash lookup table types

rmudambi:
### Description

- *Category*: refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6817

### Changes and notes
- Replaced ScalarTable, CategoricalTable, and InterpolatedTable with a single LookupTable class
- Enable Interpolation call to function when there are no continuous parameters

### Testing
Ran make check


src/vivarium/framework/lookup/table.py
```diff
@@ -93,18 +133,51 @@ def __call__(self, index: pd.Index[int]) -> T:
...
+            # Interpolate continuous parameters and categorize categorical parameters based on
+            # the population attributes.
+            requested_columns = [
+                col
+                for col in list(self.key_columns) + list(self.parameter_columns)
+                if col != "year"
+            ]
+            pop = pd.DataFrame(self.population_view.get_attributes(index, requested_columns))
+            if "year" in self.parameter_columns:
+                current_time = self._manager.clock()
```
albrja, line 168: Does this handle if a simulation has run to for example 2030, but our data is only for 2023?

albrja (approved):

src/vivarium/framework/lookup/table.py
```diff
@@ -127,249 +200,64 @@ def get_name(component_name: str, table_name: str) -> str:
...
-            self.population_view.get_attributes(
-                index, list(self.key_columns) + list(self.parameter_columns)
-            )
-        )
-
-        # specify some numeric type for columns, so they won't be objects but
-        # will be updated with whatever column type it actually is
-        result = pd.DataFrame(index=pop.index, columns=self.value_columns, dtype=np.float64)
-
-        sub_tables = pop.groupby(list(self.key_columns))
```
stevebachmeier, line 305: I don't see an obvious/direct replacement for this CategoricalTable logic

stevebachmeier (commented):
are categorical tables still handled?

patricktnast (approved):

hussain-jafari (approved):

src/vivarium/framework/lookup/table.py
rmudambi: Yes, that's what the configuration extrapolate == True does.

src/vivarium/framework/lookup/table.py
rmudambi: InterpolatedTables already had to do everything that CategoricalTables did. They just had a constraint that there had to be at least one continuous column. I've removed that constraint with the small change in interpolation.py.

stevebachmeier (approved):

PR #746, opened 2026-02-13
Remove support for population view default queries

stevebachmeier:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-XYZ

### DIFF
--- a/docs/source/concepts/population.rst
+++ b/docs/source/concepts/population.rst
@@ -81,8 +81,12 @@ Filtering Simulants
 There are two types of filtering that can be applied when using a population view
 to get attributes or private columns.
 
-First, a ``query`` argument can be passed in to any population view ``get_`` data
-method to filter the simulants returned for that specific call.
+First, a ``query`` argument can be passed in to any of the population view's 
+:meth:`~vivarium.framework.population.population_view.PopulationView.get_attributes`, 
+:meth:`~vivarium.framework.population.population_view.PopulationView.get_attribute_frame`, 
+:meth:`~vivarium.framework.population.population_view.PopulationView.get_private_columns`, or
+:meth:`~vivarium.framework.population.population_view.PopulationView.get_filtered_index`
+to filter the simulants returned for that specific call.
 
 Second, if any components have registered an untracking query, untracked simulants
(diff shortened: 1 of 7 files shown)


docs/source/concepts/population.rst
```diff
@@ -78,20 +78,13 @@ state.
...
-There are three types of filtering that can be applied when using a population view
+There are two types of filtering that can be applied when using a population view
 to get attributes or private columns.
 
-First, after a population view has been created on a component, that component can
-call :meth:`vivarium.framework.population.population_view.PopulationView.set_default_query`
-to set a query that will be used by default every time that view is used to access
-data. There is an optional ``include_default_query`` argument that defaults to True
-that can be used to bypass default query filtering if desired.
+First, a ``query`` argument can be passed in to any population view ``get_`` data
```
rmudambi: I find `get_` a bit confusing. I think you should just list the methods explicitly - we don't have that many.

src/vivarium/framework/population/population_view.py
```diff
@@ -657,30 +615,26 @@ def _update_column_and_ensure_dtype(
         )
         return new_data
 
-    def _build_query(
-        self, query: str, include_default_query: bool, include_untracked: bool
-    ) -> str:
+    def _build_query(self, query: str, include_untracked: bool) -> str:
```
rmudambi, line 609: This is not really part of the change that you are making here, but I always felt like the `_skip_tracked_query_if_initializing()` method had a bit of a code smell. We can improve this imo if you move this logic inside `self._build_query()` since they are always used in conjunction:

```python
def _build_query(self, query, include_untracked):
    include_untracked = (
        include_untracked or self._manager.get_current_state() in [
            lifecycle_states.INITIALIZATION, lifecycle_states.POPULATION_CREATION
        ]
    )
    return pop_utils.combine_queries(...)
```

rmudambi (approved):

albrja (approved):

src/vivarium/framework/population/population_view.py
stevebachmeier: Good call

PR #747, opened 2026-02-19
Remove InitializerComponentSet

stevebachmeier:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5453

### Changes and notes
 
The InitializerComponentSet did some basic validations, but having an entire 
extra class seems unnecessary so I moved it into PopulationManager.

### DIFF
--- a/src/vivarium/component.py
+++ b/src/vivarium/component.py
@@ -22,7 +22,6 @@ from layered_config_tree import ConfigurationError, LayeredConfigTree
 
 from vivarium.framework.artifact import ArtifactException
 from vivarium.framework.lifecycle import LifeCycleError, lifecycle_states
-from vivarium.framework.population import PopulationError
 from vivarium.types import LookupTableData
 
 if TYPE_CHECKING:
@@ -195,6 +194,8 @@ class Component(ABC):
             If the component does not have access to the state table.
         """
         if self._population_view is None:
+            from vivarium.framework.population.exceptions import PopulationError
+
             raise PopulationError(
                 f"Component '{self.name}' does not have access to the state table. "
                 "This is likely due to having called this prior to simulation setup."
--- a/src/vivarium/framework/population/manager.py
+++ b/src/vivarium/framework/population/manager.py
@@ -8,7 +8,6 @@ from __future__ import annotations
 
 from collections.abc import Callable, Sequence
 from dataclasses import dataclass
-from types import MethodType
 from typing import TYPE_CHECKING, Any, Literal, overload
 
 import pandas as pd
@@ -419,22 +418,12 @@ class PopulationManager(Manager):
         """
 
         component = self._get_current_component_or_manager()
-        if not isinstance(initializer, MethodType):
-            raise TypeError(
-                "Population initializers must be methods of vivarium Components or managers. "
-                f"You provided {initializer} which is of type {type(initializer)}."
-            )
+
         if not (isinstance(component, Component) or isinstance(component, Manager)):
             raise AttributeError(
                 "Population initializers must be methods of vivarium Components or Managers. "
                 f"You provided {initializer} which is bound to {component} that "
                 f"is of type {type(component)}."
-            )
-        if not hasattr(component, "name"):
-            raise AttributeError(
-                "Population initializers must be methods of named simulation components. "
-                f"You provided {initializer} which is bound to {component} that has no "
-                f"'name' attribute."
             )
 
         if columns is None:
--- a/src/vivarium/framework/population/manager.py
+++ b/src/vivarium/framework/population/manager.py
@@ -408,23 +408,11 @@ class PopulationManager(Manager):
 
         Raises
         ------
-        TypeError
-            If the ``initializer`` is not a method type.
-        AttributeError
-            If the ``initializer`` is not bound to a Component or the PopulationManager
-            or if the bound component does not have a name attribute.
         PopulationError
             If this component name has already registered private columns.
         """
 
         component = self._get_current_component_or_manager()
-
-        if not (isinstance(component, Component) or isinstance(component, Manager)):
-            raise AttributeError(
-                "Population initializers must be methods of vivarium Components or Managers. "
-                f"You provided {initializer} which is bound to {component} that "
-                f"is of type {type(component)}."
-            )
 
         if columns is None:
             columns = []
--- a/src/vivarium/framework/population/manager.py
+++ b/src/vivarium/framework/population/manager.py
@@ -6,7 +6,7 @@ Population Manager
 """
 from __future__ import annotations
 
-from collections.abc import Callable, Iterable, Sequence
+from collections.abc import Callable, Sequence
 from dataclasses import dataclass
 from types import MethodType
 from typing import TYPE_CHECKING, Any, Literal, overload
@@ -14,6 +14,7 @@ from typing import TYPE_CHECKING, Any, Literal, overload
 import pandas as pd
 
 import vivarium.framework.population.utilities as pop_utils
+from vivarium.component import Component
 from vivarium.framework.event import Event
 from vivarium.framework.lifecycle import lifecycle_states
 from vivarium.framework.population.exceptions import PopulationError
@@ -22,7 +23,6 @@ from vivarium.framework.resource import Resource
 from vivarium.manager import Manager
 
 if TYPE_CHECKING:
-    from vivarium import Component
     from vivarium.framework.engine import Builder
     from vivarium.types import ClockStepSize, ClockTime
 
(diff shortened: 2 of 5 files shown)


src/vivarium/framework/population/manager.py
```diff
@@ -49,76 +49,6 @@ class SimulantData:
...
-        TypeError
-            If the initializer is not an object method.
-        AttributeError
-            If the object bound to the method does not have a name attribute.
-        PopulationError
-            If the component bound to the method already has an initializer
-            registered or if the columns produced are duplicates of columns
-            another initializer produces.
-        """
-        if not isinstance(initializer, MethodType):
```
stevebachmeier, line 82: This check is moved into PopulationManager

src/vivarium/framework/population/manager.py
```diff
@@ -49,76 +49,6 @@ class SimulantData:
...
-        # TODO: raise error once all active Component implementations have been refactored
-        # if not (isinstance(component, Component) or isinstance(component, PopulationManager)):
-        #     raise AttributeError(
-        #         "Population initializers must be methods of vivarium Components "
-        #         "or the simulation's PopulationManager. "
-        #         f"You provided {initializer} which is bound to {component} that "
-        #         f"is of type {type(component)} which does not inherit from "
-        #         "Component."
-        #     )
-        if not hasattr(component, "name"):
```
stevebachmeier, line 98: This check is moved into PopulationManager

src/vivarium/framework/population/manager.py
```diff
@@ -49,76 +49,6 @@ class SimulantData:
...
-            raise AttributeError(
-                "Population initializers must be methods of named simulation components. "
-                f"You provided {initializer} which is bound to {component} that has no "
-                f"'name' attribute."
-            )
-
-        component_name = component.name
-        for column in columns_produced:
-            if column in self._columns_produced:
-                raise PopulationError(
```
stevebachmeier, line 108: This is already being done in the PopulationManager

src/vivarium/framework/population/manager.py
```diff
@@ -49,76 +49,6 @@ class SimulantData:
...
-                    f"Component {component_name} and component {self._columns_produced[column]} "
-                    f"have both registered initializers for column {column}."
-                )
-            self._columns_produced[column] = component_name
-        self._components[component_name] = list(columns_produced)
```
stevebachmeier, line 113: This is just wrong since a single component can now register multiple initializers. The correct approach is already being done in the PopulationManager via the `private_column_metadata` default dict which _extends_ columns rather than overwriting.

src/vivarium/framework/population/manager.py
```diff
@@ -480,11 +409,34 @@ def register_initializer(
...
             If this component name has already registered private columns.
         """
 
         component = self._get_current_component_or_manager()
+        if not isinstance(initializer, MethodType):
+            raise TypeError(
+                "Population initializers must be methods of vivarium Components or managers. "
+                f"You provided {initializer} which is of type {type(initializer)}."
+            )
+        if not (isinstance(component, Component) or isinstance(component, Manager)):
```
stevebachmeier: The (commented out) code for this check in the InitializerComponentSet actually wanted to assert that component was of type Component or specifically PopulationManager. The problem there is that _any_ manager can technically register an initializer, e.g. the DateTimeClock manager.

tests/framework/population/test_manager.py
```diff
@@ -25,29 +25,7 @@
...
 from vivarium.framework.engine import Builder
 from vivarium.framework.population.exceptions import PopulationError
-from vivarium.framework.population.manager import (
-    InitializerComponentSet,
-    PopulationManager,
-    SimulantData,
-)
-
-
-def test_initializer_set_fail_type() -> None:
```
stevebachmeier, line 35: New spiritual successor below (`test_register_initializer_not_a_method_raises`)

tests/framework/population/test_manager.py
```diff
@@ -66,35 +44,6 @@ def other_initializer(self, simulant_data: SimulantData) -> None:
         pass
 
 
-def test_initializer_set_fail_attr() -> None:
```
stevebachmeier, line 69: New spiritual successor below (`test_register_initializer_unnamed_component_raises`)

tests/framework/population/test_manager.py
```diff
@@ -66,35 +44,6 @@ def other_initializer(self, simulant_data: SimulantData) -> None:
...
-    component_set = InitializerComponentSet()
-
-    with pytest.raises(AttributeError):
-        component_set.add(NonComponent().initializer, ["test_column"])
-
-
-def test_initializer_set_duplicate_columns() -> None:
```
stevebachmeier, line 76: Deleted b/c we have this test elsewhere

tests/framework/population/test_manager.py
```diff
@@ -66,35 +44,6 @@ def other_initializer(self, simulant_data: SimulantData) -> None:
...
-
-    component_set.add(component1.initializer, columns)
-    with pytest.raises(PopulationError, match="both registered initializers"):
-        component_set.add(component2.initializer, columns)
-
-    with pytest.raises(PopulationError, match="both registered initializers"):
-        component_set.add(component2.initializer, ["sneaky_column"] + columns)
-
-
-def test_initializer_set() -> None:
```
stevebachmeier, line 90: Deleted b/c it's now irrelevant

src/vivarium/framework/population/manager.py
```diff
@@ -480,11 +409,34 @@ def register_initializer(
...
+            If the ``initializer`` is not a method type.
+        AttributeError
+            If the ``initializer`` is not bound to a Component or the PopulationManager
+            or if the bound component does not have a name attribute.
         PopulationError
             If this component name has already registered private columns.
         """
 
         component = self._get_current_component_or_manager()
+        if not isinstance(initializer, MethodType):
```
rmudambi: I know we were previously doing this validation, but is it actually necessary? What would go wrong if the initializer of a component was function, a class method, or even a method on some other component or class?

If we do need this validation, we should also validate that `initializer.__self__ == self._get_current_component_or_manager()`.

src/vivarium/framework/population/manager.py
rmudambi: `component` as currently defined is guaranteed to be a Component or a Manager. You really want to do this
```suggestion
        if component != initializer.__self__:
```

src/vivarium/framework/population/manager.py
```diff
@@ -480,11 +409,34 @@ def register_initializer(
...
+                "Population initializers must be methods of vivarium Components or managers. "
+                f"You provided {initializer} which is of type {type(initializer)}."
+            )
+        if not (isinstance(component, Component) or isinstance(component, Manager)):
+            raise AttributeError(
+                "Population initializers must be methods of vivarium Components or Managers. "
+                f"You provided {initializer} which is bound to {component} that "
+                f"is of type {type(component)}."
+            )
+        if not hasattr(component, "name"):
```
rmudambi: This check is no longer needed. All Components and Managers are guaranteed to have a `name` attribute.

src/vivarium/framework/resource/interface.py
```diff
@@ -75,7 +75,7 @@ def add_resources(
 
     def add_private_columns(
         self,
-        initializer: Callable[[SimulantData], None] | None,
+        initializer: Callable[[SimulantData], None],
```
rmudambi, line 78: Why was None permitted before? Was that a mistake?

src/vivarium/component.py
```diff
@@ -195,6 +194,8 @@ def population_view(self) -> PopulationView:
             If the component does not have access to the state table.
         """
         if self._population_view is None:
+            from vivarium.framework.population.exceptions import PopulationError
```
rmudambi, line 197: Did this cause a circular import? That's rather surprising, since the `exceptions` module shouldn't be importing much of anything.

src/vivarium/framework/population/manager.py
stevebachmeier: I thought about this as well and agree it doesn't seem necessary.

src/vivarium/framework/population/manager.py
stevebachmeier: We should talk about this b/c (1) the line you recommend deleting _is_ checking that component is a Component or Manager whereas the line you replace it with does not, and (2) to the point you made in the other comment what if we wanted to initialize a function instead of a method?

src/vivarium/framework/resource/interface.py
stevebachmeier: Yeah, just missed it

src/vivarium/component.py
stevebachmeier: Yeah, it did. It's initializing the population package which imports manager.py which imports Component.

```
ImportError while loading conftest '/Users/sbachmei/repos/vivarium/tests/conftest.py'.
tests/conftest.py:15: in <module>
    from vivarium.framework.configuration import (
src/vivarium/__init__.py:15: in <module>
    from vivarium.component import Component
src/vivarium/component.py:19: in <module>
    from vivarium.framework.population.exceptions import PopulationError
src/vivarium/framework/population/__init__.py:16: in <module>
    from vivarium.framework.population.manager import PopulationManager, SimulantData
src/vivarium/framework/population/manager.py:16: in <module>
    from vivarium.component import Component
E   ImportError: cannot import name 'Component' from partially initialized module 'vivarium.component' (most likely due to a circular import) (/Users/sbachmei/repos/vivarium/src/vivarium/component.py)
```

src/vivarium/framework/population/manager.py
rmudambi: This entire check is moot if we no longer check that the initializer is a `MethodType`. 

FWIW, you don't need to check that `component` is a Component or a Manager, because `self._get_current_component_or_manager()` is type-hinted to return an object of those types.

stevebachmeier: Oh, good point. Ok, so basically just literally delete all of the InitializerComponentSet class since we really are already doing everything it was doing - just in different ways

albrja (approved):

rmudambi (approved):

PR #748, opened 2026-02-19
Sbachmei/raise if registering duplicate initializer

stevebachmeier:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-XYZ


src/vivarium/framework/time/interface.py
```diff
@@ -67,10 +67,8 @@ def register_step_size_modifier(
             Modifier of the step size pipeline. Modifiers can take an index
             and should return a series of step sizes.
         required_resources
-            A list of resources that need to be properly sourced before the
-            pipeline source is called. This is a list of strings, pipelines,
-            or randomness streams.
-
+            A list of resources that the producer requires. A string represents
```
stevebachmeier, line 70: Took the opportunity to improve these docstrings since it was previously unclear what a string represented.

src/vivarium/framework/population/manager.py
```diff
@@ -436,6 +438,11 @@ def register_initializer(
                 f"You provided {initializer} which is bound to {component} that has no "
                 f"'name' attribute."
             )
+        if initializer in self._registered_initializers:
```
hussain-jafari, line 417: Looks good as long as this is the right way of checking - callables can be tricky.

hussain-jafari (approved):

src/vivarium/framework/population/manager.py
```diff
@@ -436,6 +438,11 @@ def register_initializer(
...
+            raise PopulationError(
+                f"The initializer '{initializer.__qualname__}' has already been registered. "
```
rmudambi, line 419: What is `__qualname__`?

rmudambi (approved):

src/vivarium/framework/population/manager.py
stevebachmeier: I think so? 

```
>>> from vivarium import Component
>>> class ComponentA(Component):
...     def on_initialize():
...             pass
... 
>>> class ComponentB(Component):
...     def on_initialize():
...             pass
... 
>>> componont_a = ComponentA()
>>> component_b = ComponentB()
>>>
>>> registered_initializers = [componont_a.on_initialize]
>>>
>>> componont_a.on_initialize in registered_initializers
True
>>>
>>> component_b.on_initialize in registered_initializers
False
```

src/vivarium/framework/population/manager.py
stevebachmeier: It's the "qualified" name that shows the dotpath to the object. In this case it's a nice way to see which component the method is attached to (that's slightly cleaner than the \_\_repr\_\_)
```
(Pdb) initializer.__qualname__
'DeathsObserver.on_initialize_simulants'
(Pdb) initializer.__name__
'on_initialize_simulants'
```

albrja (approved):

PR #753, opened 2026-02-25
use pop view for InteractiveContext.get_population()

stevebachmeier:
### Description

- *Category*: 
- *JIRA issue*: na

### DIFF
--- a/src/vivarium/interface/interactive.py
+++ b/src/vivarium/interface/interactive.py
@@ -188,26 +188,23 @@ class InteractiveContext(SimulationContext):
         -------
             The current state of requested population attributes.
         """
-        returned_attributes: str | list[str] | tuple[str, ...]
         if attributes is None:
-            returned_attributes = self.get_attribute_names()
-            if len(returned_attributes) == 1:
-                returned_attributes = returned_attributes[0]
-        else:
-            returned_attributes = attributes
+            attributes = self.get_attribute_names()
+            if len(attributes) == 1:
+                attributes = attributes[0]
         try:
             return self._population_view.get_attributes(
-                index=self.get_population_index(), attributes=returned_attributes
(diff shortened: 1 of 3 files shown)


albrja (approved):

src/vivarium/interface/interactive.py
```diff
@@ -187,16 +188,25 @@ def get_population(
...
+        if attributes is None:
+            attributes = self.get_attribute_names()
+            if len(attributes) == 1:
+                attributes = attributes[0]
+        try:
+            return self._population_view.get_attributes(
+                index=self.get_population_index(), attributes=attributes
+            )
+        except ValueError as e:
+            if "call `get_attribute_frame()` instead" in str(e):
```
rmudambi: Can't we figure out which we should call rather than trying one and then calling the other if it fails?

rmudambi (commented):
We should have a test that covers this failure case?

src/vivarium/interface/interactive.py
stevebachmeier: I can't. It would require the user to know the attribute beforehand.

stevebachmeier: Alternatively, I think we could never pass in a single value to get_attributes() and just never squeeze.

rmudambi: Nit: Is something like this cleaner? Then I think you wouldn't need to raise the RuntimeError.
```python
if attributes is None:
    ...
if isinstance(attributes, str):
    try:
        return self._population_view.get_attributes(...)
    except ValueError as e:
        if "call `get_attribute_frame()` instead" in str(e):
            return self._population_view.get_attribute_frame(...)
        raise
return self._population.get_attributes(...)
```

rmudambi (approved):

PR #754, opened 2026-02-26
add required resources to all resource types and eliminate ResourceGroup

rmudambi:
### Description

- *Category*: refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6839

### Changes and notes
- Add required resources to all Resource types
- Defer evaluation of required resources until the moment we create the resource graph
- Add lifecycle constraints on `ResourceManager.get_graph()` so that we can't create it before the resource requirements have been finalized
- Introduce new Resource called Initializer - this eliminates the need for ResourceGroups
- Build the resource dependency graph with Resource objects rather than ResourceGroups
- Renamed `add_resources()` to `add_resource()`, since we always add resources one at a time.
- Simplify `add_resource()` so that the only argument is the Resource itself
- Create `resource_type` property for Resources to simplify getting the right resource name when evaluating an `AttributePipeline`
- Remove the need to add AttributePipelines to the ResourceManager on post-setup. Now we can do it when it is created.
- Standardized type-hinting of required resources to be an `Iterable`. Previously it was a mix of `Iterable` and `Sequence`
- Repurposed ValueModifer's `get_modifier_name()` to get names of initializers as well.

### Testing
All existing tests pass


src/vivarium/framework/resource/resource.py
```diff
@@ -1,27 +1,38 @@
...
-        """The type of the resource."""
         self.name = name
         """The name of the resource."""
         self._component = component
         """The component that creates the resource. Can be None if not yet set."""
+        self._required_resources: list[str | Resource] = list(required_resources)
+        """The resources required to produce this resource."""
+
+    @property
+    def resource_type(self) -> str:
```
albrja, line 33: This seems like it is an abstract method.

albrja (approved):

src/vivarium/framework/resource/manager.py
```diff
@@ -79,56 +73,44 @@ def setup(self, builder: Builder) -> None:
...
         )
-        builder.event.register_listener(lifecycle_states.POST_SETUP, self.on_post_setup)
         builder.lifecycle.add_constraint(
-            self.add_resources,
+            self.add_resource,
             allow_during=[lifecycle_states.SETUP, lifecycle_states.POST_SETUP],
         )
+        builder.lifecycle.add_constraint(
+            self.add_private_columns,
+            allow_during=[lifecycle_states.SETUP, lifecycle_states.POST_SETUP],
```
stevebachmeier: should we only limit to on setup?

src/vivarium/framework/resource/resource.py
```diff
@@ -37,23 +48,84 @@ def resource_id(self) -> str:
...
+
+
+class Initializer(Resource):
+    """A resource representing a method for initializing simulant state."""
 
-class NullResource(Resource):
-    """A node in the dependency graph that does not produce any resources."""
+    def __init__(
+        self,
+        index: int,
```
stevebachmeier: nit: maybe initializer_index? Just b/c my brain short-circuited thinking it was the population index

stevebachmeier (approved):

patricktnast (approved):

PR #756, opened 2026-02-26
Sbachmei/mic 6812/handle nested pipeline queries

stevebachmeier:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6812


src/vivarium/framework/population/population_view.py
```diff
@@ -189,7 +193,7 @@ def get_attribute_frame(
         index: pd.Index[int],
         attribute: str,
         query: str = "",
-        include_untracked: bool = False,
+        include_untracked: bool | None = None,
```
albrja, line 195: This seems weird that a argument that should be boolean can be None.

src/vivarium/framework/population/population_view.py
```diff
@@ -606,22 +623,42 @@ def _update_column_and_ensure_dtype(
...
-        We explicitly set 'include_untracked' to True during initialization and
-        population creation lifecycle phases.
+        Only the tracked query is affected. Any explicit ``query`` argument is
+        always preserved so that pipeline sources can further subdivide the index.
         """
-        include_untracked = include_untracked or self._manager.get_current_state() in [
-            lifecycle_states.INITIALIZATION,
-            lifecycle_states.POPULATION_CREATION,
-        ]
+        skip_tracked_query = include_untracked is True or (
```
albrja, line 645: I think I am having trouble following the mental map of this but if you are trying to solve whether or not to include tracked simulants in a nested pipeline, wouldn't you still have the argument from the top level? Or another way to ask this is if I want to include untracked simulants in my query (from my original call at the top level) then wouldn't I always want that if it was a nested pipeline? I think I am missing something that made adding None to this necessary.

src/vivarium/framework/population/population_view.py
stevebachmeier: None just represents that they didn't pass anythingn in. That gets handled differently later

stevebachmeier: So basically if they don't pass anything in (None), then we handle the tracking query the way we think they want (i.e. exclude untracked simulants during outer call unless on initialization but then do not append the tracking filter on subsequent nested calls (which is actually a no-op since they've already been removed from the outer call).

But if the user wants to do something out of the ordinary, they are free to explicitly pass in True or False.

It boils down to the default behavior used to be passing in False, but we wanted a slightly different default behavior which is False in outer calls and True in nested calls.

src/vivarium/framework/population/manager.py
rmudambi, line 616: Unrelated to this PR, but it is a bug:
If we skip the post-processor, we will not return requested query columns.

src/vivarium/framework/engine.py
```diff
@@ -224,14 +224,6 @@ def __init__(
         self._lifecycle.add_constraint(
             self.add_components, allow_during=[lifecycle_states.INITIALIZATION]
         )
-        self._lifecycle.add_constraint(
```
rmudambi, line 227: Why did you remove this constraint? The population isn't created until `lifecycle_states.POPULATION_CREATION` which is after `POST_SETUP`, so it doesn't make any sense to call it before then.

src/vivarium/framework/population/population_view.py
stevebachmeier: You're basically right. Even if we didn't change it if include_untracked is None, then it would add the filter but since the population is already reduced from the outer call it would just be a no-op. So this just removes redundant/usesull querying.

What we've allowed for, though, is if for some weird reason the user really thinks they _always_ want True or _always_ want False, they can now do it.

src/vivarium/interface/interactive.py
```diff
@@ -183,6 +196,14 @@ def get_population(
         attributes
             The attribute pipelines to include in the returned table. If None, all
             attributes are included.
+        query
+            Additional conditions used to filter the index.
+        include_untracked
+            Whether to include untracked simulants. If None (default), untracked
+            simulants are excluded from the first call unless during initialization
+            or population-creation and the tracked query is automatically suppressed
+            inside subsequent nested pipeline evaluations. Untracked simulants are
```
rmudambi: I don't think this docstring needs to talk about initialization or nested pipelines. When calling this from the interactive context, I think it will always behave exactly as they ask it to.

If you adopt my suggested docstring change, I think it becomes a bit clearer that this is true.

src/vivarium/framework/population/population_view.py
```diff
@@ -143,7 +143,11 @@ def get_attributes(
         query
             Additional conditions used to filter the index.
         include_untracked
-            Whether to include untracked simulants.
+            Whether to include untracked simulants. If None (default), untracked
+            simulants are excluded from the first call unless during initialization
+            or population-creation and the tracked query is automatically suppressed
+            inside subsequent nested pipeline evaluations. Untracked simulants are
+            always included if True and always excluded if False.
```
rmudambi:
```suggestion
            Whether to include untracked simulants. If None (default), untracked
            simulants are excluded unless this pipeline was called during initialization
            or population-creation or inside another pipeline call. Untracked simulants are
            always included if True and always excluded if False.
```

tests/framework/population/test_population_view.py
```diff
@@ -864,21 +864,87 @@ def test__update_column_and_ensure_dtype_unmatched_dtype() -> None:
 ##################################################
 
 
-def test__skip_tracked_query_if_initializing(
+@pytest.mark.parametrize("include_untracked", [None, False, True])
+@pytest.mark.parametrize(
+    "lifecycle_state",
+    [state.lower() for state in dir(lifecycle_states) if not state.startswith("_")],
```
rmudambi: As I said in another comment, we shouldn't allow these calls before `POPULATION_CREATION`. Also, you probably don't need to exhaustively test all lifecycle phases. I'd suggest just `POPULATION_CREATION` and `TIME_STEP` would be enough.

tests/framework/population/test_population_view.py
```diff
@@ -864,21 +864,87 @@ def test__update_column_and_ensure_dtype_unmatched_dtype() -> None:
...
+    assert pies_and_cubes_pop_mgr.pipeline_evaluation_depth == 0
+    assert (
+        pv._build_query(query, include_untracked=None) == f"({query}) and ({tracking_query})"
+    )
+    assert (
+        pv._build_query(query, include_untracked=False) == f"({query}) and ({tracking_query})"
+    )
+    assert pv._build_query(query, include_untracked=True) == f"({query})"
+
+    for depth in range(1, 6):
```
rmudambi: Do we need to go this deep? Wouldn't just going one (or at most two) levels deep do enough to test this?

tests/interface/test_interactive.py
```diff
@@ -117,15 +149,276 @@ def test_get_population_squeezing() -> None:
...
+    "outer" attribute pipieline whose source method calls another attribute
+    "foo". Note that "outer" is by definition never a simple pipeline
+    because its source is not a list of columns.
+    """
+
+    #########
+    # Tests #
+    #########
+
+    @pytest.mark.parametrize("is_simple_inner_attribute", [True, False])
```
rmudambi: Do we need the simple case for these tests?

tests/interface/test_interactive.py
```diff
@@ -117,15 +149,276 @@ def test_get_population_squeezing() -> None:
...
+            kwargs["include_untracked"] = include_untracked
+
+        # Baseline: no tracked query yet
+        pop = sim.get_population(**kwargs)  # type: ignore[call-overload]
+        assert set(pop["inner"]) == {0, 1, 2}
+        if outer_column is not None:
+            assert set(pop[outer_column]) == expected_baseline_outer
+
+        # Register tracked query and patch depth
+        self._register_tracked_query(sim, tracked_query, mocker)
```
rmudambi: Shouldn't this happen in `_create_sim()`?

src/vivarium/framework/population/manager.py
stevebachmeier: Good catch. MIC-6855

src/vivarium/framework/engine.py
stevebachmeier: Because I actually removed SimulationContext.get_population() below! This is in favor of the reworked InteractiveContext.get_population() method which I don't think requires constraining.

I guess I should have done this in a separate PR but everything was getting kinda convoluted

src/vivarium/framework/engine.py
```diff
@@ -373,12 +366,6 @@ def add_components(self, component_list: list[Component]) -> None:
         """Adds new components to the simulation."""
         self._component_manager.add_components(component_list)
 
-    def get_population(self) -> pd.Series[Any] | pd.DataFrame:
```
stevebachmeier, line 376: Removed this so removed the corresponding constraint above

src/vivarium/interface/interactive.py
stevebachmeier: So I _think_ what you're saying is that we only want to support True or False (the default), but if False we actually pass None to the population_view calls. I'll do that for now assuming my interpretation is correct.

tests/framework/population/test_population_view.py
stevebachmeier: I think we are appropriately not allowing these calls before population creation by restricting `PopulationManager.get_population` during setup and post-setup?

stevebachmeier: https://github.com/ihmeuw/vivarium/blob/release_candidate/v4.0.0/src/vivarium/framework/population/manager.py#L126-L132

tests/interface/test_interactive.py
stevebachmeier: I wasn't sure. As of now, no I don't think so b/c as discussed the index is already reduced by the time we're popping it off of the private_columns dataframe. This is one of those tests that AI likely would have said "don't bother" but I'd already written it. Happy to remove this parametrization if you want.

tests/interface/test_interactive.py
stevebachmeier: Well the problem is that we first make an assertion without having things tracked. Then we track here and make further assertions. I could probably refactor since we don't need to necessarily check prior to registering a tracked query in every test if you think I should.

tests/framework/population/test_population_view.py
```diff
@@ -867,7 +867,7 @@ def test__update_column_and_ensure_dtype_unmatched_dtype() -> None:
 @pytest.mark.parametrize("include_untracked", [None, False, True])
 @pytest.mark.parametrize(
     "lifecycle_state",
-    [state.lower() for state in dir(lifecycle_states) if not state.startswith("_")],
+    ["initialization", "time_step"],
```
rmudambi: This should be `["population_creation", "time_step"]`

src/vivarium/interface/interactive.py
```diff
@@ -166,15 +166,28 @@ def take_steps(
...
+        include_untracked: bool = False,
+    ) -> pd.DataFrame:
         ...
 
     def get_population(
-        self, attributes: str | list[str] | tuple[str, ...] | None = None
+        self,
+        attributes: str | list[str] | tuple[str, ...] | None = None,
+        query: str = "",
+        include_untracked: bool = False,
```
stevebachmeier, line 190: @rmudambi this is what you meant, right? We don't really want bool | None here?

tests/interface/test_interactive.py
rmudambi: I'd suggest removing it to reduce the load on our test suite

rmudambi (approved):

albrja (approved):

PR #767, opened 2026-03-11
enable registering pipelines with a sequence of post-processors

rmudambi:
### Description

- *Category*: feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6575

### Changes and notes
- Enable registering attribute and rate producers with a sequence of post-processors
-  Allow register_rate_producer to take a preferred combiner
- Add type overloads to _configure_pipeline
- Allow calling value pipelines (not attribute piplelines) during post-setup
- Add a raw_union_post_processor which doesn't take an index argument (this is needed for paf pipelines)

### Testing
Added a new test for rate producers
Ran make check in vivarium and VPH


hussain-jafari (approved):

src/vivarium/framework/values/pipeline.py
```diff
@@ -215,8 +215,9 @@ def _call(self, *args: Any, skip_post_processor: bool = False, **kwargs: Any) ->
         value = self.source(*args, **kwargs)
         for mutator in self.mutators:
             value = self.combiner(value, mutator, *args, **kwargs)
-        if self.post_processor and not skip_post_processor:
-            return self.post_processor(value, self.manager)
+        if not skip_post_processor:
+            for processor in self.post_processor:
```
stevebachmeier, line 219: nit: wonder if it's worth changing this to plural

src/vivarium/framework/values/post_processors.py
```diff
@@ -79,9 +79,7 @@ def rescale_post_processor(
         raise NotImplementedError
 
 
-def union_post_processor(
-    index: pd.Index[int], value: list[NumberLike], manager: ValuesManager
-) -> pd.Series[Any] | pd.DataFrame:
+def raw_union_post_processor(value: list[NumberLike], manager: ValuesManager) -> NumberLike:
```
stevebachmeier, line 82: Why split this raw version out?

stevebachmeier (approved):

src/vivarium/framework/values/pipeline.py
rmudambi: I didn't, because I figured that was a breaking change

src/vivarium/framework/values/post_processors.py
rmudambi: We use the raw version in the calibration constant pipeline. We can't use the original one, because that requires the simulation index and broadcasts the values to that index

patricktnast (approved):

PR #776, opened 2026-04-03
Implement PopulationView.get mode parameter

rmudambi:
### Description

- *Category*: feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6935

### Changes and notes
- Implement PopulationView.get mode parameter
- Fix bugs in pipeline sources
- Deprecate `skip_post_processor` parameter
- Added missing overloads to `__get_attributes`

### Testing
Introduced new tests
Ran make check

### DIFF
--- a/src/vivarium/framework/population/manager.py
+++ b/src/vivarium/framework/population/manager.py
@@ -476,7 +476,7 @@ class PopulationManager(Manager):
         index: pd.Index[int] | None = None,
         query: str = "",
         squeeze: Literal[True] = True,
-        skip_post_processor: Literal[False] = False,
+        mode: Literal["default"] = "default",
     ) -> pd.Series[Any] | pd.DataFrame:
         ...
 
@@ -487,7 +487,7 @@ class PopulationManager(Manager):
         index: pd.Index[int] | None = None,
         query: str = "",
         squeeze: Literal[False] = ...,
-        skip_post_processor: Literal[False] = False,
+        mode: Literal["default"] = "default",
     ) -> pd.DataFrame:
         ...
 
@@ -498,7 +498,7 @@ class PopulationManager(Manager):
         index: pd.Index[int] | None = None,
         query: str = "",
         squeeze: Literal[True, False] = True,
-        skip_post_processor: Literal[True] = ...,
+        mode: Literal["source", "skip_post_processor"] = ...,
     ) -> Any:
         ...
 
@@ -508,7 +508,7 @@ class PopulationManager(Manager):
         index: pd.Index[int] | None = None,
         query: str = "",
         squeeze: Literal[True, False] = True,
-        skip_post_processor: Literal[True, False] = False,
+        mode: Literal["default", "source", "skip_post_processor"] = "default",
     ) -> Any:
         """Provides a copy of the population state table.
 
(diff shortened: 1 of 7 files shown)


patricktnast (approved):

src/vivarium/framework/population/manager.py
```diff
@@ -498,7 +498,7 @@ def get_population(
         index: pd.Index[int] | None = None,
         query: str = "",
         squeeze: Literal[True, False] = True,
-        skip_post_processor: Literal[True] = ...,
+        mode: Literal["source", "skip_post_processor"] = ...,
```
stevebachmeier: nit: It kind of bothers me that two of the modes are nouns and this one is a verb. I can't really think of anything better offhand though. Maybe something like `processing: Literal["full", "partial", "none"] = "full"`? May not be worth the effot though

src/vivarium/framework/population/population_view.py
```diff
@@ -168,7 +188,21 @@ def get(
...
         ValueError
             If the result is expected to be a Series but is not.
+            If an invalid mode is provided.
         """
+        valid_modes = ("default", "source", "skip_post_processor")
+        if mode not in valid_modes:
+            raise ValueError(f"Invalid mode '{mode}'. Must be one of {valid_modes}.")
+
+        # Translate skip_post_processor into mode
+        if skip_post_processor:
```
stevebachmeier, line 197: imo every time we add a deprecation warning we should add a corresponding test that simply starts failing in XXX months. I've done that several times now and it's pretty conventent and ensures we don't end up w/ bloated signatures or code blocks for too long.

stevebachmeier (approved):

src/vivarium/framework/population/population_view.py
stevebachmeier, line 197: Should probably raise for conflicting requests, e.g. `skip_post_processor = True and mode != "skip_post_processor"` (and the opposite). And add tests.

src/vivarium/framework/population/manager.py
```diff
@@ -672,15 +671,15 @@ def _get_attributes(
...
     ) -> Any:
         ...
 
     def _get_attributes(
         self,
         idx: pd.Index[int],
         requested_attributes: Sequence[str],
-        skip_post_processor: Literal[True, False] = False,
+        mode: Literal["default", "source", "skip_post_processor"] = "default",
```
albrja: These strings should be containerized.

src/vivarium/framework/population/population_view.py
albrja: +1

albrja (approved):

src/vivarium/framework/population/population_view.py
rmudambi: I actually disagree. I think the only time we *might* want to error here is if they select `skip_post_processor=True` and `mode="source"`. Otherwise they might just be using the defaults.

src/vivarium/framework/population/manager.py
rmudambi: These strings are user facing, so if we modify them, that would be a breaking change. Removing the friction in changing them by abstracting them to constants would actually be a bad idea in my opinion.

PR #784, opened 2026-04-28
Albrja/mic 6867/observer priority

albrja:
### Allow observations to have specific priority on lifecycle events
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6867

### Changes and notes
-Event and observation now have priority attributes
-ResultsManager registers listener for every lifecycle event priority

## Code-reviewer:
PR Review: Add support for observations to specify priority for lifecycle events
Summary
This PR adds a priority parameter to the observation registration API in the results framework. Observations can now specify an integer priority (0–9, default 5) that controls when they fire relative to other event listeners in the same lifecycle phase. The implementation works by registering the ResultsManager as a listener at every priority level (0–9) for each lifecycle phase, and then filtering observations in get_observations() to only return those whose priority matches the current event's priority. This enables observers to guarantee their observations run before or after other components' event handlers.

Design
Registering 40 listeners (10 × 4 phases) is eager and wasteful in the common case — manager.py:80-100. When all observations use the default priority (the typical case), 36 of 40 gather_results calls per step do nothing — they create an Event, traverse grouped_observations, find zero matches, and return. Consider registering only at the default priority eagerly, then lazily registering additional priority listeners when a non-default-priority observation is actually registered.

get_observations does a full linear scan of all observations per call — context.py:368-375. The grouped_observations dict keys on (lifecycle_state, PopulationFilter, stratifications) but not on priority. Each call iterates every observation for the phase and checks observation.priority == event.priority. A secondary index keyed by (phase, priority) would make this O(1) lookup instead of O(n).

gather_results re-filters observations that get_observations already selected — context.py:318-330. After get_observations returns the priority-filtered list, gather_results iterates grouped_observations again and uses if observation in event_observations (list membership, O(n)) in nested loops. Converting event_observations to a set would improve this, though the re-traversal pattern is pre-existing.

No validation of priority range — manager.py:266-310, interface.py:199. A user can pass priority=15 or priority=-1. The observation registers silently but never fires, since no event is emitted at that priority. This is a silent-failure bug. Add a bounds check in ResultsManager.register_observation() that raises ValueError for priorities outside range(10).

Maintainability
Magic number range(10) couples results manager to event system — manager.py:80. The 10-priority assumption is hardcoded in both ResultsManager.setup() and EventChannel.__init__ (manager.py:96) but is not a named constant. If either drifts, observations at certain priorities silently never fire. Extract a shared constant (e.g., NUM_EVENT_PRIORITIES = 10 in vivarium/component.py alongside DEFAULT_EVENT_PRIORITY).

No comment explaining the 40-listener registration strategy — manager.py:80-100. A developer reading setup() for the first time will find the for priority in range(10) loop surprising. Add a brief comment explaining why the manager registers at every priority level.

DRY
Four identical on_* handler methods — manager.py:110-122. on_time_step_prepare, on_time_step, on_time_step_cleanup, and on_collect_metrics all have the same body: self.gather_results(event). Since the Event already carries its name (lifecycle state) and priority, these four methods are functionally identical. Register self.gather_results directly as the listener to eliminate four trivial wrappers:

for priority in range(10):    for phase in [TIME_STEP_PREPARE, TIME_STEP, TIME_STEP_CLEANUP, COLLECT_METRICS]:        builder.event.register_listener(phase, self.gather_results, priority=priority)
Note: this is a pre-existing pattern, but the PR makes it more visible by expanding the registration from 4 to 40.

PopulationFilter(pop_filter, include_untracked) construction repeated in all 4 interface registration methods — interface.py:258, interface.py:314, interface.py:390, interface.py:459. Pre-existing, but adding priority as another common parameter deepens the repetition.

Tests
test_observation_priority_ordering could pass even if priority filtering is broken — test_manager.py:889-927. The "early" observation (priority 2) is registered before "default" (priority 5). If the priority filter were removed and observations fired in registration order, the test would still pass. Add a variant where the higher-priority observation is registered first to prove ordering is by priority, not registration order.

No unit test for get_observations filtering by priority — test_context.py:600-649. The existing test_get_observations only uses DEFAULT_EVENT_PRIORITY. A test should register observations at different priorities and verify that get_observations returns only those matching the event's priority.

No boundary tests for priority 0 and 9 — If range(10) in setup() were accidentally changed to range(1, 10), no test would catch the off-by-one.

No test for out-of-range priority values — If validation is added (per Design finding #4), it needs a test. If validation is intentionally absent, a test should document the silent-failure behavior.

Existing registration tests don't assert observation.priority — test_interface.py. test_register_stratified_observation and similar tests verify name, when, population_filter, etc., but none checks observation.priority. If priority were silently dropped during registration, no test would fail.

Documentation
register_observation docstring in ResultsContext omits priority — context.py:220-260. The Parameters section lists observation_type, name, population_filter, and when, but not priority, requires_attributes, or **kwargs, despite all being explicit parameters.

Observation.priority docstring doesn't state valid range or default — observation.py:88-89. Reads "The priority level of this observation. Observations with lower priority levels are observed first." Should mention valid values are 0–9 and the default is 5.

Subclass Attributes docstrings omit priority — observation.py. The Attributes sections for UnstratifiedObservation, StratifiedObservation, AddingObservation, and ConcatenatingObservation list all parameters except priority.

CHANGELOG entry is terse — CHANGELOG.rst:3. "Add support for observations to specify priority for lifecycle events" doesn't explain what priority means, the valid range, the default, or which methods gained the parameter.

Functionality
Silent no-op for out-of-range priorities is a correctness risk. A user passing priority=10 gets an observation that registers successfully, appears in self.observations, but is never matched by get_observations because no event is ever emitted at priority 10. The get_results() method would still return the observation's initialized (zero/empty) results, potentially causing confusing output. This is the most critical functional issue in the PR.

Negative priority values exploit Python's negative indexing. Passing priority=-1 to register_adding_observation would not cause an error anywhere in the results framework. In the event system, EventChannel.listeners[-1] would be listeners[9] (Python negative indexing), so the observation would actually fire at priority 9. This is semantically wrong and could mask bugs. The bounds check recommended in Design #4 would fix this.

gather_results called with observations=[] returns early before _prepare_population — manager.py:131-135. This is correct: when no observations match the current priority, the method short-circuits. No correctness issue here.

Event.priority type is int with no constraint — manager.py:61. The Event dataclass is a @dataclass (pre-existing), and priority is typed as int with no validation. This is a pre-existing issue that the PR inherits rather than introduces.

Minor Nits
manager.py:82 — The for priority in range(10) loop body could use a list of (phase, handler) tuples to avoid repeating the builder.event.register_listener call 4 times per iteration.

test_manager.py:897 — The PriorityOrderingObserver inner class could use a class-level execution_order list instead of closing over a mutable from the enclosing scope, which would make the test more robust if parallelized.

Overall
The PR delivers a clean and useful feature — allowing observations to control their firing priority relative to other event listeners. The core mechanism (matching observation.priority == event.priority) is correct and integrates naturally with the existing event system.

The highest-priority improvements are:

Add priority validation to prevent silent failures with out-of-range values (Design #4, Functionality #1-#2)
Add a get_observations unit test that exercises priority filtering (Tests #2) and a reverse-registration-order integration test (Tests #1)
Extract the hardcoded range(10) into a named constant shared with the event system (Maintainability #1)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,7 @@
+**4.1.2 - 04/28/26**
+
+- Add support for observations to specify priority for lifecycle events
+
 **4.1.1 - 04/21/26**
 
 - Raise error if registering duplicate Pipelines
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,7 @@
-**4.1.2 - 04/28/26**
+**4.1.2 - 04/29/26**
 
 - Add support for observations to specify priority for lifecycle events
+- Added priority attribute to `Event`
 
 **4.1.1 - 04/21/26**
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,4 +1,4 @@
-**4.1.2 - 04/29/26**
+**4.1.2 - 05/04/26**
 
 - Add support for observations to specify priority for lifecycle events
 - Added priority attribute to `Event`
--- a/src/vivarium/component.py
+++ b/src/vivarium/component.py
@@ -33,6 +33,9 @@ if TYPE_CHECKING:
     from vivarium.framework.population import PopulationView
     from vivarium.types import DataInput
 
+NUM_EVENT_PRIORITIES = 10
+"""The number of priority levels available for event listeners."""
+
 DEFAULT_EVENT_PRIORITY = 5
 """The default priority at which events will be triggered."""
 
--- a/src/vivarium/framework/results/interface.py
+++ b/src/vivarium/framework/results/interface.py
@@ -14,6 +14,7 @@ from typing import TYPE_CHECKING, Any, NamedTuple, Sequence, Union
 import pandas as pd
 from pandas.core.groupby.generic import DataFrameGroupBy
 
+from vivarium.component import DEFAULT_EVENT_PRIORITY
 from vivarium.framework.event import Event
 from vivarium.framework.lifecycle import lifecycle_states
 from vivarium.framework.results.observation import (
@@ -195,6 +196,7 @@ class ResultsInterface(Interface):
         pop_filter: str = "",
         include_untracked: bool = False,
         when: str = lifecycle_states.COLLECT_METRICS,
+        priority: int = DEFAULT_EVENT_PRIORITY,
         requires_attributes: list[str] = [],
         results_updater: ResultsUpdater = _required_function_placeholder,
         results_formatter: ResultsFormatter = _default_stratified_observation_formatter,
@@ -219,6 +221,9 @@ class ResultsInterface(Interface):
         when
             Name of the lifecycle phase the observation should happen. Valid values are:
             "time_step__prepare", "time_step", "time_step__cleanup", or "collect_metrics".
+        priority
+            The priority level of this observation within its lifecycle phase.
+            Observations with lower priority values are observed first.
         requires_attributes
             The population attributes that are required by the `aggregator`.
         results_updater
@@ -249,6 +254,7 @@ class ResultsInterface(Interface):
             name=name,
             population_filter=PopulationFilter(pop_filter, include_untracked),
             when=when,
+            priority=priority,
             requires_attributes=requires_attributes,
             results_updater=results_updater,
             results_formatter=results_formatter,
@@ -265,6 +271,7 @@ class ResultsInterface(Interface):
         pop_filter: str = "",
         include_untracked: bool = False,
         when: str = lifecycle_states.COLLECT_METRICS,
+        priority: int = DEFAULT_EVENT_PRIORITY,
         requires_attributes: list[str] = [],
         results_gatherer: ResultsGatherer = _required_function_placeholder,
         results_updater: ResultsUpdater = _required_function_placeholder,
@@ -286,6 +293,9 @@ class ResultsInterface(Interface):
         when
             Name of the lifecycle phase the observation should happen. Valid values are:
             "time_step__prepare", "time_step", "time_step__cleanup", or "collect_metrics".
+        priority
+            The priority level of this observation within its lifecycle phase.
+            Observations with lower priority values are observed first.
         requires_attributes
             The population attributes that are required by the `results_gatherer`.
         results_gatherer
@@ -312,6 +322,7 @@ class ResultsInterface(Interface):
             name=name,
             population_filter=PopulationFilter(pop_filter, include_untracked),
             when=when,
+            priority=priority,
             requires_attributes=requires_attributes,
             results_updater=results_updater,
             results_gatherer=results_gatherer,
@@ -325,6 +336,7 @@ class ResultsInterface(Interface):
         pop_filter: str = "",
         include_untracked: bool = False,
         when: str = lifecycle_states.COLLECT_METRICS,
+        priority: int = DEFAULT_EVENT_PRIORITY,
         requires_attributes: list[str] = [],
         results_formatter: ResultsFormatter = _default_stratified_observation_formatter,
         additional_stratifications: list[str] = [],
@@ -355,6 +367,9 @@ class ResultsInterface(Interface):
         when
             Name of the lifecycle phase the observation should happen. Valid values are:
             "time_step__prepare", "time_step", "time_step__cleanup", or "collect_metrics".
+        priority
+            The priority level of this observation within its lifecycle phase.
+            Observations with lower priority values are observed first.
         requires_attributes
             The population attributes that are required by the `aggregator`.
         results_formatter
@@ -377,6 +392,7 @@ class ResultsInterface(Interface):
             name=name,
             population_filter=PopulationFilter(pop_filter, include_untracked),
             when=when,
+            priority=priority,
             requires_attributes=requires_attributes,
             results_formatter=results_formatter,
             additional_stratifications=additional_stratifications,
@@ -392,6 +408,7 @@ class ResultsInterface(Interface):
         pop_filter: str = "",
         include_untracked: bool = False,
         when: str = lifecycle_states.COLLECT_METRICS,
+        priority: int = DEFAULT_EVENT_PRIORITY,
         requires_attributes: list[str] = [],
         results_formatter: ResultsFormatter = _default_unstratified_observation_formatter,
         to_observe: Callable[[Event], bool] = lambda event: True,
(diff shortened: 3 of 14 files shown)


src/vivarium/framework/results/interface.py
```diff
@@ -219,6 +221,9 @@ def register_stratified_observation(
         when
             Name of the lifecycle phase the observation should happen. Valid values are:
             "time_step__prepare", "time_step", "time_step__cleanup", or "collect_metrics".
+        priority
+            The priority level of this observation within its lifecycle phase.
+            Observations with lower priority values are observed first.
```
rmudambi: This docstring doesn't seem quite right to me. It suggests that this is defining the order in which observations are recorded. Really what it does is say observations are recorded during this numbered priority stage of the given lifecycle state. Look at other documentation around the "priority" in the lifecycle, and try to match that language.

tests/framework/results/test_manager.py
```diff
@@ -879,3 +884,100 @@ def _assert_standard_index(df: pd.DataFrame) -> None:
     assert not isinstance(df.index, pd.MultiIndex)
     assert df.index.names == [None]
     assert df.index.dtype == "int64"
+
+
+def test_observation_priority_ordering() -> None:
+    """Test that observations with lower priority fire before those with higher priority."""
```
rmudambi: We don't care about the order in which observations fire - we care that they fire on the correct priority stage of a given lifecycle state.

CHANGELOG.rst
```diff
@@ -1,3 +1,7 @@
+**4.1.2 - 04/28/26**
+
+- Add support for observations to specify priority for lifecycle events
```
rmudambi, line 3: Also note that you added the priority attribute to `Event`

src/vivarium/framework/results/observation.py
rmudambi, line 149: We're missing the priority argument here and elsewhere in this module.

src/vivarium/framework/results/interface.py
```diff
@@ -286,6 +292,9 @@ def register_unstratified_observation(
         when
             Name of the lifecycle phase the observation should happen. Valid values are:
             "time_step__prepare", "time_step", "time_step__cleanup", or "collect_metrics".
+        priority
+            The priority of the lifecycle phase this observation will record. Observations
+            with lower priority levels are recorded first.
```
rmudambi: This docstring should match `register_stratified_observation()`

src/vivarium/framework/results/interface.py
```diff
@@ -355,6 +366,9 @@ def register_adding_observation(
         when
             Name of the lifecycle phase the observation should happen. Valid values are:
             "time_step__prepare", "time_step", "time_step__cleanup", or "collect_metrics".
+        priority
+            The priority of the lifecycle phase this observation will record. Observations
```
rmudambi: See above

src/vivarium/framework/results/interface.py
```diff
@@ -418,6 +434,9 @@ def register_concatenating_observation(
...
```
rmudambi: See above

src/vivarium/framework/results/manager.py
```diff
@@ -274,13 +286,22 @@ def register_observation(
         when
             Name of the lifecycle phase the observation should happen. Valid values are:
             "time_step__prepare", "time_step", "time_step__cleanup", or "collect_metrics".
+        priority
+            The priority of the lifecycle phase this observation will record. Observations
```
rmudambi: See comment in interface

src/vivarium/framework/results/observation.py
```diff
@@ -84,6 +85,9 @@ class Observation(ABC):
     """Method or function that determines whether to perform an observation on this Event."""
     stratifications: tuple[Stratification, ...] | None = None
     """Optional tuple of the Stratifications this observation should use."""
+    priority: int = DEFAULT_EVENT_PRIORITY
+    """The priority of the lifecycle phase this observation will record. Observations
```
rmudambi: See comment in interface.

rmudambi (approved):
Approved assuming you fix the docstrings

src/vivarium/framework/results/manager.py
```diff
@@ -274,13 +286,21 @@ def register_observation(
...
+        priority
+            The priority level of the lifecycle phase (see `when`) that this observation will record.
         requires_attributes
             The population attributes that are required to compute the observation.
         **kwargs
             Additional keyword arguments to be passed to the observation's constructor.
         """
         self.logger.debug(f"Registering observation {name}")
 
+        if not (0 <= priority < NUM_EVENT_PRIORITIES):
```
stevebachmeier, line 298: Why is this check here in results but not in `emit` when instantiating the events?

src/vivarium/component.py
```diff
@@ -33,6 +33,9 @@
     from vivarium.framework.population import PopulationView
     from vivarium.types import DataInput
 
+NUM_EVENT_PRIORITIES = 10
```
stevebachmeier, line 36: Hmm...I thought that part of this effort was to allow for this number to be configurable. I think we'd want to expose this (and perhaps the default priority below while we're at it) via properties or something?

stevebachmeier (changes requested):
Don't we want the number of priority levels to be configurable?

albrja:
@stevebachmeier It is. When you register an observation via the results interface, there is a new argument for priority.

src/vivarium/component.py
albrja: It is. This is just as constant so we don't have 10 floating around in several places.

src/vivarium/framework/results/manager.py
albrja: 2 reasons:
1. This is the public API to add observations. Users could pass any priority level in that they wanted and it would not fail until later on (in emit specifically). This gives a fast fail with a clear error message.
2. In emit, we are creating events and passing them to event listeners for each priority and listener. Self.listeners is defined as: `self.listeners: list[list[Callable[[Event], None]]] = [
            [] for _ in range(NUM_EVENT_PRIORITIES)
        ]`
So there, we are naturally constrained already by the number of priorities. We could add this check but it would be unnecessary IMO

stevebachmeier (approved):

patricktnast (approved):

PR #788, opened 2026-05-08
Albrja/mic 6950/observer categorical dtype

albrja:
### Write out observer outputs as categorical dtype
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6950

### Changes and notes
-write parquet files with categorical dtype to save memory

### Code reviewer
PR Review: Write observer outputs as categorical dtype (#788)
Summary
This PR converts stratification columns in observer results to use pandas CategoricalDtype. Two source files are modified: create_expanded_df() in observation.py now casts the initial stratification DataFrame to CategoricalDtype(), and stratify() in stratification.py now returns a pd.Series[CategoricalDtype] with explicit categories and ordered=True. Test assertions in test_manager.py are updated to accommodate the new dtype in index comparisons. The CHANGELOG adds a 4.1.4 entry.

Design
Inconsistent CategoricalDtype construction between initialization and runtime. In observation.py:321, CategoricalDtype() (no args, unordered, inferred categories) is used, while in stratification.py:103-105, CategoricalDtype(categories=self.categories, ordered=True) is used. When add_results at observation.py:478 performs updated_results[col] += new_observations[col], index alignment occurs between unordered-categorical (initialized) and ordered-categorical (observed) indexes. This works in current pandas but is fragile — a future pandas version could tighten categorical alignment semantics. Suggestion: Use explicit categories and consistent ordered in create_expanded_df by iterating over self.stratifications:

for stratification in self.stratifications:    df[stratification.name] = df[stratification.name].astype(        CategoricalDtype(categories=stratification.categories, ordered=True)    )
ordered=True is semantically misleading for non-ordinal stratifications. In stratification.py:103-105, every stratification is returned as ordered. For nominal categories like student_house, ordering is meaningless and enables silent nonsensical comparisons (e.g., series > "gryffindor"). Pandas sorts categoricals by category list order regardless of the ordered flag. Suggestion: Default to ordered=False unless ordering is semantically meaningful; consider adding an ordered attribute to Stratification if ordinal stratifications are needed.

Unnecessary categorical conversion of the "all" degenerate case. In observation.py:323-327, the single "all" value is cast to CategoricalDtype(). But the no-stratification path in _get_groups at context.py:479 uses groupby(lambda _: "all"), producing a plain string key. The categorical conversion provides no benefit here and adds unnecessary complexity.

_expand_index is now potentially redundant. With categorical columns and observed=False in _get_groups at context.py:474, groupby results already include all category combinations. _expand_index at observation.py:382-390 is effectively a no-op. Consider adding a comment noting this is a defensive fallback.

Maintainability
No inline comment or docstring explaining why categorical dtype is used. Neither create_expanded_df nor stratify() explains the motivation (memory efficiency? Parquet compatibility? Enforcing valid categories?). A future developer may remove it as unnecessary complexity. Suggestion: Add a brief comment at the conversion sites.

HACK comment in test explains how but not why. The comment at test_manager.py:293 explains the MultiIndex workaround but not why CategoricalDtype is needed on the levels. Suggestion: Add: # The production code converts stratification columns to CategoricalDtype before set_index, so expected indices must also be categorical to pass .equals() comparison.

Implicit coupling between create_expanded_df and _expand_index. _expand_index calls pd.MultiIndex.from_product(aggregates.index.levels) where .levels now returns CategoricalIndex objects. Whether from_product preserves categorical dtype is an undocumented pandas implementation detail. This is the most fragile implicit assumption in the change.

DRY
Duplicated assertion block in test_stratified_observation_results. The block at test_manager.py:671-700 after each sim.step() is nearly identical (only the exponent differs). Suggestion: Extract a local helper _assert_cat_bomb_results(sim, exponent).

.astype(CategoricalDtype()) duplicated in both branches of create_expanded_df. At observation.py:316-327, both the if and else branches end with .astype(CategoricalDtype()). Suggestion: Apply the cast once after the branch:

# ... if/else creates df ...df = df.astype(CategoricalDtype())
Tests
test_no_stratifications__raw_results_initialization doesn't validate categorical dtype. At test_manager.py:318, assert raw_results.index.equals(pd.Index(["all"])) passes because Index.equals() compares values, not dtypes. A regression removing .astype(CategoricalDtype()) would go undetected. Suggestion: Either assert pd.CategoricalIndex or remove the categorical conversion for the "all" case (per Design finding #3).

test_stratification in test_stratification.py does not verify return dtype or ordering. At test_stratification.py:56-57, only value equality is checked. The ordered=True behavior and CategoricalDtype return are completely untested at the unit level. Suggestion:

assert isinstance(output.dtype, CategoricalDtype)assert output.cat.ordered is Trueassert list(output.cat.categories) == HOUSE_CATEGORIES
No unit test for create_expanded_df categorical behavior. test_observation.py has no tests for create_expanded_df at all. Coverage is only indirect through integration tests.

HACK-built expected index uses unparameterized CategoricalDtype. At test_manager.py:296, CategoricalDtype (no categories, unordered) is used to build the expected index, which matches the production code's create_expanded_df behavior but would not catch a regression where the production code switched to parameterized categories.

Categorical dtype survival through update cycle is untested. If add_results silently casts categorical indexes back to object, the categorical guarantee would be broken, but no test checks this.

Documentation
CHANGELOG entry is terse. CHANGELOG.rst:3 says "Feature: Write observer outputs as categorical dtype" without explaining the motivation or impact. Suggestion: Expand to indicate why (e.g., "for more efficient storage and consistent groupby behavior").

stratify() docstring omits the categorical return type. At stratification.py:93, the Returns section says "A Series containing the mapped values" without mentioning categorical dtype. Suggestion: "A categorical Series containing the mapped values to be used for stratifying."

create_expanded_df() docstring doesn't mention categorical index. At observation.py:292, the docstring says "Initializes a dataframe of 0s" but doesn't mention the stratification index is now categorical.

Functionality
Index alignment between initialized and observed results is fragile. add_results at observation.py:478 does updated_results[col] += new_observations[col]. The existing_results (from create_expanded_df) has an unordered categorical index with inferred categories, while new_observations (from get_complete_stratified_results → stratify()) has an ordered categorical index with explicit categories. Current pandas aligns by label, so this works. But the dtype mismatch means existing_results.index.dtype != new_observations.index.dtype, which is an invariant violation waiting to surface.

The "all" path has a categorical/non-categorical mismatch. create_expanded_df produces CategoricalIndex(["all"]), but _get_groups((), filtered_pop) produces groups with a plain string key "all". The resulting new_observations from get_complete_stratified_results has a plain Index(["all"]). When add_results aligns these, categorical-vs-object alignment works in current pandas, but this is an undocumented coincidence.

CatBombObserver.update_cats at helpers.py:242-248 does custom index-based operations. This uses existing_df["value"] == 0 masking and assignment — these operations should be robust to categorical index dtype since they operate on values, not indices. No issue found here.

Minor Nits
observation.py:30: CategoricalDtype is imported from pandas.api.types in both observation.py and stratification.py. This is fine — it's the canonical pandas import path.

test_manager.py:296: CategoricalDtype is used as a class (not an instance) in .astype(CategoricalDtype) — this works because pandas accepts the class as shorthand for CategoricalDtype(), but it's inconsistent with the instance usage elsewhere. Consider using CategoricalDtype() for consistency.

Overall
The core idea of using categorical dtype for stratification columns is sound and provides benefits for downstream consumers (e.g., parquet serialization, memory, enforcing valid values). However, the implementation has a consistency gap between the two categorical dtype constructions (Finding Design #1 / Functionality #1) that should be resolved before merge — the initialized results should use the same explicit categories and ordering as the runtime stratification. The test coverage for the new dtype behavior has meaningful gaps: the unit-level tests for stratify() and create_expanded_df() don't validate the dtype at all, and the no-stratification test doesn't detect the change. The ordered=True choice should be reconsidered for non-ordinal stratifications.

### DIFF
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -336,6 +336,9 @@ class SimulationContext:
             results_dir = self.configuration.output_data.results_directory
             for measure, df in results.items():
                 output_file = Path(results_dir) / f"{measure}.parquet"
+                df = df.copy()
+                object_cols = df.select_dtypes(include=["object"]).columns
+                df[object_cols] = df[object_cols].astype("category")
                 df.to_parquet(output_file, index=False)
         except ConfigurationKeyError:
             self._logger.info("No results directory set; results are not written to disk.")
--- a/src/vivarium/framework/engine.py
+++ b/src/vivarium/framework/engine.py
@@ -336,9 +336,6 @@ class SimulationContext:
             results_dir = self.configuration.output_data.results_directory
             for measure, df in results.items():
                 output_file = Path(results_dir) / f"{measure}.parquet"
-                df = df.copy()
-                object_cols = df.select_dtypes(include=["object"]).columns
-                df[object_cols] = df[object_cols].astype("category")
                 df.to_parquet(output_file, index=False)
         except ConfigurationKeyError:
             self._logger.info("No results directory set; results are not written to disk.")
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,7 @@
+**4.1.4 - 05/08/26**
+
+- Feature: Write observer outputs as categorical dtype 
+
 **4.1.3 - 05/07/26**
 
 - Refactor: Make Interpolation __call__ method more performant.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,6 @@
 **4.1.5 - 05/08/26**
 
-- Feature: Write observer outputs as categorical dtype 
+- Feature: Convert stratified observer object columns to categorical
 
 **4.1.4 - 05/07/26**
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,10 @@
-**4.1.4 - 05/08/26**
+**4.1.5 - 05/08/26**
 
 - Feature: Write observer outputs as categorical dtype 
+
+**4.1.4 - 05/07/26**
+
+- Optimization: Subset to non-zero values before applying multiplication_combiner
 
 **4.1.3 - 05/07/26**
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,4 +1,4 @@
-**4.1.5 - 05/08/26**
+**4.1.5 - 05/11/26**
 
 - Feature: Convert stratified observer object columns to categorical
 
--- a/src/vivarium/framework/results/interface.py
+++ b/src/vivarium/framework/results/interface.py
@@ -55,17 +55,24 @@ def _required_function_placeholder(
     return pd.DataFrame()
 
 
+def _convert_object_cols_to_categorical(results: pd.DataFrame) -> pd.DataFrame:
+    """Convert object dtype columns to categorical dtype."""
+    object_cols = results.select_dtypes(include=["object"]).columns
+    results[object_cols] = results[object_cols].astype("category")
+    return results
+
+
 def _default_stratified_observation_formatter(
     measure: str, results: pd.DataFrame
 ) -> pd.DataFrame:
-    """Resets the results index."""
-    return results.reset_index()
+    """Reset the results index and convert object columns to categorical dtype."""
+    return _convert_object_cols_to_categorical(results.reset_index())
 
 
 def _default_unstratified_observation_formatter(
     measure: str, results: pd.DataFrame
 ) -> pd.DataFrame:
-    """Returns the results unchanged."""
+    """Return the results unchanged."""
     return results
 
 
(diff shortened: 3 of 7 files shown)


src/vivarium/framework/engine.py
```diff
@@ -336,6 +336,9 @@ def _write_results(self, results: dict[str, pd.DataFrame]) -> None:
             results_dir = self.configuration.output_data.results_directory
             for measure, df in results.items():
                 output_file = Path(results_dir) / f"{measure}.parquet"
+                df = df.copy()
+                object_cols = df.select_dtypes(include=["object"]).columns
+                df[object_cols] = df[object_cols].astype("category")
```
rmudambi: This isn't the right place to do this. Observer outputs aren't guaranteed to be dataframes. You should do this within the results formatter.

rmudambi (changes requested):

rmudambi (approved):

hussain-jafari (approved):

